PR #31, opened 2015-07-02
EC2 API additions

s2maki:
Added a few API calls for VPC manipulation and other miscellany.


amitmurthy:
If these changes are due to newer functionality in the API, we can just try with a newer EC2 wsdl - the current one is  https://github.com/amitmurthy/AWS.jl/blob/master/gen/wsdl/ec2_2013_02_01.wsdl - and regenerate the boilerplate code.

s2maki:
That's a really good point.  I'm not sure how we missed that.

I'll take a look at updating it and will replace this pull request as necessary.

amitmurthy:
I have removed the code generation from EC2 WSDL since AWS seems to have stopped publishing it.
We can merge this PR upon a rebase and removal of Julia 0.4 deprecation warnings.

PR #32, opened 2015-07-06
libcrypto isn't available for Windows

s2maki:
Substitute libeay32.dll when using Windows; both 32 and 64 bit versions provided.

These DLLs were copied from https://code.google.com/p/openssl-for-windows/, licensed under GPLv3.  Direct insertion into this repo may not be the best option for long-term maintenance, but it works for me.


tkelman:
The GPL license at that google code page makes no sense, openssl is under a GPL-incompatible license. Putting binaries here would also require distributing this package under the openssl license, not the current MIT.

A better option would be to get libcrypto from either https://build.opensuse.org/package/show/windows:mingw:win64/mingw64-libressl or https://build.opensuse.org/package/show/windows:mingw:win64/mingw64-openssl using BinDeps and the WinRPM.jl package.

s2maki:
Agreed. I wasn't able to find those at the time I went looking, but that does appear to be a superior solution.

s2maki:
replaced with #43

PR #44, opened 2015-09-21
EC2 API additions -- rebased

s2maki:
As requested in #31, I've rebased against latest and eliminated deprecation warnings.


amitmurthy:
Thanks!

PR #51, opened 2015-10-29
Changed urls in S3.jl to include the region.

jademackay:
This is required for any region other than US-standard (North Virginia).


amitmurthy:
Thanks.

PR #53, opened 2015-11-23
use AWS CLI tools, fix S3.put_object signature

sebastiang:
The AWS CLI tools will honor the following variables:
- AWS_ACCESS_KEY_ID
- AWS_SECRET_ACCESS_KEY
- AWS_DEFAULT_REGION

This commit updates the package to do the same, while preserving
existing defaults and extending the configuration-file alternative
to include specifying a region.


sebastiang:
Also updated `S3.put_object` to accept `IO` where it was previously needing `IOStream`. (I had an `IOBuffer` I wanted to give it.) The method body checks `isa(data,IO)` anyway.

amitmurthy:
Thanks.

PR #61, opened 2016-01-12
The following changes have been done:

mdpradeep:
1) Removed HTTPClient dependency by moving to Requests.jl
    2) Removed dependency on LibExpat by moving to LightXML.jl
    3) Implementation of SQS APIs
    4) Documentation update

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,7 @@
-julia 0.4-
+julia 0.4
 BinDeps
-LibExpat
-HTTPClient
+LightXML
+Requests
 JSON
+URIParser
 @windows WinRPM
--- a/README.md
+++ b/README.md
@@ -1,14 +1,20 @@
 ## AWS - Julia interface to Amazon Web Services
 
-This package is a WIP for providing a native Julia interface to the Amazon Web Services API
+This package provides a native Julia interface to the Amazon Web Services API
 
-Initially, the EC2 and S3 API will be supported.
+The following services are supported as of now:
+
+- EC2
+- S3
+- SQS
 
 
 ### Current status
-- Most of the APIs are yet untested. Any testing will be helpful. The REST API does not match exactly in certain cases
+- The basic APIs of EC2, S3 and SQS are tested
+- The advanced APIs (e.g. VPC related APIs, etc) of EC2 are yet untested. Any testing will be helpful. 
+- The REST API does not match exactly in certain cases
   with the WSDL. For the EC2 API, the bulk of the code is generated from the WSDL while it has been translated by hand for the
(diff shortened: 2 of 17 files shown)


mdpradeep:
@ViralBShah

REQUIRE
```diff
@@ -1,6 +1,7 @@
 julia 0.4-
```
tkelman: remove the trailing `-`

tkelman:
This could likely use some testing from existing users of the package, since it's a pretty major rewrite. Would certainly want to change the minor version number for the first tag after merging this.

ViralBShah:
Not sure how to locate users to ask them to test. I believe that reasonably, testing will only happen after tagging.

ViralBShah:
Would be great if @amitmurthy and @tanmaykm can have a look.

amitmurthy:
Nice work. LGTM.

All tests passing locally?

mdpradeep:
Yes, all tests in test/test*.jl are passing.

Regards,
Pradeep

tkelman:
Add a `test/runtests.jl` file so `Pkg.test` can work?

amitmurthy:
@tkelman that would require AWS credentials. All the tests actually use AWS resources and are charged.

mdpradeep:
I can, but it requires AWS credentials. I could create a separate id, but then anyone would be able to launch VMs, etc… So, not sure whether this is a good idea or not.

Regards,
Pradeep

amitmurthy:
No, not a good idea to make the credentials public.

tkelman:
Travis has features where you can encrypt secure env vars. Depends on the resource cost per time the tests run whether it's worth setting up. Would Pkg.test fail immediately if you don't have credentials set up?

amitmurthy:
Yes. The EC2 tests check if servers can be booted up using the AWS API, the S3 tests if buckets can be created and accessed and so on.

Travis encryption features may be worth looking into.

ViralBShah:
I think the resource costs will be tiny, and we can probably do it with the free tier on a new ID.

amitmurthy:
We can merge this now while we explore options for safely passing credentials to Travis.

REQUIRE
amitmurthy: @mdpradeep please update this and we can merge.

mdpradeep:
Updated

amitmurthy:
Can you squash the commits too?

mdpradeep:
Done. Let me know if this is fine.

ViralBShah:
Should we tag a new release, and bump the minor number?

tkelman:
Hard to say whether the tests here are representative of real world use cases, but when this gets tagged it should absolutely change at least the minor version number.

Issue #29, opened 2015-06-04
Conversion error with update of AWS package S3 API


AndrewBMartin:
I was getting this error when I installed julia 0.3.8 on Ubuntu Trusty:
```julia
ERROR: `convert` has no method matching convert(::Type{String}, ::Array{String,1})
 in convert at base.jl:13
 in do_request at /home/ubuntu/.julia/v0.3/AWS/src/S3.jl:571
 in get_object at /home/ubuntu/.julia/v0.3/AWS/src/S3.jl:392
```
The do_http method was returning an object where the http_headers dictionary had values of Array{String, 1} instead of Type{String} as the S3Response object expected.

I've fixed this temporarily in my codebase by checking to see if each header key has a value, and then if it does, assigning the first index of the array to the S3Response object.

For the version of AWS in the julia package repo, something might need to be changed to either have do_http return http_headers as a dictionary of Type{String} or update the S3Response type to have fields of Array{String, 1} instead of String.

merl-dev:
(I haven't made the code base changes to get around this)

Suddenly all of our S3 retrieval routines are broken, and this seems to have taken place right after you made some updates to the AWS.jl library 2 days ago.

We are heavily dependent on this library for our production systems, and all has come to a dead stop since.  Could you either revert or fix?  I'm gonna have about 100 hundred traders screaming at me if we don't get this online by tomorrow morning (we work sundays).
julia> ERROR: `convert` has no method matching convert(::Type{Int64}, ::Array{Int64,1})
        in convert at base.jl:13
        in do_request at /home/ubuntu/.julia/v0.3/AWS/src/S3.jl:571
        in get_object at /home/ubuntu/.julia/v0.3/AWS/src/S3.jl:392
        in get_s3 at /home/ubuntu/Merl/src/Merl_aws.jl:56

the last line here refers to the following call
resp = AWS.S3.get_object(env, bucket, src)
where env = your default AWSEnv method but with our ID and KEY and REGION: env=AWSEnv(;id=aws_id,key=aws_skey,ep=s3_region,timeout=60.0)
bucket is just a string with bucket name
and src is just the file name

amitmurthy:
The issue is actually with the latest HTTPClient commit.

https://github.com/JuliaWeb/HTTPClient.jl/commit/18a1cec7c934ab6d7d42916010ad0de31394f892 was changed a ways back to return an array of strings as part of the headers dict.

[...]

PR #66, opened 2016-05-21
fixed convert when S3 returns error, issue #29

merl-dev:
Only a couple of minor changes were required to fix conversion errors when credentials are invalid, incorrectly specified, requesting bucket contents for non-existent buckets. 

The error can now be found in the http_resp.obj as an AWS.S3.S3Error object.

### DIFF
--- a/src/S3.jl
+++ b/src/S3.jl
@@ -634,8 +634,8 @@ function do_request(env::AWSEnv, ro::RO; conv_to_string=true)
     if (http_resp.status > 299)
         if  (search(Base.get(http_resp.headers, "Content-Type", [""]), "/xml") != 0:-1)
             s3_resp.obj = S3Error(LightXML.root(LightXML.parse_string(bytestring(http_resp.data))))
-		else
-			s3_resp.pd = bytestring(http_resp.data)
+	else
+	    s3_resp.pd = bytestring(http_resp.data)
         end
     else
         s3_resp.obj = conv_to_string ? bytestring(http_resp.data) : http_resp.data
--- a/src/S3.jl
+++ b/src/S3.jl
@@ -634,8 +634,8 @@ function do_request(env::AWSEnv, ro::RO; conv_to_string=true)
     if (http_resp.status > 299)
         if  (search(Base.get(http_resp.headers, "Content-Type", [""]), "/xml") != 0:-1)
             s3_resp.obj = S3Error(LightXML.root(LightXML.parse_string(bytestring(http_resp.data))))
-	else
-	    s3_resp.pd = bytestring(http_resp.data)
+        else
+            s3_resp.pd = bytestring(http_resp.data)
         end
     else
         s3_resp.obj = conv_to_string ? bytestring(http_resp.data) : http_resp.data
--- a/src/S3.jl
+++ b/src/S3.jl
@@ -632,11 +632,8 @@ function do_request(env::AWSEnv, ro::RO; conv_to_string=true)
     s3_resp.headers = http_resp.headers
 
     if (http_resp.status > 299)
-        if  (search(Base.get(http_resp.headers, "Content-Type", [""]), "/xml") != 0:-1) && isa(http_resp.data, IO) && (position(http_resp.data) > 0)
-            s3_resp.pd = LightXML.root(LightXML.parse_string(bytestring(http_resp.data)))
-            if s3_resp.pd.name == "Error"
-                s3_resp.obj = S3Error(s3_resp.pd)
-            end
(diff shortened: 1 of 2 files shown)


amitmurthy:
@mdpradeep could you please review and merge this if OK? I have not looked at the AWS codebase for quite a while now.

src/s3_types.jl
```diff
@@ -602,19 +602,18 @@ CORSConfiguration(pd) = AWS.@parse_vector(AWS.S3.CORSRule, LightXML.get_elements
...
     code::Union{AbstractString, Void}
     message::Union{AbstractString, Void}
-    resource::Union{AbstractString, Void}
     hostId::Union{AbstractString, Void}
     requestId::Union{AbstractString, Void}
 end
 function S3Error(pde)
     code = LightXML.content(LightXML.find_element(pde, "Code"))
     message = LightXML.content(LightXML.find_element(pde, "Message"))
-    resource = LightXML.content(LightXML.find_element(pde, "Resource"))
```
mdpradeep: Is there a reason to remove `resource` ? It could be an attribute that could be passed in the error response as per http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

src/S3.jl
```diff
@@ -632,11 +632,8 @@ function do_request(env::AWSEnv, ro::RO; conv_to_string=true)
...
 
     if (http_resp.status > 299)
-        if  (search(Base.get(http_resp.headers, "Content-Type", [""]), "/xml") != 0:-1) && isa(http_resp.data, IO) && (position(http_resp.data) > 0)
-            s3_resp.pd = LightXML.root(LightXML.parse_string(bytestring(http_resp.data)))
-            if s3_resp.pd.name == "Error"
-                s3_resp.obj = S3Error(s3_resp.pd)
-            end
+        if  (search(Base.get(http_resp.headers, "Content-Type", [""]), "/xml") != 0:-1)
+            s3_resp.obj = S3Error(LightXML.root(LightXML.parse_string(bytestring(http_resp.data))))
 		else
```
amitmurthy: alignment.

src/s3_types.jl
merl-dev: my bad ... however, in my defense, there are other undocumented fields available in the error message  like "SignatureProvided".  My concern was that access keys and id's may leak out through this message, as pointed out in one of the comments of issue #29. I have added these fields in a push I'll make in a few minutes, just testing now.

merl-dev: I think the remark is quite valid and perhaps a search through the entire package and requisite changes are necessary, replacing all keys and id's that may be displayed with something like "xxx" .

src/S3.jl
merl-dev: I didn't find those conditions to be necessary when handling the error.  Perhaps you could point out a scenario when that might be the case and I'll glady realign.

src/s3_types.jl
merl-dev: testing revealed that the field is actually empty when the file name is incorrect, or when the bucket does not exist.  So this mat have been a redundant addition.  Problem is the AWS docs are outdated.

src/S3.jl
amitmurthy: Shouldn't the `else` be aligned with the `if`?

merl-dev: thought you were referring to the deletions I made.  Seems like that's an issue with Atom, where the alignment appears fine on my side.

src/s3_types.jl
mdpradeep: If the value of the field is empty we could have an empty check before assigning the value. But, if the `resource` attribute is removed from the response xml, then this should be fine.

src/S3.jl
amitmurthy: Those are tabs and I think you are seeing it fine since you have a setting of `tabstop=4`. Could you insert spaces instead? That way the code will be aligned for everyone. Your editor should have an option to use spaces everytime you use the tab key.

merl-dev: Thanks for the heads up Amit,  I'm also coding Javascript and R (2 spaces) and I had the Atom settings on auto. Apparently the julia code formatter and/or Atom might not make an auto switch by language type in the buffer.  
The whitespace should be fie now.

amitmurthy:
@mdpradeep good to squash and merge?

merl-dev:
anything missing or can we merge this?

amitmurthy:
@mdpradeep Bump.

amitmurthy:
I am going ahead and merging this.

mdpradeep:
Sorry, never got notified on these for some reason. Thanks @amitmurthy for merging.

Issue #84, opened 2017-05-25
Incorrect format for region specific S3 urls 


rcmcintosh:
For the AWS.S3 module, when a region is specified in the AWS config file, the current code uses an old, no longer valid URL format to build out the URL.

The old format is:
`https://s3-$(env.region).amazonaws.com" * full_path`

The correct format is:
`"https://$(env.region).s3.amazonaws.com" * full_path`

Issue #82, opened 2017-03-14
all of the S3 function do not work


jingpengw:
suddenly, all of the s3 function get an S3Error. Is that due to S3 outage again?
Could anyone else run the `tests3.jl`?

```
S3Response
content_length: 0
date: Tue, 14 Mar 2017 20:32:58 GMT
server: AmazonS3
eTag: 
http_code: 403
delete_marker: false
id_2: 
request_id: 
version_id: 
headers: Pair{AbstractString,AbstractString} Dict
    Transfer-Encoding => chunked
    status_code => 403
    http_major => 1
    x-amz-request-id => **********
    Content-Type => application/xml
    Date => Tue, 14 Mar 2017 20:32:58 GMT
    http_minor => 1
    Keep-Alive => 1
    Server => AmazonS3
    x-amz-id-2 => ****************=
obj: AWS.S3.S3Error("SignatureDoesNotMatch","The request signature we calculated does not match the signature you provided. Check your key and signing method.","********=","********")
```

jingpengwu:
I think I introduced a bug due to my PR. 
https://github.com/JuliaCloud/AWS.jl/pull/73

I just tested my old code, the `tests3.jl` works. Something is not compatible between these two branches.

mdpradeep:
I had tested `tests3.jl` (along with `testec2.jl` and `testsqs.jl`)  before I merged https://github.com/JuliaCloud/AWS.jl/pull/73 and it worked. I had to add the `LocationConstraint` stuff to get `tests3.jl` to work (https://github.com/JuliaCloud/AWS.jl/pull/81)

jingpengwu:
Thanks for the response. I got S3Error for all the operations including `list_bucket`!

jingpengwu:
I used my PR branch frequently, works fine. so it might be introduced by the other PR (https://github.com/JuliaCloud/AWS.jl/pull/79)

tanmaykm:
Should be addressed with #90 now.
Please reopen if this is still an issue for you.

PR #85, opened 2017-05-25
Use correct domain name for s3 url when region is specified

rcmcintosh:
Address #84


jingpengw:
this explains why I can not make s3 working! https://github.com/JuliaCloud/AWS.jl/issues/82

jingpengw:
I tested this PR, still do not work in my workstation!

tanmaykm:
Closing. #84 should be addressed by #90 now.

PR #109, opened 2018-01-24
List many objects

gustafsson:
This PR depends on #108.

Changes:

* `list_objects` can return all objects in a bucket. Not just the first 1000.
* `list_objects` returns a Channel instead of an Array to stream results of long listings.
* `list_objects_details` returns details of a listing, such as timestamp of last modified and etag in `AWS.S3.Contents` objects.


tanmaykm:
lgtm. thanks!

PR #110, opened 2018-01-29
read_iam_credentials if no other credentials are given

gustafsson:
From within an EC2 node we can access temporary AWS credentials through the IAM profile.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -49,6 +49,18 @@ function read_aws_config(configFileName::String = joinpath(homedir(), ".aws/conf
     end
 end
 
+"""Only works from within an EC2 node, blocks until Base.connect times out otherwise"""
+function read_iam_credentials()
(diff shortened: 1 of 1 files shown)


gustafsson:
For reference, I think this is similar to how s3cmd implemented this logic https://github.com/s3tools/s3cmd/blob/333bb792d7488b0d87cfa3e9998be21f7d6f4a40/S3/Config.py#L210

tanmaykm:
Thanks @gustafsson 

But isn't this the same as what has been implemented [here](https://github.com/JuliaCloud/AWS.jl/blob/4ca32795fb3e7d4a98f9b4299963fff9c65192a4/src/AWS.jl#L190-L217)?

gustafsson:
Ups, I missed that

tanmaykm:
okay. will close this as it is already available.

gustafsson:
Thanks

PR #122, opened 2020-04-09
Port AWSCredentials from AWSCore

mattBrzezinski:
## Overview
This merge request is for the port of `AWSCredentials` as part of the re-write. I'm currently happy with the state that this component is in so it is almost a 1 to 1 mapping. There are a couple of changes however;

1) Removal of `LazyJSON` in favour of `JSON`. I don't see a point in using multiple JSON packages, so I removed `LazyJSON`.
2) Since we are still dependent on `AWSCore` at this stage we run into naming conflicts for `AWSCredentials`. You can see the qualifiers [here](https://github.com/JuliaCloud/AWS.jl/blob/c94a154379d5cbb69f3d1d4ff64e1c29829d9d31/test/AWSCredentials.jl#L15) and through the `test/AWSCredentials.jl` package, I have left reminders to remove the qualification afterwards.
3) Since we are using `AWSCore.aws_config()` to create the configuration object, it uses `AWSCore.AWSCredentials` object, there are cases which this breaks tests such as [here](https://github.com/JuliaCloud/AWS.jl/blob/c94a154379d5cbb69f3d1d4ff64e1c29829d9d31/test/AWSCredentials.jl#L240). I have marked these tests as `@test_broken`, and left a comment above. The alternative would be to take these credential objects, and re-created them under the `AWS` module. But since the next merge request will be for creating `AWSConfig`, I figured it's quicker just to do it this way.

## Next Steps
The next merge request will be to create the `AWSConfig` object, being based off of this [proposal](https://github.com/JuliaCloud/AWSCore.jl/pull/63) by Alex.


omus:
> Removal of `LazyJSON` in favour of `JSON`. I don't see a point in using multiple JSON packages, so I removed `LazyJSON`.

I believe LazyJSON was used for performance reasons. It would be good to do some benchmarking.

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
 end
 show(io::IO, e::InvalidFileName) = println(io, e.message)
 
+
+struct AWSException <: Exception
+    code
+    message
+    info
+    cause
+end
```
omus, line 23: Specify types?

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
+    if occursin(r"^application/x-amz-json-1.[01]$", content_type(e))
+        info = JSON.parse(http_message(e))
+        if haskey(info, "__type")
+            code = split(info["__type"], "#")[end]
+        end
+    end
+
+    # Extract API error code from XML error message...
+    if (content_type(e) in ["", "application/xml", "text/xml"]
+    &&  length(http_message(e)) > 0)
```
omus, line 49: Indent or refactor?

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
+    message = "AWSException"
+    info = Dict()
+
+    # Extract API error code from Lambda-style JSON error message...
+    if occursin(r"json$", content_type(e))
+        info = JSON.parse(http_message(e))
+    end
+
+    # Extract API error code from JSON error message...
+    if occursin(r"^application/x-amz-json-1.[01]$", content_type(e))
```
omus, line 40:
```suggestion
    if occursin(r"^application/x-amz-json-1\.[01]$", content_type(e))
```

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
 
+
+struct AWSException <: Exception
+    code
+    message
+    info
+    cause
+end
+show(io::IO, e::AWSException) = println(io, string(e.code,
+                       e.message == "" ? "" : (" -- " * e.message), "\n",
```
omus, line 25: Just use a multi-line function
```suggestion
                       isempty(e.message) ? "" : (" -- " * e.message), "\n",
```

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
+
+    info = get(info, "Errors", info)
+    info = get(info, "Error", info)
+    code = get(info, "Code", code)
+    message = get(info, "Message", message)
+    message = get(info, "message", message)
+
+    AWSException(code, message, info, e)
+end
 end
```
omus, line 61:
```suggestion

end
```

Project.toml
```diff
@@ -5,9 +5,13 @@ version = "1.0.0"
 
 [deps]
 AWSCore = "4f1ea46c-232b-54a6-9b17-cc2d0f3e6598"
+Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+IniFile = "83e8ac13-25f8-5344-8a64-a9f2b223428f"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
+Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
+Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
```
omus, line 14: Add compat entries for all of these

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
+using AWSCore: Services
+using Base
```
nicoleepp, line 2: What are we using from Base? 

Could we maybe try specifying it out

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
...
+"""
+When you interact with AWS, you specify your [AWS Security Credentials](http://docs.aws.amazon.com/general/latest/gr/aws-security-credentials.html)
+to verify who you are and whether you have permission to access the resources that you are requesting.
+AWS uses the security credentials to authenticate and authorize your requests.
+The fields `access_key_id` and `secret_key` hold the access keys used to authenticate API requests
+(see [Creating, Modifying, and Viewing Access Keys](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#Using_CreateAccessKey)).
+[Temporary Security Credentials](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html) require the extra session `token` field.
+The `user_arn` and `account_number` fields are used to cache the result of the [`aws_user_arn`](@ref) and [`aws_account_number`](@ref) functions.
+
+AWSCore searches for credentials in a series of possible locations and stop as soon as it finds credentials.
```
nicoleepp, line 36:
```suggestion
AWSCore searches for credentials in a series of possible locations and stops as soon as it finds credentials.
```

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
...
+The order of precedence for this search is as follows:
+
+1. Passing credentials directly to the `AWSCredentials` constructor
+2. [Environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html)
+3. Shared credential file [(~/.aws/credentials)](http://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html)
+4. AWS config file [(~/.aws/config)](http://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html)
+5. Assume Role provider via the aws config file
+6. Instance metadata service on an Amazon EC2 instance that has an IAM role configured.
```
nicoleepp, line 44:
```suggestion
6. Instance metadata service on an Amazon EC2 instance that has an IAM role configured
```

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
...
+    3. ~/.aws/config
+    4. EC2 or ECS metadata
+
+# Keywords
+- `profile::AbstractString`: Specific profile used to search for AWSCredentials
+
+# Throws
+- `error("Can't find AWS Credentials")`: AWSCredentials could not be found
+"""
+function AWSCredentials(;profile=nothing)
```
nicoleepp, line 126:
```suggestion
function AWSCredentials(; profile=nothing)
```

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
...
+
+dot_aws_config_file() = get(ENV, "AWS_CONFIG_FILE", joinpath(homedir(), ".aws", "config"))
+dot_aws_credentials_file() = get(ENV, "AWS_SHARED_CREDENTIALS_FILE", joinpath(homedir(), ".aws", "credentials"))
+localhost_maybe_ec2() = localhost_is_ec2() || isfile("/sys/devices/virtual/dmi/id/product_uuid")
+localhost_is_lambda() = haskey(ENV, "LAMBDA_TASK_ROOT")
+_aws_get_profile() = get(ENV, "AWS_DEFAULT_PROFILE", get(ENV, "AWS_PROFILE", "default"))
+
+
+"""
+    AWSCredentials(;profile=nothing) -> Union{AWSCredentials, Nothing}
```
nicoleepp, line 109:
```suggestion
    AWSCredentials(; profile=nothing) -> Union{AWSCredentials, Nothing}
```

src/AWSCredentials.jl
```diff
@@ -0,0 +1,612 @@
...
+end
+
+
+"""
+    ec2_instance_credentials() -> AWSCredentials
+
+Parse the EC2 metadata to retrieve AWSCredentials.
+"""
+function ec2_instance_credentials()
+    info  = _ec2_metadata("iam/info")
```
nicoleepp, line 300: Extra spaces before the `=` here and below?

mattBrzezinski: Absolutely not, good 👀

Project.toml
nicoleepp: Still need to add compat entry for Mocking

src/AWSCredentials.jl
```diff
@@ -0,0 +1,611 @@
...
+    source_profile === nothing && return nothing
+    credentials = nothing
+
+    for f in [dot_aws_credentials, dot_aws_config]
+        credentials = f(source_profile)
+        credentials === nothing || break
+    end
+
+    credentials === nothing && return nothing
+    config = AWSConfig(:creds=>credentials, :region=>aws_get_region(source_profile, ini))
```
nicoleepp, line 550:
```suggestion
    config = AWSConfig(:creds => credentials, :region => aws_get_region(source_profile, ini))
```

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
 
+
+struct AWSException <: Exception
+    code::String
+    message::String
+    info::Union{String, Nothing}
+    cause::HTTP.StatusError
+end
+show(io::IO, e::AWSException) = println(io, string(e.code,
+                       isempty(e.message) ? "" : (" -- " * e.message), "\n", e.cause))
```
nicoleepp, line 25: Should be a multi-line function

src/AWSExceptions.jl
```diff
@@ -12,4 +14,48 @@ struct InvalidFileName <: Exception
...
+
+
+function AWSException(e::HTTP.StatusError)
+    code = string(http_status(e))
+    message = "AWSException"
+    info = Dict()
```
nicoleepp, line 31: Add types of the dict elements

info = Dict{String, Dict}?

test/AWSCredentials.jl
```diff
@@ -0,0 +1,549 @@
+macro test_ecode(error_codes, expr)
+    quote
+        try
+            $expr
+            @test false
+        catch e
+            if e isa AWSException
+                ecode(e) in [$error_codes;]
+            end
```
nicoleepp, line 9: Should we rethrow the error if it wasn't an AWSException?

mattBrzezinski:
> > Removal of `LazyJSON` in favour of `JSON`. I don't see a point in using multiple JSON packages, so I removed `LazyJSON`.
> 
> I believe LazyJSON was used for performance reasons. It would be good to do some benchmarking.

Below is a benchmark comparison between `JSON.jl` and `LazyJSON.jl` when loading the `deps/metadata.json` file.

Time Comparison:
| Package | Average | Median | Run 1 | Run 2 | Run 3 | Run 4 | Run 5 |
|---|---|---|---|---|---|---|---|
| LazyJSON.jl | 0.041143 | 0.04072 | 0.044690 | .038206 | 0.041633 | 0.040720 | 0.040466 |
| JSON.jl | 0.4899396 | 0.487541 | 0.499561 | 0.472634 | 0.487541 | 0.482708 | 0.507254 |

`LazyJSON.jl` is **0.4487966 seconds faster** on average compared to `JSON.jl` Memory is constant between all runs with `LazyJSON.jl` using `5.805 MiB` and `JSON.jl` using `10.281 MiB`.

I'll create a new commit moving from `JSON.jl` to `LazyJSON.jl` across this repo. The only downside to `LazyJSON.jl` is that it is not well maintained, but we can take it under our wing while we're at it.

mattBrzezinski:
I spent the last little while trying to convert from using `JSON` to `LazyJSON`, and at this time I don't think it's worth putting in the effort.

`LazyJSON` does not support a lot of functionality which `JSON` does, such as `setindex!()`, `get()`, etc. and it is slowly evolving into re-writing a lot of the code which is already in `v1`. I can see now why `AWSCore.jl` is using a combination of these packages.

src/AWSCredentials.jl
```diff
@@ -0,0 +1,611 @@
...
+        return nothing
+    end
+
+    uri = ENV["AWS_CONTAINER_CREDENTIALS_RELATIVE_URI"]
+
+    response = @mock http_get("http://169.254.170.2$uri")
+    new_creds = String(response.body)
+    new_creds = JSON.parse(new_creds)
+
+    expiry = DateTime(strip(new_creds["Expiration"], 'Z'))
```
omus, line 347: Doesn't really matter:
```suggestion
    expiry = DateTime(rstrip(new_creds["Expiration"], 'Z'))
```

src/AWSExceptions.jl
```diff
@@ -12,4 +14,50 @@ struct InvalidFileName <: Exception
...
+    if occursin(r"^application/x-amz-json-1\.[01]$", content_type(e))
+        info = JSON.parse(http_message(e))
+        if haskey(info, "__type")
+            code = split(info["__type"], "#")[end]
+        end
+    end
+
+    # Extract API error code from XML error message...
+    error_content_types = ["", "application/xml", "text/xml"]
+    if (content_type(e) in error_content_types && length(http_message(e)) > 0)
```
omus, line 50:
```suggestion
    if content_type(e) in error_content_types && length(http_message(e)) > 0
```

src/AWSExceptions.jl
```diff
@@ -12,4 +14,50 @@ struct InvalidFileName <: Exception
...
+        info = parse_xml(http_message(e))
+    end
+
+    info = get(info, "Errors", info)
+    info = get(info, "Error", info)
```
omus, line 55: This code is a little bizarre. Right now all of these would unpack in the same way:

``` julia
info = Dict("Errors" => 1, "Error", => 2)
info = Dict("Errors" => Dict("Error", => 2))
info = Dict("Error", => 2)
```
What is the structure you want to unpack?

Project.toml
```diff
@@ -5,16 +5,22 @@ version = "1.0.0"
...
 SymDict = "2da68c74-98d7-5633-99d6-8493888d7b1e"
 
 [compat]
 AWSCore = "0.6"
 HTTP = "0.8"
+IniFile = "0.5"
 JSON = "0.18, 0.19, 0.20, 0.21"
```
omus, line 21:
```suggestion
JSON = "0.18, 0.19, 0.20, 0.21"
Mocking = "0.7"
```

src/AWSExceptions.jl
mattBrzezinski: This is just ported over from `AWSCore.jl`, to me it looks like AWS is returning either one or multiple errors in their response and this is just handling both of those cases.

I'll play around with this more but I believe that AWS will only return back `Errors` or `Error` but not both. I'll write some more tests to capture all these different cases.

mattBrzezinski: I updated this to have an `XML` and `JSON` request testing, I've spent some time looking around but I can't seem to find any examples online of other request errors that AWS throws back and their structures.

I'm a bit wary of updating this since it is working currently in `AWSCore.jl` and no one has logged an issue about it. Seems like this knowledge has been lost a bit. :(

omus: Ok, I guess leave it as is for now then

PR #124, opened 2020-06-22
Properly escape parameters in REST definitions

mattBrzezinski:
## Notes

**When reviewing only look at this [commit](https://github.com/JuliaCloud/AWS.jl/pull/124/commits/8e24d7ff2104bb86ea696537b58904b02ce2a8ab), to avoid the re-generated definitions.**

This is going into `v1` not `master`. I also just realized that some service requests have required parameters in the `querystring` not the `uri` so some definitions do NOT make sense, e.g.:

```julia
CreateAnalyzer(analyzerName, type) = accessanalyzer("PUT", "/analyzer")
```

## Proposed Changes

There are two functional changes with this PR:
1) Properly escape parameters when passing them along to the low-level definition

```julia
DeleteAnalyzer(args) = accessanalyzer("DELETE", "/analyzer/{analyzerName}", args)
```

This change will instead will properly escape the parameter:
```julia
DeleteAnalyzer(analyzerName) = accessanalyzer("DELETE", "/analyzer/$(analyzerName)")
```

2) Properly define `REST JSON` high-level wrappers to mimic `REST XML` as functionality wise they are the same.


src/AWSMetadataUtilities.jl
```diff
@@ -348,17 +348,12 @@ function _generate_high_level_definition(
...
-        if !isempty(required_parameters)
-            operation_definition *= "\n$name(args) = $service_name(\"$method\", \"$request_uri\", args)\n"
-        else
-            operation_definition *= """\n
-            $name() = $service_name(\"$method\", \"$request_uri\")
-            $name(args) = $service_name(\"$method\", \"$request_uri\", args)
-            """
-        end
-    elseif protocol == "rest-xml"
+    elseif protocol in ["rest-json", "rest-xml"]
```
iamed2, line 351:
```suggestion
    elseif protocol in ("rest-json", "rest-xml")
```
prefer tuples for small inline collections as it's easier on the compiler

iamed2: It won't have any noticeable effect here but I thought I'd note it since it's generally good practice

PR #126, opened 2020-06-22
Added REST JSON request functionality

mattBrzezinski:
## TODO Before merge
- [x] Don't hard-code the glacier API version, instead pull it from the `aws-sdk-js`

## Notes
- This is merging into `v1` feature branch not `master`
- I am not including high/low level generated API definitions as I feel this is overwhelming for people reviewing the code

## Changes
This PR ports over the `REST JSON` request functionality. The other main change is to `RestJSONService` structs now have an additional field called `service_specifics`. This has been added in because certain AWS services require special casing, using AWS Glacier we have a few options on how to handle these cases.

### Problem
AWS Glacier requires the `x-amz-glacier-version=2012-06-01` in every request header.

### Solutions

1) Do nothing, the end user will always need to make a request passing along these custom headers with every request.

Easy for us since we don't have to do anything. But the users experience is quite poor having to handle these cases themselves.

2) Write `if` checks when making a `REST-XML` call, and append these headers.

Better than the first solution, however as this list of special cases grow our functor grows in complexity with `if` checks, e.g.

```julia
function (service::RestJSONService)(...)
    ...
    if service == "glacier"
        # append specific headers
    elseif service == "foo"
        # append specific headers
    end
end
```

We can do things a bit better than that.

3) Have a `service_specifics` field on the `struct`

This allows us to handle our special cases when generating the low-level API definitions so we can end up with something clean like:

```julia
const glacier = AWS.RestJSONService("glacier", "2012-06-01", Dict(:headers => Dict("x-amz-glacier-version" => "2012-06-01")))
```

And then in our functor, we just merge the `service_specifics[:headers]` to handle this special case. Or whatever else it might be.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -66,7 +66,7 @@ end
 Base.@kwdef mutable struct Request
     service::String
     api_version::String
-    request_method::String  # Enum?
+    request_method::String
 
     headers::AbstractDict{String, String}=LittleDict{String, String}()
     content::String=""
@@ -77,7 +77,26 @@ Base.@kwdef mutable struct Request
     response_stream::Union{Base.BufferStream, Nothing}=nothing
     http_options::Array=[]
     return_raw::Bool=false
-    ordered_json_dict::Bool=true
+    response_dict_type::Type=LittleDict
+
+    function Request(
+        service::String,
+        api_version::String,
+        request_method::String,
+
+        headers::AbstractDict{String, String},
+        content::String,
+        resource::String,
+        url::String,
+
+        return_stream::Bool,
+        response_stream::Union{Base.BufferStream, Nothing},
+        http_options::Array,
+        return_raw::Bool,
+        response_dict_type::Type{T},
+    ) where T <: AbstractDict
+        return new(service, api_version, request_method, headers, content, resource, url, return_stream, response_stream, http_options, return_raw, response_dict_type)
+    end
 end
 
 # Needs to be included after the definition of struct otherwise it cannot find them
@@ -292,9 +311,11 @@ function do_request(aws::AWSConfig, request::Request; return_headers::Bool=false
         end
     end
 
+    response_dict_type = request.response_dict_type
+
     # For HEAD request, return headers...
     if request.request_method == "HEAD"
-        return Dict(response.headers)
+        return response_dict_type(response.headers)
     end
 
     # Return response stream if requested...
@@ -318,9 +339,9 @@ function do_request(aws::AWSConfig, request::Request; return_headers::Bool=false
     body = String(copy(response.body))
 
     if occursin(r"/xml", mime)
-        xml_dict_type = OrderedDict{Union{Symbol, String}, Any}
+        xml_dict_type = response_dict_type{Union{Symbol, String}, Any}
 
-        return (return_headers ? (xml_dict(body, xml_dict_type), Dict(response.headers)) : xml_dict(body, xml_dict_type))
+        return (return_headers ? (xml_dict(body, xml_dict_type), response_dict_type(response.headers)) : xml_dict(body, xml_dict_type))
     end
 
     if endswith(mime, "json")
@@ -328,9 +349,9 @@ function do_request(aws::AWSConfig, request::Request; return_headers::Bool=false
             return (return_headers ? (nothing, response.headers) : nothing)
         end
 
-        info = request.ordered_json_dict ? JSON.parse(body, dicttype=OrderedDict) : JSON.parse(body)
+        info = JSON.parse(body, dicttype=response_dict_type)
 
-        return (return_headers ? (info, Dict(response.headers)) : info)
+        return (return_headers ? (info, response_dict_type(response.headers)) : info)
     end
 
     # Return raw data by default...
@@ -370,6 +391,15 @@ function _return_headers(args::AbstractDict{String, <:Any})
     return return_headers
 end
 
+function _set_response_dict_type!(request::Request, args::AbstractDict{String, <:Any})
+    response_dict_type = get(args, "response_dict_type", "")
+    response_dict_type = eval(Meta.parse(response_dict_type))
+
+    if response_dict_type != nothing
+        request.response_dict_type = response_dict_type
+    end
+end
+
 """
     (service::RestXMLService)(
         aws::AWSConfig, request_method::String, request_uri::String, args=[];
@@ -401,8 +431,9 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
         http_options=get(args, "http_options", []),
         response_stream=get(args, "response_stream", nothing),
         return_raw=get(args, "return_raw", false),
-        ordered_json_dict=get(args, "ordered_json_dict", true),
     )
+
+    _set_response_dict_type!(request, args)
 
     delete!(args, "headers")
     delete!(args, "body")
@@ -469,9 +500,9 @@ function (service::RestJSONService)(
         http_options=get(args, "http_options", []),
         response_stream=get(args, "response_stream", nothing),
         return_raw=get(args, "return_raw", false),
-        ordered_json_dict=get(args, "ordered_json_dict", true),
     )
 
+    _set_response_dict_type!(request, args)
     request.url = _generate_service_url(aws.region, request.service, request.resource)
 
     if !isempty(service.service_specific_headers)
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -58,10 +58,10 @@ struct RestJSONService
     name::String
     api_version::String
 
-    service_specifics::LittleDict{String, AbstractDict}
+    service_specific_headers::LittleDict{String, String}
 
-    RestJSONService(name::String, api_version::String) = new(name, api_version, Dict())
-    RestJSONService(name::String, api_version::String, service_specifics::LittleDict{String, <:AbstractDict}) = new(name, api_version, service_specifics)
+    RestJSONService(name::String, api_version::String) = new(name, api_version, LittleDict())
+    RestJSONService(name::String, api_version::String, service_specific_headers::LittleDict{String, String}) = new(name, api_version, service_specific_headers)
 end
 
 # Needs to be included after the definition of struct otherwise it cannot find them
@@ -454,8 +454,8 @@ function (service::RestJSONService)(aws::AWS.AWSConfig, request_method::String, 
 
     request[:headers] = Dict{String, String}(get(args, "headers", []))
 
-    if haskey(service.service_specifics, "headers")
-        request[:headers] = merge(request[:headers], service.service_specifics["headers"])
+    if !isempty(service.service_specific_headers)
+        request[:headers] = merge(request[:headers], service.service_specific_headers)
     end
 
     request[:headers]["Content-Type"] = "application/json"
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -8,12 +8,9 @@ using Mocking
 using OrderedCollections: LittleDict, OrderedDict
 using Retry
 using Sockets
-using SymDict
 using XMLDict
 
-export @service, AWSServices, RestJSONService, JSONService, RestXMLService, QueryService, AWSConfig, AWSExceptions, set_user_agent
-
-const AWSRequest = SymbolDict
+export @service, AWSServices, RestJSONService, JSONService, RestXMLService, QueryService, AWSConfig, AWSExceptions, set_user_agent, Request
 
 include("AWSCredentials.jl")
 include("AWSConfig.jl")
@@ -64,31 +61,48 @@ struct RestJSONService
     RestJSONService(name::String, api_version::String, service_specific_headers::LittleDict{String, String}) = new(name, api_version, service_specific_headers)
 end
 
+Base.@kwdef mutable struct Request
+    service::String
+    api_version::String
+    request_method::String  # Enum?
+
+    headers::AbstractDict{String, String}=LittleDict{String, String}()
+    content::String=""
+    resource::String=""
+    url::String=""
+
+    return_stream::Bool=false
+    response_stream::Union{Base.BufferStream, Nothing}=nothing
+    http_options::Array=[]
+    return_raw::Bool=false
+    ordered_json_dict::Bool=true
+end
+
 # Needs to be included after the definition of struct otherwise it cannot find them
 include("AWSServices.jl")
 
 http_status(e::HTTP.StatusError) = e.status
 header(e::HTTP.StatusError, k, d="") = HTTP.header(e.response, k, d)
 
-function _sign!(aws::AWSConfig, request::LittleDict; time::DateTime=now(Dates.UTC))
-    if request[:service] in ["sdb", "importexport"]
+function _sign!(aws::AWSConfig, request::Request; time::DateTime=now(Dates.UTC))
+    if request.service in ["sdb", "importexport"]
         _sign_aws2!(aws, request, time)
     else
         _sign_aws4!(aws, request, time)
     end
 end
 
-function _sign_aws2!(aws::AWSConfig, request::LittleDict, time::DateTime)
+function _sign_aws2!(aws::AWSConfig, request::Request, time::DateTime)
     # Create AWS Signature Version 2 Authentication query parameters.
     # http://docs.aws.amazon.com/general/latest/gr/signature-version-2.html
 
     query = Dict{String, String}()
-    for elem in split(request[:content], '&', keepempty=false)
+    for elem in split(request.content, '&', keepempty=false)
         (n, v) = split(elem, "=")
         query[n] = HTTP.unescapeuri(v)
     end
 
-    request[:headers]["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
+    request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
 
     creds = check_credentials(aws.credentials)
     query["AWSAccessKeyId"] = creds.access_key_id
@@ -101,16 +115,16 @@ function _sign_aws2!(aws::AWSConfig, request::LittleDict, time::DateTime)
     end
 
     query = Pair[k => query[k] for k in sort(collect(keys(query)))]
-    uri = HTTP.URI(request[:url])
+    uri = HTTP.URI(request.url)
     to_sign = "POST\n$(uri.host)\n$(uri.path)\n$(HTTP.escapeuri(query))"
     push!(query, "Signature" => digest(MD_SHA256, to_sign, creds.secret_key) |> base64encode |> strip)
 
-    request[:content] = HTTP.escapeuri(query)
+    request.content = HTTP.escapeuri(query)
 
     return request
 end
 
-function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
+function _sign_aws4!(aws::AWSConfig, request::Request, time::DateTime)
     # Create AWS Signature Version 4 Authentication Headers.
     # http://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
 
@@ -118,7 +132,7 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
     datetime = Dates.format(time, dateformat"yyyymmdd\THHMMSS\Z")
 
     # Authentication scope...
-    authentication_scope = [date, aws.region, request[:service], "aws4_request"]
+    authentication_scope = [date, aws.region, request.service, "aws4_request"]
 
     creds = check_credentials(aws.credentials)
     signing_key = "AWS4$(creds.secret_key)"
@@ -131,33 +145,34 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
     authentication_scope = join(authentication_scope, "/")
 
     # SHA256 hash of content...
-    content_hash = bytes2hex(digest(MD_SHA256, request[:content]))
+    content_hash = bytes2hex(digest(MD_SHA256, request.content))
 
     # HTTP headers...
-    delete!(request[:headers], "Authorization")
-    merge!(request[:headers], Dict(
+    delete!(request.headers, "Authorization")
+
+    merge!(request.headers, Dict(
         "x-amz-content-sha256" => content_hash,
         "x-amz-date" => datetime,
-        "Content-MD5" => base64encode(digest(MD_MD5, request[:content]))
+        "Content-MD5" => base64encode(digest(MD_MD5, request.content))
     ))
 
     if !isempty(creds.token)
-        request[:headers]["x-amz-security-token"] = creds.token
+        request.headers["x-amz-security-token"] = creds.token
     end
 
     # Sort and lowercase() Headers to produce canonical form...
-    canonical_headers = ["$(lowercase(k)):$(strip(v))" for (k,v) in request[:headers]]
-    signed_headers = join(sort([lowercase(k) for k in keys(request[:headers])]), ";")
+    canonical_headers = ["$(lowercase(k)):$(strip(v))" for (k,v) in request.headers]
+    signed_headers = join(sort([lowercase(k) for k in keys(request.headers)]), ";")
 
     # Sort Query String...
-    uri = HTTP.URI(request[:url])
+    uri = HTTP.URI(request.url)
     query = HTTP.URIs.queryparams(uri.query)
     query = Pair[k => query[k] for k in sort(collect(keys(query)))]
 
     # Create hash of canonical request...
     canonical_form = string(
-        request[:request_method], "\n",
-        request[:service] == "s3" ? uri.path : HTTP.escapepath(uri.path), "\n",
+        request.request_method, "\n",
+        request.service == "s3" ? uri.path : HTTP.escapepath(uri.path), "\n",
         HTTP.escapeuri(query), "\n",
         join(sort(canonical_headers), "\n"), "\n\n",
         signed_headers, "\n",
@@ -171,7 +186,7 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
     signature = bytes2hex(digest(MD_SHA256, string_to_sign, signing_key))
 
     # Append Authorization header...
-    request[:headers]["Authorization"] = string(
+    request.headers["Authorization"] = string(
         "AWS4-HMAC-SHA256 ",
         "Credential=$(creds.access_key_id)/$authentication_scope, ",
         "SignedHeaders=$signed_headers, ",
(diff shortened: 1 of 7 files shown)


src/AWSMetadataUtilities.jl
```diff
@@ -134,11 +134,19 @@ function _generate_low_level_definition(service::Dict)
     service_id = replace(lowercase(service["serviceId"]), ' ' => '_')
     api_version = service["apiVersion"]
 
+    service_specifics = Dict(
+        "glacier" => "Dict(:headers => Dict(\"x-amz-glacier-version\" => \"2012-06-01\"))"
```
mattBrzezinski, line 138: I am unsure how to handle the hardcoding of this API version here, any suggestions?

src/AWS.jl
```diff
@@ -307,11 +312,29 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
         return (return_headers ? (xml_dict(body, xml_dict_type), Dict(response.headers)) : xml_dict(body, xml_dict_type))
     end
 
+    if occursin(r"json$", mime)
+        if isempty(response.body)
+            return nothing
```
iamed2, line 317: I know this exists in AWSCore, but why not also return the headers when `return_headers`?

src/AWS.jl
```diff
@@ -57,6 +57,11 @@ end
 struct RestJSONService
     name::String
     api_version::String
+
+    service_specifics::Dict{Symbol, Any}
```
iamed2, line 61: We only use headers right now; could this just be `service_specific_headers::Dict{String, String}`? Or even use a `LittleDict{String, String}`.

src/AWS.jl
```diff
@@ -335,7 +358,7 @@ function _generate_service_url(region::String, request::AbstractDict)
     return string("https://", endpoint, ".", isempty(region) ? "" : "$region.", SERVICE_HOST, request[:resource])
 end
 
-function _return_headers(args::Dict{String, Any})
+function _return_headers(args::AbstractDict)
```
iamed2, line 361: Why was this loosened?

src/AWSMetadataUtilities.jl
iamed2: I assume you posted this before deciding to pull the version out of `service["apiVersion"]`?

src/AWS.jl
```diff
@@ -307,11 +312,29 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
...
     end
 
+    if occursin(r"json$", mime)
+        if isempty(response.body)
+            return nothing
+        end
+
+        info = (get(request, :ordered_json_dict, true) ? JSON.parse(body, dicttype=OrderedDict) : JSON.parse(body))
+
+        @protected try
```
iamed2, line 322: When is this used? This seems excessively clever. Given that we're making a new version maybe we can dump it?

Notably, the XML API path does not seem to filter out the similar wrapper tags. I also cannot find this happening in other SDKs, but perhaps I did not look hard enough.

src/AWS.jl
mattBrzezinski: Originally I thought it didn't matter too much if it was an `AbstractDict`, but I guess we're not exporting this function so it can stay typed.

src/AWS.jl
mattBrzezinski: I just ported it over, I wasn't able to find any use cases myself but figured it was added in for a reason.

I guess this is a perfect example of why comments are good, so knowledge isn't lost in the void. Oh well, if this becomes problematic we can re-introduce it back in.

src/AWSServices.jl
```diff
@@ -94,13 +96,14 @@ const forecastquery = AWS.JSONService("forecastquery", "2018-06-26", "1.1", "Ama
...
 const gamelift = AWS.JSONService("gamelift", "2015-10-01", "1.1", "GameLift")
-const glacier = AWS.RestJSONService("glacier", "2012-06-01")
+const glacier = AWS.RestJSONService("glacier", "2012-06-01", LittleDict("headers" => LittleDict("x-amz-glacier-version" => "2012-06-01")))
 const global_accelerator = AWS.JSONService("globalaccelerator", "2018-08-08", "1.1", "GlobalAccelerator_V20180706")
 const glue = AWS.JSONService("glue", "2017-03-31", "1.1", "AWSGlue")
 const greengrass = AWS.RestJSONService("greengrass", "2017-06-07")
 const groundstation = AWS.RestJSONService("groundstation", "2019-05-23")
 const guardduty = AWS.RestJSONService("guardduty", "2017-11-28")
 const health = AWS.JSONService("health", "2016-08-04", "1.1", "AWSHealth_20160804")
+const honeycode = AWS.RestJSONService("honeycode", "2020-03-01")
```
mattBrzezinski, line 106: `Honeycode` just launched two days ago, while updating this PR it brought along this line. Nbd, it can just be left in here since the high-level stuff will get generated for it at some point before launch.

src/AWS.jl
```diff
@@ -57,6 +57,11 @@ end
 struct RestJSONService
     name::String
     api_version::String
+
+    service_specifics::LittleDict{String, AbstractDict}
```
iamed2, line 61: We only use headers right now; could this just be `service_specific_headers::LittleDict{String, String}`?

mattBrzezinski: Yes, that makes sense. I can update this in a couple.

mattBrzezinski:
When you get a chance to re-view this let me know if there are any additional changes you'd like to see @iamed2

src/AWS.jl
```diff
@@ -454,8 +454,8 @@ function (service::RestJSONService)(aws::AWS.AWSConfig, request_method::String,
 
     request[:headers] = Dict{String, String}(get(args, "headers", []))
 
-    if haskey(service.service_specifics, "headers")
-        request[:headers] = merge(request[:headers], service.service_specifics["headers"])
+    if !isempty(service.service_specific_headers)
+        request[:headers] = merge(request[:headers], service.service_specific_headers)
```
iamed2, line 458:
```suggestion
        merge!(request[:headers], service.service_specific_headers)
```

src/AWS.jl
```diff
@@ -427,16 +441,31 @@ end
...
 function (service::RestJSONService)(aws::AWS.AWSConfig, request_method::String, request_uri::String, args=[])
-    return AWSCore.service_rest_json(
-        aws;
-        service=service.name,
-        version=service.api_version,
-        verb=request_method,
-        resource=request_uri,
-        args=args
-    )
+    args = stringdict(args)
```
iamed2, line 444: Why do we use String keys sometimes and Symbol keys other times? And LittleDict sometimes but Dict others? AbstractDict is also used sometimes when later code requires something more specific. And key and value types are not consistently specified. 

The next thing after this PR should be standardizing dict types. In general, dicts created internally which expect to have fewer than 32 items should be LittleDicts, and key+value types should be prescribed wherever possible. We should also restrict the `args` interface for services to something more specific to avoid the AWSCore hassle we've been going through. 

I also think we should drop `SymDict`.

src/AWS.jl
```diff
@@ -307,11 +312,21 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
...
+
+        return (return_headers ? (info, Dict(response.headers)) : info)
+    end
+
     # Return raw data by default...
     return (return_headers ? (response.body, response.headers) : response.body)
 end
 
-function _generate_rest_resource(request_uri::String, args::Dict{String, Any})
+function _generate_rest_resource(request_uri::String, args::AbstractDict)
```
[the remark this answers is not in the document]
mattBrzezinski: Yeah I wasn't sure if this should be an `AbstractDict` or not, since it's an internal function. But I'll change these up!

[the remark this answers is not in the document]
mattBrzezinski: I believe we need to do `request::AbstractDict{Symbol, <:Any}`. Here's a quick example,

```julia
f(d::AbstractDict{String, Any}) = println(d)
f2(d::AbstractDict{String, <:Any}) = println(d)

d = Dict("foo"=>"bar")
d2 = Dict{String, Any}("foo"=>"bar")

f(d)
> ERROR: MethodError: no method matching f(::Dict{String,String})

f(d2)
> Dict{String,Any}("foo" => "bar")

f2(d)
> Dict("foo" => "bar")

f2(d2)
> Dict{String,Any}("foo" => "bar")
```

test/AWS.jl
```diff
@@ -376,20 +376,21 @@ end
...
         max_keys = 1
-        result = AWSServices.s3("GET", "/$bucket_name"; max_keys=max_keys)
+        result = AWSServices.s3("GET", "/$bucket_name", Dict("max_keys"=>max_keys))
 
         @test max_keys == length([result["ListBucketResult"]["Contents"]])
     end
 
     @testset "POST - w/ Params" begin
-        AWSServices.s3("POST", "/$bucket_name?delete"; Body="""
+        AWSServices.s3("POST", "/$bucket_name?delete", Dict("Body"=>
```
mattBrzezinski, line 385: Bit unsure of the BlueStyle formatting here.

src/AWS.jl
```diff
@@ -8,12 +8,9 @@ using Mocking
 using OrderedCollections: LittleDict, OrderedDict
 using Retry
 using Sockets
-using SymDict
 using XMLDict
 
-export @service, AWSServices, RestJSONService, JSONService, RestXMLService, QueryService, AWSConfig, AWSExceptions, set_user_agent
-
-const AWSRequest = SymbolDict
+export @service, AWSServices, RestJSONService, JSONService, RestXMLService, QueryService, AWSConfig, AWSExceptions, set_user_agent, Request
```
omus, line 13: I'd group the types together but I can see already the exports are kinda chaotic

src/AWS.jl
```diff
@@ -57,6 +54,28 @@ end
 struct RestJSONService
     name::String
     api_version::String
+
+    service_specific_headers::LittleDict{String, String}
+
+    RestJSONService(name::String, api_version::String) = new(name, api_version, LittleDict())
+    RestJSONService(name::String, api_version::String, service_specific_headers::LittleDict{String, String}) = new(name, api_version, service_specific_headers)
```
omus, line 61: As these really don't do any enforcement I would make these outer constructors instead of inner

omus: Additionally:
```julia
function RestJSONService(name::String, api_version::String, service_specific_headers::LittleDict{String, String}=LittleDict{String,String}())
    RestJSONService(name, api_version, service_specific_headers)
end
```
Defines both methods

src/AWS.jl
```diff
@@ -57,6 +54,28 @@ end
...
+    headers::AbstractDict{String, String}=LittleDict{String, String}()
+    content::String=""
+    resource::String=""
+    url::String=""
+
+    return_stream::Bool=false
+    response_stream::Union{Base.BufferStream, Nothing}=nothing
+    http_options::Array=[]
+    return_raw::Bool=false
+    ordered_json_dict::Bool=true
```
omus, line 78: Maybe just store the type of the output dictionary if this can be done generically

src/AWS.jl
```diff
@@ -126,33 +145,34 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
...
     if !isempty(creds.token)
-        request[:headers]["x-amz-security-token"] = creds.token
+        request.headers["x-amz-security-token"] = creds.token
     end
 
     # Sort and lowercase() Headers to produce canonical form...
-    canonical_headers = ["$(lowercase(k)):$(strip(v))" for (k,v) in request[:headers]]
-    signed_headers = join(sort([lowercase(k) for k in keys(request[:headers])]), ";")
+    canonical_headers = ["$(lowercase(k)):$(strip(v))" for (k,v) in request.headers]
+    signed_headers = join(sort([lowercase(k) for k in keys(request.headers)]), ";")
```
omus, line 165: Functions for the `Request` type?

src/AWS.jl
```diff
@@ -176,17 +196,17 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
...
         io = Base.BufferStream()
-        request[:response_stream] = io
-        push!(options, (:response_stream, io))
+        request.response_stream = io
+        push!(options, ("response_stream", io))
     end
 
-    if haskey(request, :http_options)
-        for option in request[:http_options]
+    if !isempty(request.http_options)
```
omus, line 208: Can safely just remove this check now as the `for` will take care of it

src/AWS.jl
```diff
@@ -233,27 +253,27 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
         "PriorRequestNotComplete"
     ]
 
-    if !haskey(request, :headers)
-        request[:headers] = Dict{String, String}()
+    if isempty(request.headers)
+        request.headers = LittleDict{String, String}()
     end
```
omus, line 258: I think this is unnecessary now

src/AWS.jl
```diff
@@ -233,27 +253,27 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
...
+            request.url = HTTP.header(response, "Location")
             continue
         end
     catch e
         if e isa HTTP.StatusError
             e = AWSException(e)
         end
 
-        @retry if :message in fieldnames(typeof(e)) && occursin("Signature expired", e.message) end
+        @retry if "message" in fieldnames(typeof(e)) && occursin("Signature expired", e.message) end
```
omus, line 276: I think this was updated by mistake

src/AWS.jl
```diff
@@ -307,11 +327,21 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
         return (return_headers ? (xml_dict(body, xml_dict_type), Dict(response.headers)) : xml_dict(body, xml_dict_type))
     end
 
+    if occursin(r"json$", mime)
```
omus, line 330:
```suggestion
    if endswith(mime, "json")
```

src/AWS.jl
```diff
@@ -323,19 +353,18 @@ function _generate_rest_resource(request_uri::String, args::Dict{String, Any})
     return request_uri
 end
 
-function _generate_service_url(region::String, request::AbstractDict)
+function _generate_service_url(region::String, service::String, resource::String)
     SERVICE_HOST = "amazonaws.com"
-    endpoint = get(request, :endpoint, request[:service])
     regionless_endpoints = ("iam", "route53")
```
omus, line 358: Rename to `regionless_services`?

src/AWS.jl
```diff
@@ -365,40 +394,41 @@ Perform a RestXML request to AWS
...
+        http_options=get(args, "http_options", []),
+        response_stream=get(args, "response_stream", nothing),
+        return_raw=get(args, "return_raw", false),
+        ordered_json_dict=get(args, "ordered_json_dict", true),
+    )
 
     delete!(args, "headers")
-    delete!(args, "Body")
-
+    delete!(args, "body")
```
omus, line 412: Why did the key change?

src/AWS.jl
```diff
@@ -365,40 +394,41 @@ Perform a RestXML request to AWS
...
     end
 
-    request[:url] = _generate_service_url(aws.region, request)
+    request.url = _generate_service_url(aws.region, request.service, request.resource)
 
     return do_request(aws, request; return_headers=return_headers)
 end
-(service::RestXMLService)(request_method::String, request_uri::String, args=[]) = service(AWSConfig(), request_method, request_uri, args)
-(service::RestXMLService)(a...; b...) = service(a..., b)
+(service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}) = service(AWSConfig(), request_method, request_uri, args)
```
omus, line 430: All-in-one:
```suggestion
(service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String,Any}()) = service(AWSConfig(), request_method, request_uri, args)
```

src/AWS.jl
```diff
@@ -426,17 +456,40 @@ end
...
+    request = Request(
         service=service.name,
-        version=service.api_version,
-        verb=request_method,
-        resource=request_uri,
-        args=args
+        api_version=service.api_version,
+        request_method=request_method,
+        headers=LittleDict{String, String}(get(args, "headers", [])),
+        resource = _generate_rest_resource(request_uri, args),
```
omus, line 472: Keep things consistent
```suggestion
        resource=_generate_rest_resource(request_uri, args),
```

src/AWS.jl
```diff
@@ -426,17 +456,40 @@ end
...
+
+    request.headers["Content-Type"] = "application/json"
+    delete!(args, "headers")
+    request.content = json(args)
+
+    do_request(aws, request; return_headers=return_headers)
 end
-(service::RestJSONService)(request_method::String, request_uri::String, args=[]) = service(aws_config(), request_method, request_uri, args)
-(service::RestJSONService)(a...; b...) = service(a..., b)
+(service::RestJSONService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}) = service(AWSConfig(), request_method, request_uri, args)
```
omus, line 492:
```suggestion
(service::RestJSONService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String,Any}()) = service(AWSConfig(), request_method, request_uri, args)
```

test/AWS.jl
```diff
@@ -335,7 +365,7 @@ end
 
     @testset "PUT - w/ Params" begin
         body = "sample-file-body"
-        AWSServices.s3("PUT", "/$bucket_name/$file_name"; Body=body)
+        AWSServices.s3("PUT", "/$bucket_name/$file_name", Dict("body"=>body))
```
omus, line 368: I'm not sure what the convention in this package is but I usually put spaces around `=>`
```suggestion
        AWSServices.s3("PUT", "/$bucket_name/$file_name", Dict("body" => body))
```

test/AWS.jl
```diff
@@ -347,20 +377,21 @@ end
 
     @testset "GET - w/ Params" begin
         max_keys = 1
-        result = AWSServices.s3("GET", "/$bucket_name"; max_keys=max_keys)
+        result = AWSServices.s3("GET", "/$bucket_name", Dict("max_keys"=>max_keys))
```
omus, line 380:
```suggestion
        result = AWSServices.s3("GET", "/$bucket_name", Dict("max_keys" => max_keys))
```

test/AWS.jl
```diff
@@ -347,20 +377,21 @@ end
...
             <Delete xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
                 <Object>
                     <Key>$file_name</Key>
                 </Object>
             </Delete>
-        """
-        )
+            """
+
+        AWSServices.s3("POST", "/$bucket_name?delete", Dict("body"=>body))
```
omus, line 394:
```suggestion
        AWSServices.s3("POST", "/$bucket_name?delete", Dict("body" => body))
```

test/AWS.jl
```diff
@@ -381,3 +412,51 @@ end
...
+    vault_names = ["aws.jl.test-01---$timestamp", "aws.jl.test-02---$timestamp", ]
+
+    @testset "PUT" begin
+        for vault in vault_names
+            AWSServices.glacier("PUT", "/-/vaults/$vault")
+        end
+    end
+
+    @testset "POST" begin
+        tags = Dict("Tags"=> LittleDict("Tag-01"=>"Tag-01", "Tag-02"=>"Tag-02"))
```
omus, line 427:
```suggestion
        tags = Dict("Tags" => LittleDict("Tag-01" => "Tag-01", "Tag-02" => "Tag-02"))
```

src/AWS.jl
mattBrzezinski: I wanted to make this consistent when passing in other values, `headers` begins with a lowercase character, `body` should as well.

src/AWS.jl
mattBrzezinski: Could you expand on this? I'm not quite following along.

src/AWS.jl
mattBrzezinski: Going to play around and see if I can come up with something generic here

src/AWS.jl
```diff
@@ -57,6 +56,28 @@ end
...
+    request_method::String  # Enum?
+
+    headers::AbstractDict{String, String}=LittleDict{String, String}()
+    content::String=""
+    resource::String=""
+    url::String=""
+
+    return_stream::Bool=false
+    response_stream::Union{Base.BufferStream, Nothing}=nothing
+    http_options::Array=[]
```
iamed2, line 78: Can this be typed?

src/AWS.jl
iamed2: I think he means something like:

```julia
canonical_headers(request::Request) = ["$(lowercase(k)):$(strip(v))" for (k,v) in request.headers]
```

src/AWS.jl
```diff
@@ -176,30 +198,28 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
...
+function _http_request(aws::AWSConfig, request::Request)
     options = []
 
-    if get(request, :return_stream, false)
+    if request.return_stream
         io = Base.BufferStream()
-        request[:response_stream] = io
-        push!(options, (:response_stream, io))
+        request.response_stream = io
+        push!(options, ("response_stream", io))
```
iamed2, line 207: I think we must have missed testing this if this doesn't cause errors. `options` gets passed as kwargs so `response_stream` should be a Symbol, not a string.

src/AWS.jl
```diff
@@ -176,30 +198,28 @@ function _sign_aws4!(aws::AWSConfig, request::LittleDict, time::DateTime)
...
     end
 
-    if haskey(request, :http_options)
-        for option in request[:http_options]
-            push!(options, option)
-        end
+    for option in request.http_options
+        push!(options, option)
```
iamed2, line 211: Could we avoid constructing `options` altogether? We could set `response_stream` to `nothing` (the default) if we're not using it, pass it as a kwarg to `HTTP.request` unconditionally, and then just splat `request.http_options` directly.

mattBrzezinski:
In this [commit](https://github.com/JuliaCloud/AWS.jl/pull/126/commits/37edc32e8b98b79c7aee6d456394c28443cc3d92) I replaced the `OrderedJSONDict=true` field in `struct Request` with a DataType to override instead.

After discussing with @omus the `OrderedDict` maybe useful in the case of AWS IAM as per https://github.com/JuliaCloud/AWSCore.jl/pull/25 , however IAM uses Query requests and I have not implemented those yet.

I think ignoring (but keeping this in mind) for now is an okay approach and when the `QueryService` is implemented, write tests for IAM which verify that this case requires ordering. We can then have a conversation on defaulting to `OrderedDict` for all request types, for certain ones, or for certain services.

src/AWS.jl
omus: ☝️ That is what I meant. If these aren't used multiple times it's probably not worth it but having these as separate functions will allow unit tests to be written for this logic.

mattBrzezinski: Discussed on Slack, since these are only being used in the sigv4 code I'm just going to leave it as is.

I did however make both variables `join(sort())` instead of doing `canonical` in the creation of `canonical_form`.

src/AWS.jl
omus: Was this marked as resolved accidentally or do you want to leave things as is?

src/AWS.jl
```diff
@@ -57,6 +56,47 @@ end
...
+    headers::AbstractDict{String, String}=LittleDict{String, String}()
+    content::String=""
+    resource::String=""
+    url::String=""
+
+    return_stream::Bool=false
+    response_stream::Union{Base.BufferStream, Nothing}=nothing
+    http_options::Array{String}=[]
+    return_raw::Bool=false
+    response_dict_type::Type=LittleDict
```
omus, line 80: Doing this removes the need for the inner constructor I believe:
```suggestion
    response_dict_type::Type{<:AbstractDict}=LittleDict
```

src/AWS.jl
```diff
@@ -345,6 +380,15 @@ function _return_headers(args::Dict{String, Any})
     return return_headers
 end
 
+function _set_response_dict_type!(request::Request, args::AbstractDict{String, <:Any})
+    response_dict_type = get(args, "response_dict_type", "")
+    response_dict_type = eval(Meta.parse(response_dict_type))
```
omus, line 385: I'm hoping there is a better way to do this... Can the args passed into `(::RestXMLService)(...)` not include a type?

iamed2: IMO this must be changed to be a type, and `eval` is not to be used.

mattBrzezinski: Yes they can, I was getting caught up in thinking that this will be passed in as a string and we would need to convert it.

Too much context switching for me these last few days I think. I'll update all these in a bit.

src/AWS.jl
mattBrzezinski: This was accidentally resolved!

src/AWS.jl
```diff
@@ -422,7 +394,9 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
         return_raw=get(args, "return_raw", false),
     )
 
-    _set_response_dict_type!(request, args)
+    if haskey(args, "response_dict_type")
+        request.response_dict_type = get(args, "response_dict_type")
```
iamed2, line 398:
```suggestion
        request.response_dict_type = args["response_dict_type"]
```

src/AWS.jl
```diff
@@ -489,10 +463,12 @@ function (service::RestJSONService)(
         http_options=get(args, "http_options", []),
         response_stream=get(args, "response_stream", nothing),
         return_raw=get(args, "return_raw", false),
-        response_dict_type=OrderedDict,  #
     )
 
-    _set_response_dict_type!(request, args)
+    if haskey(args, "response_dict_type")
+        request.response_dict_type = get(args, "response_dict_type")
```
iamed2, line 469:
```suggestion
        request.response_dict_type = args["response_dict_type"]
```

src/AWS.jl
```diff
@@ -302,16 +308,26 @@ function do_request(aws::AWSConfig, request::AbstractDict; return_headers::Bool=
     body = String(copy(response.body))
 
     if occursin(r"/xml", mime)
-        xml_dict_type = OrderedDict{Union{Symbol, String}, Any}
+        xml_dict_type = response_dict_type{Union{Symbol, String}, Any}
```
iamed2, line 311: You actually can't depend on these parameters being the same. That said, I'm pretty sure all the major dict types support this so I think it's fine to keep it.

src/AWS.jl
```diff
@@ -57,6 +56,27 @@ end
 struct RestJSONService
     name::String
     api_version::String
+
+    service_specific_headers::LittleDict{String, String}
+
+    RestJSONService(name::String, api_version::String, service_specific_headers::LittleDict{String, String}=LittleDict{String, String}()) = new(name, api_version, service_specific_headers)
+end
```
iamed2, line 63:
```suggestion
    service_specific_headers::LittleDict{String, String}
end

RestJSONService(name::String, api_version::String) = RestJSONService(name, api_version, LittleDict{String, String}())
```

PR #127, opened 2020-07-11
Implemented Query Requests

mattBrzezinski:
## Overview
This merge request contains the implementation of Query Service requests. I have also modified the `if x in y` statements to use `Tuple` instead of `Array` for a very, very minor performance benefit as pointed out by @iamed2 in a previous PR.

There was a discussion in a previous PR about the default usage of `OrderedDict` for IAM, [here](https://github.com/JuliaCloud/AWS.jl/pull/126#issuecomment-655096785). Which leads to the original PR which introduce this change, https://github.com/JuliaCloud/AWSCore.jl/pull/25)

Looking into this further, I am not understanding the necessity for the change. When the response is returned back from AWS your IAM policy is a string. When you modify an IAM policy document you are creating a new version of it, not replacing the existing one. Even so, if you are passing in this policy document as a `Dict` your request would fail. It must come through as a `String`. Maybe I am missing something here @omus but I believe that this is still safe. Using the example from  https://github.com/JuliaCloud/AWSCore.jl/pull/25, you would pull down the policy version (as a String), you could parse it into some sort of `Dict`, modify it, then would need to stringify it and make the request.

## Design Decision
In `AWSCore.jl` for `IAM, STS, SQS, SNS` services we had set the request content `ContentType` to `JSON`. I have decided to not implement this feature, as it does not accurately reflect what the AWS docs specify as a response, [see Sample Response](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_DeleteMessageBatch.html).

This results in a small, but important change in responses. Here is an example for a `SQS DeleteMessageBatch` request.

### ContentType=JSON
```julia
{
  "DeleteMessageBatchResponse": {
    "DeleteMessageBatchResult": {
      "Failed": null,
      "Successful": [
        {
          "Id": "msg1"
        },
        {
          "Id": "msg2"
        }
      ]
    },
    "ResponseMetadata": {
      "RequestId": "redacted"
    }
  }
}
```

### ContentType not set
```julia
{
  "version": "1.0",
  "DeleteMessageBatchResponse": {
    "DeleteMessageBatchResult": {
      "DeleteMessageBatchResultEntry": [
        {
          "Id": "msg1"
        },
        {
          "Id": "msg2"
        }
      ]
    },
    "ResponseMetadata": {
      "RequestId": "redacted"
    }
  }
}
```

There are other cases of odd behaviour with using `ContentType=JSON` such as `SQS ReceiveMessage`. This request returns only one message, yet with JSON it comes back as an array. It also does not include important information such as `ReceiptHandle`, and again does not reflect what the documentation states as a [response](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html).

### JSON
```julia
"messages": [
  {
    "Attributes": null,
    "Body": "Hello for AWS.jl",
    "MD5OfBody": "9f501028c282e05583de12e3bfd4ca8b",
    "MD5OfMessageAttributes": null,
    "MessageAttributes": null,
    "MessageId": "6495bfc8-e726-4370-8e89-743436ddcb32",
  }
]
```

### ContentType not set
```julia
"Message": {
  "MessageId": "64ae707c-a354-45d4-9271-f58260fe4928",
  "ReceiptHandle": "redacted",
  "MD5OfBody": "9f501028c282e05583de12e3bfd4ca8b",
  "Body": "Hello for AWS.jl"
}
```

## Future Work
There are only a handful of tasks left for the completion of the `v1` re-write:
- RestJSON functionality
- Removal of `AWSCore.jl` wrappings in the package
- GitHub Actions for auto-generation of code
- Various small cleanups
- Merge `v1->master`, and have a beer 🎉

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -444,7 +444,7 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
 end
 (service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, Any}()) = service(AWSConfig(), request_method, request_uri, args)
 
-function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::AbstractDict{String, <:Any})
+function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}())
     POST_RESOURCE = "/"
     return_headers = _return_headers(args)
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -463,15 +463,9 @@ function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::Ab
 
     request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
 
-    if service.name == "iam"
-        aws.region = "us-east-1"
-    end
-
-    if isempty(request.content)
-        args["Action"] = operation
-        args["Version"] = service.api_version
-        request.content = HTTP.escapeuri(_flatten_query(service.name, args))
-    end
+    args["Action"] = operation
+    args["Version"] = service.api_version
+    request.content = HTTP.escapeuri(_flatten_query(service.name, args))
 
     do_request(aws, request; return_headers=return_headers)
 end
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -86,7 +86,7 @@ http_status(e::HTTP.StatusError) = e.status
 header(e::HTTP.StatusError, k, d="") = HTTP.header(e.response, k, d)
 
 function _sign!(aws::AWSConfig, request::Request; time::DateTime=now(Dates.UTC))
-    if request.service in ["sdb", "importexport"]
+    if request.service in ("sdb", "importexport")
         _sign_aws2!(aws, request, time)
     else
         _sign_aws4!(aws, request, time)
@@ -299,6 +299,7 @@ function do_request(aws::AWSConfig, request::Request; return_headers::Bool=false
 
     # Parse response data according to mimetype...
     mime = HTTP.header(response, "Content-Type", "")
+
     if isempty(mime)
         if length(response.body) > 5 && String(response.body[1:5]) == "<?xml"
             mime = "text/xml"
@@ -360,6 +361,30 @@ function _return_headers(args::AbstractDict{String, <:Any})
     return return_headers
 end
 
+function _flatten_query(service::String, query::AbstractDict{String, <: Any}, prefix::String="")
+    result = Dict{String,String}()
+
+    for (k, v) in query
+        if typeof(v) <: AbstractDict
+            merge!(result, _flatten_query(service, v, "$prefix$k."))
+        elseif typeof(v) <: AbstractArray
+            for (i, j) in enumerate(v)
+                suffix = service in ("ec2", "sqs") ? "" : ".member"
+                prefix_key = "$prefix$k$suffix.$i"
+
+                if typeof(j) <: AbstractDict
+                    merge!(result, _flatten_query(service, j, "$prefix_key."))
+                else
+                    result[prefix_key] = string(j)
+                end
+            end
+        else
+            result["$prefix$k"] = string(v)
+        end
+    end
+
+    return result
+end
 
 """
     (service::RestXMLService)(
@@ -419,17 +444,30 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
 end
 (service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, Any}()) = service(AWSConfig(), request_method, request_uri, args)
 
-function (service::QueryService)(aws::AWS.AWSConfig, operation, args=[])
-    return AWSCore.service_query(
-        aws;
+function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::AbstractDict{String, <:Any})
+    POST_RESOURCE = "/"
+    return_headers = _return_headers(args)
+
+    request = Request(
         service=service.name,
-        version=service.api_version,
-        operation=operation,
-        args=args
+        api_version=service.api_version,
+        resource=POST_RESOURCE,
+        request_method="POST",
+        headers=LittleDict("Content-Type"=>"application/x-www-form-urlencoded; charset=utf-8"),
+        url = _generate_service_url(aws.region, service.name, POST_RESOURCE)
     )
+
+    args["Action"] = operation
+    args["Version"] = service.api_version
+
+    if service.name == "iam"
+        aws.region = "us-east-1"
+    end
+
+    request.content = HTTP.escapeuri(_flatten_query(service.name, args))
+    do_request(aws, request; return_headers=return_headers)
 end
-(service::QueryService)(operation, args=[]) = service(aws_config(), operation, args)
-(service::QueryService)(a...; b...) = service(a..., b)
+(service::QueryService)(operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}()) = service(AWSConfig(), operation, args)
 
 function (service::JSONService)(aws::AWS.AWSConfig, operation, args=[])
     return AWSCore.service_json(
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -453,18 +453,26 @@ function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::Ab
         api_version=service.api_version,
         resource=POST_RESOURCE,
         request_method="POST",
-        headers=LittleDict("Content-Type"=>"application/x-www-form-urlencoded; charset=utf-8"),
-        url = _generate_service_url(aws.region, service.name, POST_RESOURCE)
+        headers=LittleDict{String, String}(get(args, "headers", [])),
+        url=_generate_service_url(aws.region, service.name, POST_RESOURCE),
+        return_stream=get(args, "return_stream", false),
+        http_options=get(args, "http_options", []),
+        response_stream=get(args, "response_stream", nothing),
+        return_raw=get(args, "return_raw", false),
     )
 
-    args["Action"] = operation
-    args["Version"] = service.api_version
+    request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
 
     if service.name == "iam"
         aws.region = "us-east-1"
     end
 
-    request.content = HTTP.escapeuri(_flatten_query(service.name, args))
+    if isempty(request.content)
+        args["Action"] = operation
+        args["Version"] = service.api_version
+        request.content = HTTP.escapeuri(_flatten_query(service.name, args))
+    end
+
     do_request(aws, request; return_headers=return_headers)
 end
 (service::QueryService)(operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}()) = service(AWSConfig(), operation, args)
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -361,25 +361,23 @@ function _return_headers(args::AbstractDict{String, <:Any})
     return return_headers
 end
 
-function _flatten_query(service::String, query::AbstractDict{String, <: Any}, prefix::String="")
-    result = Dict{String,String}()
-
+function _flatten_query(service::String, query::AbstractDict{String, <:Any}, result::Vector{Pair{String, String}}=Vector{Pair{String, String}}(), prefix::String="")
     for (k, v) in query
         if typeof(v) <: AbstractDict
-            merge!(result, _flatten_query(service, v, "$prefix$k."))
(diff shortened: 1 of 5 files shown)


test/AWS.jl
```diff
@@ -341,6 +341,160 @@ end
...
+            "Version"=> "2012-10-17",
+            "Statement"=> [
+                LittleDict(
+                    "Effect"=>"Allow",
+                    "Action"=>["s3:Get*", "s3:List*"],
+                    "Resource"=> ["arn:aws:s3:::my-bucket/shared/*"]
+                )
+            ],
+        )
+        expected_policy_document = JSON.json(expected_policy_document)
```
mattBrzezinski, line 388: I defined `expected_policy_document` explicitly as a `LittleDict` just to show case that it's easy enough to convert it to a string and pass it along to the function.

iamed2:
Setting ContentType=JSON removes some of the inconsistencies with XML parsing we run into. 

If you had a request that returns multiple messages, and it _happened_ to return only one, you'd get a single item instead of an array. This can be pretty surprising to users and result in some pretty complicated parsing code. This is XMLDict's doing. 

BTW, this is what you get from Boto3 for DeleteMessageBatch:

```
{
    'Successful': [
        {
            'Id': 'string'
        },
    ],
    'Failed': [
        {
            'Id': 'string',
            'SenderFault': True|False,
            'Code': 'string',
            'Message': 'string'
        },
    ]
}
```

The AWS SDK for JS also has "Successful" and "Failed" sections; not having that seems like a problem! Unless they have a different name in the XML?

iamed2:
> It also does not include important information such as ReceiptHandle

I get this returned when I make a ReceiveMessage request using AWSCore. I'm not sure what happened to make yours not have this property.

mattBrzezinski:
> It would be nice if we could discuss this with a non-mutating API call so we can experiment more.

## SQS ListQueues

### No queues exist
#### ContentType=JSON
```julia
{
  "ListQueuesResponse": {
    "ListQueuesResult": {
      "NextExclusiveStartQueueName": null,
      "NextToken": null,
      "queueUrls": null
    },
    "ResponseMetadata": {
      "RequestId": "b7b3b953-36f4-5140-8448-eed5ab64b844"
    }
  }
}
```

#### ContentType not set
```julia
{
  "version": "1.0",
  "ListQueuesResponse": {
    "ListQueuesResult": {},
    "ResponseMetadata": {
      "RequestId": "e263e3ef-0b8e-5ca8-aab3-b46996c7278b"
    }
  }
}
```

### One queue exists
#### ContentType=JSON
```julia
{
  "ListQueuesResponse": {
    "ListQueuesResult": {
      "NextExclusiveStartQueueName": null,
      "NextToken": null,
      "queueUrls": [
        "https://sqs.us-east-1.amazonaws.com/redacted/a"
      ]
    },
    "ResponseMetadata": {
      "RequestId": "2718544b-ea89-5ab3-8ef6-03cd741efeee"
    }
  }
}
```
#### ContentType not set
```julia
{
  "version": "1.0",
  "ListQueuesResponse": {
    "ListQueuesResult": {
      "QueueUrl": "https://sqs.us-east-1.amazonaws.com/redacted/a"
    },
    "ResponseMetadata": {
      "RequestId": "fdf9033e-8313-5343-968d-54b81d7b315a"
    }
  }
}
```

### Multiple queues exist
#### ContentType=JSON
```julia
{
  "ListQueuesResponse": {
    "ListQueuesResult": {
      "NextExclusiveStartQueueName": null,
      "NextToken": null,
      "queueUrls": [
        "https://sqs.us-east-1.amazonaws.com/redacted/a",
        "https://sqs.us-east-1.amazonaws.com/redacted/b"
      ]
    },
    "ResponseMetadata": {
      "RequestId": "f964379a-4ecd-5af4-84f4-601cd01af38b"
    }
  }
}
```

#### ContentType not set
```julia
{
  "version": "1.0",
  "ListQueuesResponse": {
    "ListQueuesResult": {
      "QueueUrl": [
        "https://sqs.us-east-1.amazonaws.com/redacted/a",
        "https://sqs.us-east-1.amazonaws.com/redacted/b"
      ]
    },
    "ResponseMetadata": {
      "RequestId": "ac19d5fd-1c17-58c6-bc8e-827b82e1fe2f"
    }
  }
}
```

mattBrzezinski:
> > It also does not include important information such as ReceiptHandle
> 
> I get this returned when I make a ReceiveMessage request using AWSCore. I'm not sure what happened to make yours not have this property.

I must have been doing something strange when playing around with this merge request, as that does not hold true. I just made the `ReceiveMessage` request again with `ContentType=JSON` and got it;

```julia
{
  "ReceiveMessageResponse": {
    "ReceiveMessageResult": {
      "messages": [
        {
          "Attributes": null,
          "Body": "hi",
          "MD5OfBody": "8843cfb8cbd3367b0dce5a5426a022d9",
          "MD5OfMessageAttributes": null,
          "MessageAttributes": null,
          "MessageId": "3aaef1c7-836c-44b5-9676-e35cc52c0cc4",
          "ReceiptHandle": "redacted"
        }
      ]
    },
    "ResponseMetadata": {
      "RequestId": "ed302bc8-843e-5419-a4c4-7c4e1bb34efb"
    }
  }
}
```

src/AWS.jl
```diff
@@ -360,6 +361,30 @@ function _return_headers(args::AbstractDict{String, <:Any})
     return return_headers
 end
 
+function _flatten_query(service::String, query::AbstractDict{String, <: Any}, prefix::String="")
+    result = Dict{String,String}()
```
iamed2, line 365: This will remove the ordering of the outermost layer, if any. Is that okay? Perhaps, considering we never do lookups on this dictionary, we could just create a `Vector{Pair{String, String}}` (which would be accepted by `HTTP.escapeuri`).

Also, how about a `_flatten_query` wrapper that calls a `_flatten_query!` that puts everything in the same container, rather than creating new ones?

src/AWS.jl
```diff
@@ -419,17 +444,38 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
...
+        return_stream=get(args, "return_stream", false),
+        http_options=get(args, "http_options", []),
+        response_stream=get(args, "response_stream", nothing),
+        return_raw=get(args, "return_raw", false),
     )
+
+    request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
+
+    if service.name == "iam"
+        aws.region = "us-east-1"
```
iamed2, line 467: This seems very bad? Mutating the input to set a region, and also setting the region for a regionless service.

src/AWS.jl
```diff
@@ -419,17 +444,38 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
...
+    end
+
+    if isempty(request.content)
```
iamed2, line 470: Is there a reason why `content` would contain anything?

test/AWS.jl
```diff
@@ -341,6 +341,147 @@ end
...
+
+        @testset "Create Policy" begin
+            response = AWSServices.iam("CreatePolicy", LittleDict("PolicyName"=>expected_policy_name, "PolicyDocument"=>expected_policy_document))
+            policy_arn = response["CreatePolicyResponse"]["CreatePolicyResult"]["Policy"]["Arn"]
+        end
+
+        @testset "Get Policy" begin
```
iamed2, line 395: Tests that depend on each other should be in the same testset

src/AWS.jl
mattBrzezinski: I was under the impression that IAM calls even though are regionless, must be made to `us-east-1`. I just tested with `us-east-2`, and this does not seem to be the case anymore.

I'll update this.

src/AWS.jl
mattBrzezinski: The user could create the content themselves if they choose to.

test/AWS.jl
mattBrzezinski: Technically these are all under the `iam` testset, I just broke them down one level further to their respective request calls. I can remove this extra layer if you'd like since they are dependent on each other.

src/AWS.jl
mattBrzezinski: Ignore my previous comment, the user cannot set `content` themselves currently. And that probably would not be used.

I'll update this.

test/AWS.jl
iamed2: Yeah the main difference is in this case:

```
julia> @testset "foo" begin
           @testset "bar" begin
               error()
               foo = 1
           end
           @testset "baz" begin
               @test 0 == foo
           end
       end
```
If for some reason something outside of a test in "bar" errors, then you'll end up with an `UndefVarError` in "baz", despite only one actual source of errors. But if you structure your tests like this:
```
julia> @testset "foo" begin
           @testset "bar" begin
               error()
               foo = 1
               @test 0 == foo
           end
       end
```
you'll only get the one error.

mattBrzezinski: Makes sense, especially in this case where things are dependent on each other. I'll make the necessary changes most likely on Wednesday since tomorrow will be a JuliaCon 2020 prep / record my video day.

src/AWS.jl
mattBrzezinski: I have not tested this with every service, obviously. However I do not think that ordering should matter at all, since AWS requests a `param.n` notation specifically and should be handling that on their end?

These are the benchmark results using `args` from the `_flatten_query` test.

```julia
high_level_value = "high_level_value"
entry_1 = LittleDict("low_level_key_1"=>"low_level_value_1", "low_level_key_2"=>"low_level_value_2", "low_level_key_3"=>["ll-key-3", "ll-key-4"])
entry_2 = LittleDict("low_level_key_3"=>"low_level_value_3", "low_level_key_4"=>"low_level_value_4")

args = LittleDict("high_level_key"=>high_level_value, "high_level_array"=>[entry_1, entry_2])


@benchmark AWS._flatten_query("sqs", args)
BenchmarkTools.Trial: 
  memory estimate:  3.08 KiB
  allocs estimate:  42
  --------------
  minimum time:     2.235 μs (0.00% GC)
  median time:      2.342 μs (0.00% GC)
  mean time:        2.741 μs (6.34% GC)
  maximum time:     312.621 μs (98.44% GC)
  --------------
  samples:          10000
  evals/sample:     9
```

These numbers seem fine to me?

iamed2: Using a single data structure rather than merging them would avoid a few GCs though, which is always good. Recursively creating and destroying dictionaries seems like something we can easily avoid there.

mattBrzezinski: Good point! I'm going to play around and come up with a better solution here.

mattBrzezinski: I made some changes in https://github.com/JuliaCloud/AWS.jl/pull/127/commits/2301656f4131b88dcbd79ab61c536b9a6b01343f, I'm not sure if this is what you're looking for?

Instead of creating multiple dicts it's just an `Array{Pair{String, String}}` that get passed around recursively, objects get added in. Also updated how `strings` get created to be a bit more performant. Here's the benchmark results with the same test as previously done.

```julia
high_level_value = "high_level_value"
entry_1 = LittleDict("low_level_key_1"=>"low_level_value_1", "low_level_key_2"=>"low_level_value_2", "low_level_key_3"=>["ll-key-3", "ll-key-4"])
entry_2 = LittleDict("low_level_key_3"=>"low_level_value_3", "low_level_key_4"=>"low_level_value_4")

args = LittleDict("high_level_key"=>high_level_value, "high_level_array"=>[entry_1, entry_2])

@benchmark _flatten_query("s3", args)
BenchmarkTools.Trial: 
  memory estimate:  1.69 KiB
  allocs estimate:  38
  --------------
  minimum time:     1.769 μs (0.00% GC)
  median time:      1.823 μs (0.00% GC)
  mean time:        2.086 μs (3.69% GC)
  maximum time:     372.865 μs (98.91% GC)
  --------------
  samples:          10000
  evals/sample:     10
```

Almost half the estimated memory allocated, and shaved off `0.7 μs` off the mean time.

mattBrzezinski:
> What happens to the policy if we don't delete it?

The account which runs the test will have it exist until it is manually removed, this will not incur any charges AFAIK.

I do eventually wish to create a script which removes all custom resources to clean up an account at some point.

src/AWS.jl
```diff
@@ -360,6 +361,28 @@ function _return_headers(args::AbstractDict{String, <:Any})
     return return_headers
 end
 
+function _flatten_query(service::String, query::AbstractDict{String, <:Any}, result::Vector{Pair{String, String}}=Vector{Pair{String, String}}(), prefix::String="")
```
iamed2, line 364: This is good, but the Julia style for mutating/non-mutating function versions would look like this:

```suggestion
function _flatten_query(service::String, query::AbstractDict{String, <:Any}, prefix::String="")
    return _flatten_query!(Pair{String, String}[], service, query, prefix)
end

function _flatten_query!(result::Vector{Pair{String, String}}, service::String, query::AbstractDict{String, <:Any}, prefix::String)
```

PR #128, opened 2020-07-12
Implement JSON Service

mattBrzezinski:
## Notes
~This is based off of https://github.com/JuliaCloud/AWS.jl/pull/127, do not merge until that branch goes in!~

~This merge request is currently targetting the `MB/query-service` branch so you only see the differences between the two, when #127 goes in, this needs to target `v1`.~

## Overview
This PR implements the JSON request functionality.

## Next Steps
- GitLab Actions for auto-generation of code
- Various clean up; removal of `AWSCore`, properly use `AWSConfig`, etc.
- Add `bors`, and CI to this repo
- Create the final merge request `v1->master`, :tada:

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -444,7 +444,7 @@ function (service::RestXMLService)(aws::AWSConfig, request_method::String, reque
 end
 (service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, Any}()) = service(AWSConfig(), request_method, request_uri, args)
 
-function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::AbstractDict{String, <:Any})
+function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}())
     POST_RESOURCE = "/"
     return_headers = _return_headers(args)
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -8,12 +8,11 @@ using Mocking
 using OrderedCollections: LittleDict, OrderedDict
 using Retry
 using Sockets
-using SymDict
 using XMLDict
 
-export @service, AWSServices, RestJSONService, JSONService, RestXMLService, QueryService, AWSConfig, AWSExceptions, set_user_agent
-
-const AWSRequest = SymbolDict
+export @service
+export AWSConfig, AWSExceptions, AWSServices, Request, set_user_agent
+export JSONService, RestJSONService, RestXMLService, QueryService
 
 include("AWSCredentials.jl")
 include("AWSConfig.jl")
@@ -57,6 +56,27 @@ end
 struct RestJSONService
     name::String
     api_version::String
+
+    service_specific_headers::LittleDict{String, String}
+end
+
+RestJSONService(name::String, api_version::String) = RestJSONService(name, api_version, LittleDict{String, String}())
+
+Base.@kwdef mutable struct Request
+    service::String
+    api_version::String
+    request_method::String
+
+    headers::AbstractDict{String, String}=LittleDict{String, String}()
+    content::String=""
+    resource::String=""
+    url::String=""
+
+    return_stream::Bool=false
+    response_stream::Union{Base.BufferStream, Nothing}=nothing
+    http_options::Array{String}=[]
+    return_raw::Bool=false
+    response_dict_type::Type{<:AbstractDict}=LittleDict
 end
 
 # Needs to be included after the definition of struct otherwise it cannot find them
@@ -65,25 +85,25 @@ include("AWSServices.jl")
 http_status(e::HTTP.StatusError) = e.status
 header(e::HTTP.StatusError, k, d="") = HTTP.header(e.response, k, d)
 
-function _sign!(aws::AWSConfig, request::LittleDict; time::DateTime=now(Dates.UTC))
-    if request[:service] in ["sdb", "importexport"]
+function _sign!(aws::AWSConfig, request::Request; time::DateTime=now(Dates.UTC))
+    if request.service in ["sdb", "importexport"]
         _sign_aws2!(aws, request, time)
     else
         _sign_aws4!(aws, request, time)
     end
 end
 
-function _sign_aws2!(aws::AWSConfig, request::LittleDict, time::DateTime)
+function _sign_aws2!(aws::AWSConfig, request::Request, time::DateTime)
     # Create AWS Signature Version 2 Authentication query parameters.
     # http://docs.aws.amazon.com/general/latest/gr/signature-version-2.html
 
     query = Dict{String, String}()
-    for elem in split(request[:content], '&', keepempty=false)
+    for elem in split(request.content, '&', keepempty=false)
         (n, v) = split(elem, "=")
         query[n] = HTTP.unescapeuri(v)
     end
 
-    request[:headers]["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
+    request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
 
     creds = check_credentials(aws.credentials)
     query["AWSAccessKeyId"] = creds.access_key_id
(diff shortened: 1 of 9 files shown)


src/AWS.jl
```diff
@@ -314,6 +314,14 @@ function do_request(aws::AWSConfig, request::Request; return_headers::Bool=false
         return (return_headers ? (xml_dict(body, xml_dict_type), response_dict_type(response.headers)) : xml_dict(body, xml_dict_type))
     end
 
+    if occursin(r"/x-amz-json-1.[01]$", mime)
+        if isempty(response.body)
```
iamed2, line 318: What if `return_headers`?

src/AWS.jl
```diff
@@ -465,27 +473,45 @@ function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::Ab
 
     request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
 
-    args["Action"] = operation
-    args["Version"] = service.api_version
-    request.content = HTTP.escapeuri(_flatten_query(service.name, args))
+    if service.name == "iam"
+        aws.region = "us-east-1"
```
iamed2, line 477: I thought this was removed?

mattBrzezinski: Oops, based rebase, good catch.

src/AWS.jl
mattBrzezinski: Good call.

This also lead me down a giant rabbit hole of trying to figure out how to handle this function itself better (see Slack). To begin with and as a reminder for myself, we should sort this mime section based off the most popular request type in descending order.

test/AWS.jl
```diff
@@ -379,6 +379,42 @@ end
...
+                    "ClientRequestToken"=>string(uuid4())
+                )
+            )
+        end
+
+        @testset "get secret value" begin
+            @test _get_secret_string(secret_name) == secret_string
+        end
+
+        @testset "delete secret" begin
```
mattBrzezinski, line 406: Reminder for myself to update this in a similar sense of how `Query-Service` was done to ensure we always delete.

iamed2:
What if this:
```
        if isempty(response.body)
            return (return_headers ? (nothing, response.headers) : nothing)
        end
        info = JSON.parse(body, dicttype=response_dict_type)
        return (return_headers ? (info, response_dict_type(response.headers)) : info)
```
became this:
```
        info = isempty(response.body) ? nothing : JSON.parse(body, dicttype=response_dict_type)
        return (return_headers ? (info, response_dict_type(response.headers)) : info)
```

src/AWS.jl
```diff
@@ -473,19 +471,31 @@ function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::Ab
...
-        args=args
+        api_version=service.api_version,
+        resource=POST_RESOURCE,
+        request_method="POST",
+        headers=LittleDict{String, String}(get(args, "headers", [])),
+        url=_generate_service_url(aws.region, service.name, POST_RESOURCE),
+        return_stream=get(args, "return_stream", false),
+        http_options=get(args, "http_options", []),
+        response_stream=get(args, "response_stream", nothing),
+        return_raw=get(args, "return_raw", false),
```
iamed2, line 488:
```suggestion
        return_raw=get(args, "return_raw", false),
        content=json(args),
```

No reason to do it separately, right?

mattBrzezinski: No, previously we did some extra stuff with `args` but that was not needed.

src/AWSMetadataUtilities.jl
```diff
@@ -218,7 +218,7 @@ function _get_function_parameters(input::String, shapes::Dict)
         end
     end
 
-    return (sort(required_parameters), sort(optional_parameters))
+    return (sort!(OrderedDict(required_parameters)), sort!(OrderedDict(optional_parameters)))
```
iamed2, line 221: Why did this change have to be made? Also, how is this function used? Could these be LittleDicts?

src/AWS.jl
```diff
@@ -473,19 +465,30 @@ function (service::QueryService)(aws::AWS.AWSConfig, operation::String, args::Ab
...
-        version=service.api_version,
-        json_version=service.json_version,
-        target=service.target,
-        operation=operation,
-        args=args
+        api_version=service.api_version,
+        resource=POST_RESOURCE,
+        request_method="POST",
+        headers=LittleDict{String, String}(get(args, "headers", [])),
+        content = json(args),
```
iamed2, line 478:
```suggestion
        content=json(args),
```

src/AWSMetadataUtilities.jl
mattBrzezinski: It wasn't a required change, however I wanted to get rid of this warning.

```
┌ Warning: `sort(d::Dict; args...)` is deprecated, use `sort!(OrderedDict(d); args...)` instead.
│   caller = _get_function_parameters(::String, ::Dict{String,Any}) at AWSMetadataUtilities.jl:221
└ @ AWS.AWSMetadataUtilities ~/.julia/dev/AWS.jl/src/AWSMetadataUtilities.jl:221
```

test/AWS.jl
```diff
@@ -379,6 +379,38 @@ end
...
+
+            return response["SecretString"]
+        end
+
+        resp = AWSServices.secrets_manager("CreateSecret", LittleDict(
+                "Name"=>secret_name,
+                "SecretString"=>secret_string,
+                "ClientRequestToken"=>string(uuid4())
+            )
+        )
```
iamed2, line 398:
```suggestion
            "Name"=>secret_name,
            "SecretString"=>secret_string,
            "ClientRequestToken"=>string(uuid4()),
        ))
```

test/AWS.jl
```diff
@@ -379,6 +379,38 @@ end
...
+        function _get_secret_string(secret_name)
+            response = AWSServices.secrets_manager("GetSecretValue", LittleDict("SecretId"=>secret_name))
+
+            return response["SecretString"]
+        end
+
+        resp = AWSServices.secrets_manager("CreateSecret", LittleDict(
+                "Name"=>secret_name,
+                "SecretString"=>secret_string,
+                "ClientRequestToken"=>string(uuid4())
```
iamed2, line 396: Why do we set ClientRequestToken manually? 

If we're not currently doing this automatically, we should add it in before release IMO. If that's too hard, we should provide a `token() = string(uuid4())` function so people don't have to think much about it.

mattBrzezinski: We are not doing this automatically. I didn't realize that this was a thing with most request, I briefly read over the SecretsManager documentation for creating a secret and came across it. The AWS CLI and official AWS SDKs handle this automatically, however outside of that it is required to be set.

---

(Optional) If you include SecretString or SecretBinary , then an initial version is created as part of the secret, and this parameter specifies a unique identifier for the new version.

This value helps ensure idempotency. Secrets Manager uses this value to prevent the accidental creation of duplicate versions if there are failures and retries during a rotation. We recommend that you generate a UUID-type value to ensure uniqueness of your versions within the specified secret.

If the ClientRequestToken value isn't already associated with a version of the secret then a new version of the secret is created. 

If a version with this value already exists and the version SecretString and SecretBinary values are the same as those in the request, then the request is ignored. 

If a version with this value already exists and that version's SecretString and SecretBinary values are different from those in the request then the request fails because you cannot modify an existing version. Instead, use PutSecretValue to create a new version. This value becomes the VersionId of the new version.

src/AWSMetadataUtilities.jl
```diff
@@ -195,8 +195,8 @@ Get the required and optional parameters for a given operation.
 - `Tuple(Dict, Dict)`: (required_parameters, optional_parameters)
 """
 function _get_function_parameters(input::String, shapes::Dict)
-    required_parameters = Dict{String, Any}()
-    optional_parameters = Dict{String, Any}()
+    required_parameters = LittleDict{String, Any}()
+    optional_parameters = LittleDict{String, Any}()
```
iamed2, line 199: https://github.com/JuliaCollections/OrderedCollections.jl/issues/51

this is fine for now though

test/AWS.jl
mattBrzezinski: Issue logged, https://github.com/JuliaCloud/AWS.jl/issues/131

test/AWS.jl
```diff
@@ -379,6 +379,38 @@ end
...
+        )
+
+        try
+            @test _get_secret_string(secret_name) == secret_string
+        finally
+            AWSServices.secrets_manager("DeleteSecret", LittleDict(
+                    "SecretId"=>secret_name,
+                    "ForceDeleteWithoutRecovery"=>"true",
+                )
+            )
```
iamed2, line 407:
```suggestion
            AWSServices.secrets_manager("DeleteSecret", LittleDict(
                "SecretId"=>secret_name,
                "ForceDeleteWithoutRecovery"=>"true",
            ))
```

test/AWS.jl
```diff
@@ -379,6 +379,36 @@ end
...
+            response = AWSServices.secrets_manager("GetSecretValue", LittleDict("SecretId"=>secret_name))
+
+            return response["SecretString"]
+        end
+
+        resp = AWSServices.secrets_manager("CreateSecret", LittleDict(
+                "Name"=>secret_name,
+                "SecretString"=>secret_string,
+                "ClientRequestToken"=>string(uuid4())
+            ))
```
iamed2, line 397:
```suggestion
            "Name"=>secret_name,
            "SecretString"=>secret_string,
            "ClientRequestToken"=>string(uuid4())
        ))
```

Issue #131, opened 2020-07-21
Automatically create a ClientRequestToken


mattBrzezinski:
https://github.com/JuliaCloud/AWS.jl/pull/128#discussion_r458381622

mattBrzezinski:
Example of a function which requires an [IdempotencyToken](https://github.com/aws/aws-sdk-js/blob/master/apis/secretsmanager-2017-10-17.normal.json#L632).

Easiest way is in the high-level API generation, check if this property exists. If so, when generating it, include a kwarg for it which users can override. At the low-level API people can handle this themselves.

mattBrzezinski:
There is a design decision to make with this issue. If we choose to implement this, then the user will not be able to set the IdempotencyToken themselves due to how `mergewith` works.

We are generating our functions like so:

```julia
CreateSecret(Name, args::AbstractDict{String, <:Any}) = secrets_manager("CreateSecret", Dict{String, Any}(mergewith(merge, Dict{String, Any}("Name"=>Name, "ClientRequestToken"=>string(uuid4())), args)))
```

The problem with this lies is if the user passes in `args` containing `ClientRequestToken`, e.g.

```julia
using Compat

args = Dict("ClientRequestToken"=>"foo")

mergewith(merge, Dict("ClientRequestToken"=>"bar"), args)
> ERROR: MethodError: no method matching merge(::String, ::String)
```

If we change this to `mergewith(Dict("ClientRequestToken"=>"bar", args)` this is fine and we will end up with `ClientRequestToken=>"foo"`. However, we cannot do this because we need `merge` when dealing with nested Dictionaries for parameters.

We have three options:
- Look into an alternative way to construct the `args` dictionary which will be passed down to the request
- Auto-generate the idempotency token, however the user will not be able to override this
- Force the user to pass in the idempotency token themselves when necessary

Any thoughts / solutions on this? @omus @iamed2

omus:
I think `merge` has the exact behaviour you want in this case. There may be a way to have a custom merge to get the behaviour we want.

omus:
A quick POC:
```julia
julia> using Compat

julia> a = Dict("k" => Dict("k2"=>"v"), "ClientRequestToken" => "bar")
Dict{String,Any} with 2 entries:
  "ClientRequestToken" => "bar"
  "k"                  => Dict("k2"=>"v")

julia> b = Dict("k" => Dict("k3"=>"v"), "ClientRequestToken" => "foo")
[...]

PR #135, opened 2020-07-28
Add support to auto-generate idempotency tokens for users

mattBrzezinski:
## Resolves 
* #131

## Overview
Commit https://github.com/JuliaCloud/AWS.jl/pull/135/commits/82cf27fcfcc94b8de63be6c2fa8be213e850d1ed contains the actual changes, while the other commit contains the regenerate definitions.

When generating a high-level function these changes look at the parameters to check if they have the property `idempotentToken: true`. If a functions parameters do, we will auto-generate a `string(uuid4())` for the user so that they do not need to pass one in themselves. Alternatively, if they'd like they can.

One design problem with this which I do not like, however I cannot think of another one is how to handle all the various cases. When generating a REST definition we have to account for every combination of:

- Required parameters
- Required parameters in the URI
- Parameters in the header location
- Idempotent parameters
- Arguments passed in by the end user

There is no easy way to build strings up, and thus this results in a bit of an arrow-head design in `_generate_rest_operation_defintion()` and `_generate_json_query_operation_definition()`. However I believe this is the best solution. The alternative is very large nested `if if if`, or attempting to build strings but that because more complicated. Instead I opted to essentially build a truth table by seeing which combinations of the above are present.

I have also remove the `service_name(a...; b...) = service_name(a..., b)` as these aren't really necessary (I believe).

If anyone has an idea for an alternative solution please let me know, I'd be happy to discuss it!


src/AWSMetadataUtilities.jl
```diff
@@ -356,105 +362,218 @@ function _generate_high_level_definition(
...
+            return """\n
+                $function_name() = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(idempotent_kv, ", "))))
+                $function_name(args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, Dict{String, Any}($(join(idempotent_kv, ", "))), args...)))
+                """
+        else
+            return """\n
+                $function_name() = $service_name(\"$method\", \"$request_uri\")
+                $function_name(args::AbstractDict{String, Any}) = $service_name(\"$method\", \"$request_uri\", args)
+                """
         end
```
omus, line 477: Lots of redundancy and chances for mistakes to be introduced. An alternative:
```julia
params_str = (required_keys || idempotent) ? "Dict{String, Any}($(join(vcat(req_kv, idempotent_kv), ", "))" ? ""
headers_str = headers ? "\"headers\"=>Dict{String, Any}($(join(header_kv, ", "))" : ""
params_headers_str = join([s for s in (params_str, headers_str) if !isempty(s)], ", ")

if required_keys && (idempotent || headers)
    return """\n
        $function_name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\", $params_headers_str)
        $function_name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, $params_headers_str, args...)))
        """
elseif !required_keys && (idempotent || headers)
    return """\n
        $function_name() = $service_name(\"$method\", \"$request_uri\", $params_headers_str)
        $function_name(args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, $params_headers_str, args...)))
        """
elseif required_keys
    return """\n
        $function_name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\")
        $function_name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", args)
        """
else
    return """\n
        $function_name() = $service_name(\"$method\", \"$request_uri\")
        $function_name(args::AbstractDict{String, Any}) = $service_name(\"$method\", \"$request_uri\", args)
        """
end
```

src/AWSMetadataUtilities.jl
```diff
@@ -356,105 +362,218 @@ function _generate_high_level_definition(
...
+                """
+        elseif required_keys && required_kv
+            return """\n
+                $name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", "))))
+                $name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", ")), args...))
+                """
+        elseif required_keys
+            return """\n
+                $name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\")
+                $name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", args)
```
omus, line 460:
```suggestion
                $function_name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\")
                $function_name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", args)
```

src/AWSMetadataUtilities.jl
```diff
@@ -356,105 +362,218 @@ function _generate_high_level_definition(
...
+                """
+        elseif headers && idempotent
+            return """\n
+                $function_name() = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(idempotent_kv, ", ")), "headers"=>Dict{String, Any}($(join(header_kv, ", ")))))
+                $function_name(args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, Dict{String, Any}($(join(idempotent_kv, ", ")), "headers"=>Dict{String, Any}($(join(header_kv, ", ")))), args...)))
+                """
+        elseif required_keys && required_kv
+            return """\n
+                $name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", "))))
+                $name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", ")), args...))
```
omus, line 455:
```suggestion
                $function_name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", "))))
                $function_name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}($(join(req_kv, ", ")), args...))
```

src/AWSMetadataUtilities.jl
```diff
@@ -356,105 +362,218 @@ function _generate_high_level_definition(
...
+                """
+        elseif required_keys
+            return """\n
+                $name($(join(req_keys, ", "))) = $service_name(\"$method\", \"$request_uri\")
+                $name($(join(req_keys, ", ")), args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", args)
+                """
+        elseif headers
+            return """\n
+                $name() = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}("headers"=>Dict{String, Any}($(join(header_kv, ", ")))))
+                $name(args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, Dict{String, Any}("headers"=>Dict{String, Any}($(join(header_kv, ", ")))), args...)))
```
omus, line 465:
```suggestion
                $function_name() = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}("headers"=>Dict{String, Any}($(join(header_kv, ", ")))))
                $function_name(args::AbstractDict{String, <:Any}) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, Dict{String, Any}("headers"=>Dict{String, Any}($(join(header_kv, ", ")))), args...)))
```

src/AWSMetadata.jl
```diff
@@ -101,7 +101,9 @@ function _generate_high_level_wrapper(services::Array{OrderedDict})
         open(service_path, "w") do f
             println(f, "# This file is auto-generated by AWSMetadata.jl")
             println(f, "include(\"../AWSServices.jl\")")
+            println(f, "include(\"_utilities.jl\")")
```
omus, line 104: I think this should probably exist outside the `services` folder

src/services/_utilities.jl
```diff
@@ -0,0 +1,4 @@
+# This file contains functions used in the high level service definitions
+
+_merge(a::AbstractDict, b::AbstractDict) = merge(a, b)
+_merge(a::AbstractString, b::AbstractString) = b
```
omus, line 4: Should add some minimal tests.

src/AWSMetadata.jl
mattBrzezinski: I was not sure where this file should live. I placed it in the `services` folder, since most of these will be using it, and prefaced it with an underscore so that it lives at the top of the directory (and to indicate that it's an internal file).

Any ideas where it should go?

omus: I would just put it in `src/`. The problem with putting non services here is you'll need special checks to avoid `@service _Utilities`

src/AWSMetadataUtilities.jl
mattBrzezinski: Ugh... late last night I flubbed this up and rebased these changes into what existed. My apologizes for making reviewing this more difficult.

The example here that you posted worked well, and I implemented it with a few changes.
- The need to account for `req_keys` being present and `req_kv` being present as well
- Dropped the `args...` since our `mergewith(_merge(...))` will handle using the second case being the one that lives when both are present

src/AWSMetadata.jl
mattBrzezinski: Sounds good! Yeah I was just going to hand wave over that being a possibility, since there probably shouldn't ever be an AWS service called `_Utilities`.

I'll move this, and update the references to it!

iamed2:
I know this is late, but why not add this as a field for the service? It can check whether or not an arg was provided with the given name (stored in the field) at that level, rather than adding code to so many function calls.

mattBrzezinski:
> I know this is late, but why not add the idempotency token as a field for the service? It can check whether or not an arg was provided with the given name (stored in the field) at that level, rather than adding code to so many function calls.
> 
> Is it because the token could be at any given level in the parameter hierarchy?

It's never too late!

From what I have seen, idempotent tokens are always in "optional" parameters. Them being optional is not really true because official APIs will auto-generate them for you if they are not provided. If you're doing what we do, then they are required parameters (where applicable, sometimes they're required like for creating a secret in SecretsManager, other times they truly are optional).

Idempotent tokens are based on a function level not a service level, not all functions in a service have idempotent tokens. I do not remember off hand if idempotent tokens can be nestled in deeper levels of a request object (similar to how locationName is), but off-hand I do not think that they are.

Note, all of this is from me just reading the API definitions. There's so many AWS Services so this information may not be 100% accurate, since I don't have time to read through all of them.

iamed2:
> Idempotent tokens are based on a function level not a service level, not all functions in a service have idempotent tokens.

Right, sorry, never mind!

src/AWSMetadata.jl
```diff
@@ -101,7 +101,9 @@ function _generate_high_level_wrapper(services::Array{OrderedDict})
         open(service_path, "w") do f
             println(f, "# This file is auto-generated by AWSMetadata.jl")
             println(f, "include(\"../AWSServices.jl\")")
+            println(f, "include(\"../_utilities.jl\")")
```
iamed2, line 104: There should be no `include`s in here. Just do `using AWS` and pull everything out of there.

src/AWSMetadata.jl
```diff
@@ -100,9 +100,10 @@ function _generate_high_level_wrapper(services::Array{OrderedDict})
         service_path = joinpath(@__DIR__, "services", "$service_name.jl")
         open(service_path, "w") do f
             println(f, "# This file is auto-generated by AWSMetadata.jl")
-            println(f, "include(\"../AWSServices.jl\")")
-            println(f, "using Compat")
+            println(f, "using AWS")
             println(f, "using .AWSServices: $service_name\n")
```
iamed2, line 104:
```suggestion
            println(f, "using AWS.AWSServices: $service_name\n")
```
in case of shadowing

src/AWSMetadataUtilities.jl
```diff
@@ -286,7 +286,13 @@ function _get_function_parameters(input::String, shapes::Dict)
             parameter_name = get(input_shape["members"][parameter[1]], "locationName", parameter[1])
 
             if !haskey(required_parameters, parameter_name)
-                optional_parameters[parameter_name] = _documentation_cleaning(get(parameter[2], "documentation", ""))
+                documentation = _documentation_cleaning(get(parameter[2], "documentation", ""))
```
iamed2, line 289: Should rename to `_clean_documentation` but that's not for this PR

mattBrzezinski: https://github.com/JuliaCloud/AWS.jl/pull/144

PR #138, opened 2020-07-29
Rename high-level wrapper files to be dotfiles

omus:
The motivation behind this is to make code reviews easier by allowing these files to be easily ignored. Using dotfiles seemed to be the best way to do this

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/filtering-files-in-a-pull-request


omus:
Demonstration: https://github.com/JuliaCloud/AWS.jl/pull/139. Doesn't look like GitHub actually can filter on dotfiles

omus:
I tried out an extension like `.gen.jl` to see if that would work but it's still just recognized as a `.jl` file. We could change the actual extension but that would be annoying to view as syntax highlighting would be lost. There is also an option of using CODEOWNERS but I'm not sure we want to tie these files to a group or user (note the user cannot be invalid).

omus:
I experimented with CODEOWNERS and it just lets you filter by files you own as a code owner which isn't very useful here. Ideally dotfiles would work but that feature currently seems broken.

PR #142, opened 2020-07-31
Handle hyphens in URI parameters

### LINKED ISSUE #141: Parse out hyphens for parameters in uris
There are a few services which specify parameters in their URIs with hyphens:

e.g. apigatewayv2 GetTags
> "/v2/tags/{resource-arn}"

When we convert this in the high-level API definition we need to properly parse out the hyphen but only when it is a string surrounded by `{ }` characters to ensure that we are not removing hyphens anywhere else in the URI.

The services which follow this format:
* apigatewayv2
* dataexchange
* greengrass
* iot 1click devices service
* medialive
* mediapackage vod
* mediapackage
* mq
* pinpoint
* schemas

mattBrzezinski:
Resolves #141 

# Problem
There are a few URIs where parameters contain hyphens in their parameters, such as:

> "/v2/tags/{resource-arn}"

Currently when we translate this to Julia code we end up with the following

> "/v2/tags/$(resource-arn)"

This however is not valid for interpolation, and we need to convert this hyphen to an underscore. However, we need to also account for hyphens being in URIs outside of parameters as this is also valid so we cannot do a blanket `replace(uri, '-' => '_')`. This merge request creates a new function which will find all matches in a URI for strings between `{` and `}` characters. It will then iterate over all these matches (parameters in the URI) and replace characters so we can translate this to valid Julia interpolation later on when generating high-level APIs.

# Notes
There sadly isn't a function for replacing by passing in a Dict or multiple Pairs, a discussion is created here for it:
- https://github.com/JuliaLang/julia/issues/29849


src/AWSMetadataUtilities.jl
```diff
@@ -184,6 +184,44 @@ function _documentation_cleaning(documentation::String)
...
+
+Replace URI parameters with the appropriate syntax for Julia interpolation.
+
+Find all URI parameters, and apply the following replacements:
+* { => \$(
+* } => )
+* - => _
+* + => nothing
+
+Example;
```
nicoleepp, line 198:
```suggestion
Example:
```

src/AWSMetadataUtilities.jl
```diff
@@ -184,6 +184,44 @@ function _documentation_cleaning(documentation::String)
...
+"/v1/configurations/{configuration-id}" => "/v1/configurations/\$(configuration_id)"
+
+# Arguments
+- `uri::String`: URI to be cleaned
+
+# Returns
+- `String`: Cleaned URI
+"""
+function _uri_cleaning(uri::String)
+    uri_parameters = eachmatch(r"{.*?}", uri) # Match anything surrounded in "{ }"
```
nicoleepp, line 208:
```suggestion
    uri_parameters = eachmatch(r"{.*?}", uri)  # Match anything surrounded in "{ }"
```

src/AWSMetadataUtilities.jl
```diff
@@ -184,6 +184,44 @@ function _documentation_cleaning(documentation::String)
...
+Example:
+"/v1/configurations/{configuration-id}" => "/v1/configurations/\$(configuration_id)"
+
+# Arguments
+- `uri::String`: URI to be cleaned
+
+# Returns
+- `String`: Cleaned URI
+"""
+function _uri_cleaning(uri::String)
```
iamed2, line 207:
```suggestion
function _clean_uri(uri::String)
```
Best practice for function naming is verbs

src/AWSMetadataUtilities.jl
```diff
@@ -184,6 +184,44 @@ function _documentation_cleaning(documentation::String)
...
+"""
+    _clean_uri(uri::String)
+
+Replace URI parameters with the appropriate syntax for Julia interpolation.
+
+Find all URI parameters, and apply the following replacements:
+* { => \$(
+* } => )
+* - => _
+* + => nothing
```
mjram0s, line 196:
```suggestion
* + => ""
```
This is more clear to me - otherwise it seems like you're replacing it with `nothing`. Also maybe throw in a test for this condition?

mjram0s:
```suggestion
* + => empty string
```
Or maybe this?

test/runtests.jl
```diff
@@ -4,7 +4,7 @@ using AWS: AWSServices
 using AWS.AWSExceptions: AWSException
 using AWS.AWSMetadataUtilities: _documentation_cleaning, _filter_latest_service_version,
     _generate_low_level_definition, _generate_high_level_definition, _generate_high_level_definitions,
-    _get_aws_sdk_js_files, _get_service_and_version, _get_function_parameters,
+    _get_aws_sdk_js_files, _get_service_and_version, _get_function_parameters, _uri_cleaning,
```
iamed2, line 7:
```suggestion
    _get_aws_sdk_js_files, _get_service_and_version, _get_function_parameters, _clean_uri,
```

test/AWSMetadataUtilities.jl
```diff
@@ -122,6 +122,36 @@ end
     @test result == expected_result
 end
 
+@testset "_uri_cleaning" begin
```
mjram0s, line 125:
```suggestion
@testset "__clean_uri" begin
```
I think all of these need to changed?

mjram0s: Can you add a test for replacing `+` with `""`?

PR #147, opened 2020-08-13
Add AWSConfig as a kwarg to high-level wrappers

### LINKED ISSUE #140: Add `AWSConfig` argument to generated service functions
AWSSDK has this and I think it's important: https://github.com/JuliaCloud/AWSSDK.jl/blob/master/src/ACM.jl#L61

mattBrzezinski:
### Overview
We want to add `AWSConfig` as a parameter to our high-level wrappers in-case the end user wants to use a different set of credentials when making a request.

This merge request adds `AWSConfig` as a kwarg to the high-level wrapper functions. It also modifies the existing request functions to remove `AWSConfig` as a positional argument and instead places it as a kwarg.

Using `RestXMLService` as an example we now have the following:
```julia
function (service::RestXMLService)(
    request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, String}(); 
    aws::AWSConfig=AWSConfig()
)
```

Which will create the two functions below, with `aws` as a kwarg being available in both:
```julia
function (service::RestXMLService)(request_method::String, request_uri::String)
function (service::RestXMLService)(request_method::String, request_uri::String, args::AbstractDict{String, <:Any})
```


nicoleepp:
Is there a reason you opted for a kwarg?

mattBrzezinski:
> Is there a reason you opted for a kwarg?

Yes, I wanted to keep the high-level generation as simple as possible by effectively not having to write a second set of functions with and without `AWSConfig` being a present positional argument. I also feel like using a specific set of credentials is more of an optional argument. At least in my experience, I have very rarely needed to use a different account / profile.

I will not, this is a bit inefficient with constructing an `AWSConfig` object for every call. We could do what `AWSCore.jl` currently does which is load in a global set of credentials and reference those. However, I'm not too concerned about that currently.

iamed2:
I think the kwarg is nice, I like that we can do that now that we don't pass kwargs.

I would like the default global config; I also think it should be settable with a function, so these would default to using the global config's state rather than the state of your whole environment (e.g. reinitializing credentials from scratch). Feel free to make that another PR but you'll probably have to change the service args again.

Bikeshed: I think we should call it something else, like `aws_config`

mattBrzezinski:
> Feel free to make that another PR but you'll probably have to change the service args again.

Good point, I'll just add a commit here for it.

mattBrzezinski:
Added a global AWSConfig in this commit, https://github.com/JuliaCloud/AWS.jl/pull/147/commits/a9405979f069ee0e6acdecf2025d0dc4438ec39a

Also, AWS just released a new service; [Braket](https://aws.amazon.com/braket/)

src/AWS.jl
```diff
@@ -401,7 +403,10 @@ Perform a RestXML request to AWS
 # Returns
 - The response from AWS
 """
-function (service::RestXMLService)(aws::AWSConfig, request_method::String, request_uri::String, args::AbstractDict{String, <:Any})
+function (service::RestXMLService)(
+    request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, String}(); 
+    aws::AWSConfig=aws_config
```
iamed2, line 408: I just noticed this one says AWSConfig but the others say AWS.AWSConfig. I think this version is fine and better

src/AWS.jl
```diff
@@ -23,7 +23,9 @@ using ..AWSExceptions
 using ..AWSExceptions: AWSException
 
 global user_agent = "AWS.jl/1.0.0"
+global aws_config = AWSConfig()
 
+set_aws_config(aws::AWSConfig) = return global aws_config = aws
```
iamed2, line 28: How about:

```suggestion
global_aws_config() = aws_config
global_aws_config(aws::AWSConfig) = return global aws_config = aws
```

src/AWS.jl
mattBrzezinski: Oopsies, good catch! I think I was just copying and pasting here to save time.

Changing this and committing it.

PR #148, opened 2020-08-14
The Big One

mattBrzezinski:
## Overview
Everything is in for `AWS.jl@1.0`, so this is the big merge into `master`! Even though this code has already been reviewed once before going into the `v1` branch, we still need to go over it once more with everything put together. This is the final say in terms for design (for the most part) before it's merged and tagged.

I fully expect this merge request to sit around for a week or so, but please review it and take your time. The best way to review this merge request in my opinion is to use the `Viewed File` feature in GitHub. When in the `Files Changed` tab, press the check box on the right-hand side to keep track of which files you have already viewed.

I do NOT recommend reviewing the `src/services/*.jl` files outside of the first one, these are auto-generated and contain the entire high-level definition for an AWS Service.


.github/workflows/update_apis.yml
```diff
@@ -0,0 +1,24 @@
...
+      matrix:
+        julia-version: [1] 
+        julia-arch: [x86]
+        os: [ubuntu-latest]
+    steps:
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: ${{ matrix.julia-version }}
+      - name: UpdateAPI
+        run: julia -e 'import Pkg; Pkg.develop(Pkg.PackageSpec(url="https://github.com/JuliaCloud/AWS.git")); using AWS; AWS.AWSMetadata.parse_aws_metadata();'
```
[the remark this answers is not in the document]
mattBrzezinski: Hmm.. would it make a difference? `master` should always be tagged to a specific version, and going forward it would all be `1+`.

README.md
```diff
@@ -1,68 +1,123 @@
...
 
-- EC2
-- S3
-- SQS
-- Auto Scaling
+## Overview
+A Julia interface for [Amazon Web Services](https://aws.amazon.com).
 
+This package is a simplification of both the low-level APIs from [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl) and the high level APIs from [AWSSDK.jl](https://github.com/JuliaCloud/AWSSDK.jl).
+It includes automated code generation to ensure all new AWS services are available, as well as keeping existing services up to date.
```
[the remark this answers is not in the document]
mattBrzezinski: I think the line above this one summarizes that?

> This package is a simplification of both the low-level APIs from [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl) and the high level APIs from [AWSSDK.jl](https://github.com/JuliaCloud/AWSSDK.jl).

Do you have any alternative suggestions on how to phrase it?

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+include("AWSCredentials.jl")
+include("AWSConfig.jl")
+include("AWSExceptions.jl")
+include("AWSMetadataUtilities.jl")
+
+using ..AWSExceptions
+using ..AWSExceptions: AWSException
+
+global user_agent = "AWS.jl/1.0.0"
+global aws_config = AWSConfig()
```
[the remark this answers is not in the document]
mattBrzezinski: I'm not quite sure I understand your suggestion here. Are you proposing to use `aws_config = Ref(AWSConfig())` instead of the above?

I just quickly read through the documentation on both, and it seems that this would work? I don't really know the underlying implications of doing this, is `Ref()` quicker?

The only thing I dislike about `Ref()` is:

```julia
julia> foo = Ref("bar")
Base.RefValue{String}("bar")

julia> foo
Base.RefValue{String}("bar")

julia> foo.x
"bar"
```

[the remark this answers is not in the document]
iamed2:
```
julia> foo = Ref("bar")
Base.RefValue{String}("bar")

julia> foo
Base.RefValue{String}("bar")

julia> foo[]
"bar"
```
The correct way to pull it out is indexing

`const` with `Ref` instead of `global` becomes type-inferrable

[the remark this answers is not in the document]
mattBrzezinski: Ah gotcha, I can update this. Not a fan of the syntax but it's fine.

I guess now that high-level functions can take in `AWSConfig` objects, we can remove setting the `global aws_config` and just constantly have that as a default.

Then the end user can just do something like:

```julia
using AWS.AWSServices: s3

bucket = "foo"
s3.ListObjects(bucket)  # Default

other_creds = AWSConfig(...)
s3.ListObjects(bucket; aws_config=other_creds)  # Other profile / account
```

.github/workflows/update_apis.yml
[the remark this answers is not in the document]
omus: If you ever want to change the name of `parse_aws_metadata` you may want to have a version set. That said you may want this to break if you forget to update it.

README.md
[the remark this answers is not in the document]
omus: The current documentation doesn't state that this package aims to be a replacement of those old packages instead of just an alternative:
```
AWS.jl provides both low-level and high-level APIs for interacting with AWS services.
The package includes automated code generation to ensure that new AWS services are available, as well as keeping existing services up to date. 

This package replaces [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl) and [AWSSDK.jl](https://github.com/JuliaCloud/AWSSDK.jl) which previously provided low-level and high-level APIs respectively.`
```

.github/workflows/update_apis.yml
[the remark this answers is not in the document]
omus: Maybe worth writing a comment that this is expected to run against the latest release if that is what you want. It probably does make sense to run this against `master` though an include `rev="master` as part of the `PackageSpec`.

.travis.yml
```diff
@@ -0,0 +1,24 @@
+language: julia
+os:
+- linux
+- osx
+- windows
```
omus, line 5:
```suggestion
  - linux
  - osx
  - windows
```

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+julia:
+- 1.0
+- 1
+- nightly
```
omus, line 9:
```suggestion
  - 1.0
  - 1
  - nightly
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
-        if ismatch(r"^aws_secret_access_key", segs[1])
-            AWS_SECKEY = segs[2]
-        elseif ismatch(r"^aws_access_key_id", segs[1])
-            AWS_ID = segs[2]
-        elseif ismatch(r"^region", segs[1])
-            AWS_REGION = segs[2]
-        end
-    end
+
+using Base64
```
omus, line 3: Dates is used in this source file but isn't being imported here
```suggestion
using Base64
using Dates
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+    request.content = HTTP.escapeuri(query)
+
+    return request
+end
+
+function _sign_aws4!(aws::AWSConfig, request::Request, time::DateTime)
+    # Create AWS Signature Version 4 Authentication Headers.
+    # http://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
+
+    date = Dates.format(time,"yyyymmdd")
```
omus, line 202:
```suggestion
    date = Dates.format(time, dateformat"yyyymmdd")
```

test/AWSCredentials.jl
```diff
@@ -0,0 +1,486 @@
...
+        "AccessKeyId" => "Default-Key",
+        "SecretAccessKey" => "Default-Secret",
+
+        "Test-AccessKeyId" => "Test-Key",
+        "Test-SecretAccessKey" => "Test-Secret",
+
+        "Token" => "Test-Token",
+        "InstanceProfileArn" => "Test-Arn",
+        "RoleArn" => "Test-Arn",
+        "Expiration" => now(),
```
omus, line 320: Should probably always use UTC right?
```suggestion
        "Expiration" => now(UTC),
```

test/AWSCredentials.jl
```diff
@@ -0,0 +1,486 @@
...
+    @testset "Helper functions" begin
+        @testset "Check Credentials - EnvVars" begin
+            withenv(
+                "AWS_ACCESS_KEY_ID" => test_values["AccessKeyId"],
+                "AWS_SECRET_ACCESS_KEY" => test_values["SecretAccessKey"]
+            ) do
+                testAWSCredentials = AWSCredentials(
+                    test_values["AccessKeyId"],
+                    test_values["SecretAccessKey"],
+                    expiry=Dates.now() - Minute(10),
```
omus, line 474:
```suggestion
                    expiry=Dates.now(UTC) - Minute(10),
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
+function _now_formatted()
+    return lowercase(Dates.format(now(Dates.UTC), "yyyymmddTHHMMSSsssZ"))
```
omus, line 2: I'm assuming you want a literal `Z` at the end
```suggestion
    return lowercase(Dates.format(now(Dates.UTC), dateformat"yyyymmdd\THHMMSSsss\Z"))
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+    aws = AWS.AWSConfig()
+
+    access_key = "access-key"
+    secret_key = "ssh... it is a secret"
+
+    aws.credentials.access_key_id = access_key
+    aws.credentials.secret_key = secret_key
+
+    time = DateTime(2020)
+    date = Dates.format(time, "yyyymmdd")
```
omus, line 38:
```suggestion
    date = Dates.format(time, dateformat"yyyymmdd")
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+        "x-amz-date" => datetime,
+        "Content-MD5" => base64encode(digest(MD_MD5, request.content))
+    ))
+
+    if !isempty(creds.token)
+        request.headers["x-amz-security-token"] = creds.token
+    end
+
+    # Sort and lowercase() Headers to produce canonical form...
+    canonical_headers = join(sort(["$(lowercase(k)):$(strip(v))" for (k,v) in request.headers]), "\n")
```
omus, line 235:
```suggestion
    canonical_headers = join(sort!(["$(lowercase(k)):$(strip(v))" for (k,v) in request.headers]), "\n")
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+    signed_headers = join(sort([lowercase(k) for k in keys(request.headers)]), ";")
```
omus, line 236:
```suggestion
    signed_headers = join(sort!([lowercase(k) for k in keys(request.headers)]), ";")
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+
+    # Sort Query String...
+    uri = HTTP.URI(request.url)
+    query = HTTP.URIs.queryparams(uri.query)
+    query = Pair[k => query[k] for k in sort(collect(keys(query)))]
```
omus, line 241:
```suggestion
    query = [k => query[k] for k in sort!(collect(keys(query)))]
```

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
+    string_to_sign = "AWS4-HMAC-SHA256\n$datetime\n$authentication_scope\n$canonical_hash"
+    signature = bytes2hex(digest(MD_SHA256, string_to_sign, signing_key))
+
+    # Append Authorization header...
+    request.headers["Authorization"] = string(
+        "AWS4-HMAC-SHA256 ",
+        "Credential=$(creds.access_key_id)/$authentication_scope, ",
+        "SignedHeaders=$signed_headers, ",
+        "Signature=$signature"
+    )
```
omus, line 265:
```suggestion
    request.headers["Authorization"] = join([
            "AWS4-HMAC-SHA256 Credential=$(creds.access_key_id)/$authentication_scope",
            "SignedHeaders=$signed_headers",
            "Signature=$signature",
        ],
        ", ",
    )
```

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
+using IniFile
+using HTTP
+using Dates
+using Mocking
+using JSON
```
omus, line 5:
```suggestion
using Dates
using HTTP
using IniFile
using JSON
using Mocking
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+
+Return a list of all AWS Services and their latest version.
+
+# Arguments
+- `services::Array`: Array of all AWS APIs and their versions
+
+# Returns
+- `Array{Dict}`: Array of the latest AWS Service definitions
+"""
+function _filter_latest_service_version(services::Array)
```
omus, line 43: Also update docstring
```suggestion
function _filter_latest_service_version(services::AbstractArray)
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+    seen_services = String[]
+    latest_versions = OrderedDict[]
+
+    services = reverse(services)
+
+    for service in services
```
omus, line 49:
```suggestion
    seen_services = Set{String}()
    latest_versions = OrderedDict[]

    for service in reverse(services)
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+
+# Returns
+- `Tuple{String, String}`: `service` and `version` for an AWS service
+"""
+function _get_service_and_version(filename::String)
+    try
+        service_and_version = join(split(filename, '.')[1:end-2],'.')
+        service_and_version = split(service_and_version, '-')
+        service = join(service_and_version[1:end-3], '-')
+        version = join(service_and_version[end-2:end], '-')
```
omus, line 77: State the intent to be clear what we want to do
```suggestion
        # Remove ".normal.json" suffix
        service_and_version = join(split(filename, '.')[1:end-2],'.')
        
        service_and_version = split(service_and_version, '-')
        service = join(service_and_version[1:end-3], '-')
        version = join(service_and_version[end-2:end], '-')
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+        if e isa BoundsError
+            throw(InvalidFileName("$filename is an invalid AWS JSON filename."))
+        else
+            rethrow()
+        end
+    end
+end
+
+"""
+    _generate_low_level_definitions(services::Array{OrderedDict{String, Any}})
```
omus, line 90: Doesn't match function

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+
+Get the low-level definitions for all AWS Services.
+
+# Arguments
+- `services::Array{OrderedDict}`: List of AWS Services to generate low-level definitions
+
+# Returns
+- `String[]`: Array of low-level service code to be written into `AWSServices.jl`
+"""
+function _generate_low_level_definitions(services::Array{OrderedDict})
```
omus, line 100:
```suggestion
function _generate_low_level_definitions(services::AbstractArray{OrderedDict})
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+
+        definition = _generate_low_level_definition(service_metadata)
+        push!(service_definitions, definition)
+    end
+
+    return service_definitions
+end
+
+"""
+    _generate_low_level_definition(service::Dict{String, Any})
```
omus, line 119:
```suggestion
    _generate_low_level_definition(service::AbstractDict{String, <:Any})
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+    if protocol == "rest-xml"
+        return "const $service_id = AWS.RestXMLService(\"$service_name\", \"$api_version\")"
+    elseif protocol in ("ec2", "query")
+        return "const $service_id = AWS.QueryService(\"$service_name\", \"$api_version\")"
+    elseif protocol == "rest-json"
+         if haskey(service_specifics, service_id)
+             return "const $service_id = AWS.RestJSONService(\"$service_name\", \"$api_version\", $(service_specifics[service_id]))"
+         end
+
+        return "const $service_id = AWS.RestJSONService(\"$service_name\", \"$api_version\")"
```
omus, line 153:
```suggestion
    elseif protocol == "rest-json" && haskey(service_specifics, service_id)
        return "const $service_id = AWS.RestJSONService(\"$service_name\", \"$api_version\", $(service_specifics[service_id]))"
    elseif protocol == "rest-json"
        return "const $service_id = AWS.RestJSONService(\"$service_name\", \"$api_version\")"
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+
+"""
+    _clean_documentation(documentation::String)
+
+Clean up the documentation to make it Julia compiler and human-readable.
+
+- Remove anything in-between <> HTML tags
+- Remove any dollar signs
+- Remove any \\
+- Replace " with \"
```
omus, line 171:
```suggestion
- Remove any HTML tags
- Remove any dollar signs
- Remove any backslashes
- Escape any double-quotes
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+- `documentation::String`: Documentation to be cleaned
+
+# Returns
+- `String`: Cleaned documentation string
+"""
+function _clean_documentation(documentation::String)
+    documentation = replace(documentation, r"\<.*?\>" => "")
+    documentation = replace(documentation, '$' => ' ')
+    documentation = replace(documentation, '\\' => ' ')
+    documentation = replace(documentation, '\"' => "\\\"")
```
omus, line 183:
```suggestion
    documentation = replace(documentation, '"' => "\\\"")
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+# Arguments
+- `documentation::String`: Documentation to be cleaned
+
+# Returns
+- `String`: Cleaned documentation string
+"""
+function _clean_documentation(documentation::String)
+    documentation = replace(documentation, r"\<.*?\>" => "")
+    documentation = replace(documentation, '$' => ' ')
+    documentation = replace(documentation, '\\' => ' ')
```
omus, line 182: According to the docstring these should be removed and not replace with space?

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+# Arguments
+- `input::String`: The input object for the given operation
+- `shapes::Dict{String, Any}`: Dictionary of all shapes for this AWS Service
+
+# Returns
+- `Tuple(Dict, Dict)`: (required_parameters, optional_parameters)
+"""
+function _get_function_parameters(input::String, shapes::Dict)
+    """
+        _get_parameter_name(parameter::String, input_shape::AbstractDict){String, <:Any}
```
omus, line 240:
```suggestion
        _get_parameter_name(parameter::String, input_shape::AbstractDict{String, <:Any})
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+        if haskey(input_shape["members"][parameter], "locationName")
+            return input_shape["members"][parameter]["locationName"]
+        end
+
+        # Check to see if the nested shape has a locationName
+        nested_shape = input_shape["members"][parameter]["shape"]
+        nested_member = get(shapes[nested_shape], "member", "")
+
+        # If nested_shape[member] exists return locationName (if exists), otherwise return the original parameter name
+        return !isempty(nested_member) ? get(nested_member, "locationName", parameter) : parameter
```
omus, line 262:
```suggestion
        nested_member = get(shapes[nested_shape], "member", Dict())

        # If nested_shape[member] exists return locationName (if exists), otherwise return the original parameter name
        return get(nested_member, "locationName", parameter)
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+        end
+    end
+
+    if haskey(input_shape, "members")
+        for parameter in input_shape["members"]
+            parameter_name = get(input_shape["members"][parameter[1]], "locationName", parameter[1])
+
+            if !haskey(required_parameters, parameter_name)
+                documentation = _clean_documentation(get(parameter[2], "documentation", ""))
+                idempotent = get(parameter[2], "idempotencyToken", false)
```
omus, line 291: Variable names could use some work beyond what I suggested:
```suggestion
        for (member_key, member_value) in input_shape["members"]
            parameter_name = get(input_shape["members"][member_key], "locationName", member_key)

            if !haskey(required_parameters, parameter_name)
                documentation = _clean_documentation(get(member_value, "documentation", ""))
                idempotent = get(member_value, "idempotencyToken", false)
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+function _generate_high_level_definitions(
+    service_name::String,
+    protocol::String,
+    operations::Dict,
+    shapes::Dict
+)
+    operation_definitions = String[]
+
+    for operation in operations
+        operation = operation[2]
```
omus, line 332:
```suggestion
    for (_, operation) in operations
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+    - `required_params::AbstractDict`: Required parameters for the function
+    - `optional_params::AbstractDict`: Optional parmaeters for the function
+    - `function_name::String`: Name of the function to generate code
+    - `service_name`: Name of the service associated with the function
+
+    # Returns
+    - `String`: Function definition as Julia code
+    """
+    function _generate_json_query_opeation_definition(required_params::AbstractDict, optional_params::AbstractDict, function_name::String, service_name::String)
+        req_keys = [replace(key, "-"=>"_") for key in collect(keys(required_params))]
```
omus, line 490:
```suggestion
        req_keys = [replace(key, '-' => '_') for key in collect(keys(required_params))]
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,578 @@
...
+        operation_definition *= repeat('"', 3)
+    end
+
+    operation_definition = _generate_docstring(name, documentation, required_parameters, optional_parameters)
+
+    if protocol in ("json", "query", "ec2")
+        operation_definition *= _generate_json_query_opeation_definition(required_parameters, optional_parameters, name, service_name)
+    elseif protocol in ("rest-json", "rest-xml")
+        operation_definition *= _generate_rest_operation_defintion(required_parameters, optional_parameters, name, service_name, method, request_uri)
+    end
```
omus, line 574: `else` error?

.gitattributes
```diff
@@ -0,0 +1,3 @@
+# https://docs.github.com/en/github/administering-a-repository/customizing-how-changed-files-appear-on-github
+deps/metadata.json linguist-generated=true
+src/services/* linguist-generated=true
```
omus, line 3: I missed this auto-generated file:
```suggestion
src/AWSServices.jl linguist-generated=true
src/services/* linguist-generated=true
```

test/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,448 @@
+function _clean_high_level_definition(definition::String)
+    # Required post Julia1.5 with how triple quoted strings are dealt with.
+    definition = replace(definition, " "=>"")
+    definition = replace(definition, "\n"=>"")
```
omus, line 4: Is the comment referring to Julia 1.5 and after or release after Julia 1.5?
```suggestion
    # Required post Julia 1.5 with how triple quoted strings are dealt with.
    definition = replace(definition, " " => "")
    definition = replace(definition, "\n" => "")
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+        expected_x_amz_date = Dates.format(time, dateformat"yyyymmdd\THHMMSS\Z")
+
+        result = AWS._sign_aws4!(aws, request, time)
+        headers = result.headers
+
+        @test headers["x-amz-content-sha256"] == expected_x_amz_content_sha256
+        @test headers["Content-MD5"] == expected_content_md5
+        @test headers["x-amz-date"] == expected_x_amz_date
+
+        authorization_header = split(headers["Authorization"], ' ')
```
omus, line 86:
```suggestion
        authorization_header = split(headers["Authorization"], ' ')
        @test length(authorization_header) == 4
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+
+                    @test headers == LittleDict(expected_headers)
+                    @test typeof(headers) <: LittleDict{SubString{String}, SubString{String}}
+               end
+            end
+        end
+
+        @testset "JSON" begin
+            json_headers = ["Content-Type"=>"application/json"]
+            json_body = """{"Marker":null,"VaultList":[{"CreationDate":"2020-06-22T03:14:41.754Z","LastInventoryDate":null,"NumberOfArchives":0,"SizeInBytes":0,"VaultARN":"arn:aws:glacier:us-east-1:000:vaults/test","VaultName":"test"}]}"""
```
omus, line 258: May be nicer to maintain if you write this as a `Dict` and use JSON.jl to turn it into a string

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+            expected_body_type = LittleDict{String, Any}
+            expected_body = JSON.parse(json_body, dicttype=LittleDict)
+
+            apply(Patches._aws_http_request_patch) do
+                Patches._response!(body=json_body, headers=json_headers,)
+
+                @testset "body" begin
+                    result = AWS.do_request(aws, request)
+
+                    @test typeof(result) <: expected_body_type
```
omus, line 269:
```suggestion
                    @test result isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+                    @test result == expected_body
+                end
+
+                @testset "body and headers" begin
+                    body, headers = AWS.do_request(aws, request; return_headers=true)
+
+                    @test body == expected_body
+                    @test typeof(body) <: expected_body_type
```
omus, line 277:
```suggestion
                    @test body isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,867 @@
...
+
+                    @test headers == LittleDict(json_headers)
+                    @test typeof(headers) <: LittleDict{SubString{String}, SubString{String}}
```
omus, line 280:
```suggestion
                    @test headers isa LittleDict{SubString{String}, SubString{String}}
```

omus:
Overall I'm good with proceeding with this. My review comments are mostly just cleaning up. I did notice that the code generation isn't really tested which is something we probably want to add but can be done at a later time.

src/AWSExceptions.jl
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    # Extract API error code from XML error message...
+    error_content_types = ["", "application/xml", "text/xml"]
+    if content_type(e) in error_content_types && length(http_message(e)) > 0
+        info = parse_xml(http_message(e))
+    end
+
+    info = get(info, "Errors", info)
+    info = get(info, "Error", info)
```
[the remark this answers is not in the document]
mattBrzezinski: From my understanding there are times when we get back `info["Errors"]["Error"]` and other times when we get back `info["Error"]`.

Just below is a similar situations with `(M|m)essage`.

src/AWSMetadata.jl
```diff
@@ -0,0 +1,111 @@
...
+
+Low level wrappers are written into `src/AWSServices.jl`, while high level wrappers API
+wrappers are written into their respective files in `src/services/{service}.jl`.
+"""
+function parse_aws_metadata()
+    function _process_service(file::OrderedDict{String, Any}, version::String)
+        data_changed = true
+        push!(metadata, file["name"] => Dict("version" => version, "sha" => file["sha"]))
+        push!(services_modified, file)
+    end
```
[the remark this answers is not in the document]
mattBrzezinski: I've been placing nested functions immediately after the first one to avoid something like:

```julia
function foo()
  bar = "bar"

  function baz(str::String)
    return string(str, str)
  end

  return baz(bar)
end
```

I can log an issue in BlueStyle to discuss this a bit more

src/AWSMetadata.jl
```diff
@@ -0,0 +1,111 @@
...
+
+    metadata_path = joinpath(@__DIR__, "..", "deps", "metadata.json")
```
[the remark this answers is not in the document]
mattBrzezinski: Is there a reason why? It's only used in this function.

src/AWSMetadataUtilities.jl
mattBrzezinski: I can add that in, since this is an internal function it shouldn't ever have an error since higher level functions would catch `protocol` not being valid.

test/_utilities.jl
```diff
@@ -0,0 +1,64 @@
+@testset "_merge - AbstractDict" begin
+    @testset "Simple" begin
+        expected = Dict("a" => 1, "b" => 2)
+
+        a = Dict("a"=>1)
+        b = Dict("b"=>2)
+
+        @test expected == AWS._merge(a, b)
+
```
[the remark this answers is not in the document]
mattBrzezinski: I'm not sure where I got this convention from, but all of the tests are written in the opposite form.

I logged an issue for this; https://github.com/JuliaCloud/AWS.jl/issues/154

src/AWSMetadata.jl
```diff
@@ -0,0 +1,111 @@
...
+        operations = sort(_generate_high_level_definitions(service_name, protocol, operations, shapes))
+
+        service_path = joinpath(@__DIR__, "services", "$service_name.jl")
+        open(service_path, "w") do f
+            println(f, "# This file is auto-generated by AWSMetadata.jl")
+            println(f, "using AWS")
+            println(f, "using AWS.AWSServices: $service_name\n")
+            println(f, "using Compat")
+            println(f, "using UUIDs")
+            print(f, join(operations, "\n"))
```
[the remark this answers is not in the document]
mattBrzezinski: Not sure I'm following this.

src/AWSMetadata.jl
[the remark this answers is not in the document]
mattBrzezinski: https://github.com/invenia/BlueStyle/issues/58

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+        env_var_credentials,
+        () -> dot_aws_credentials(profile),
+        () -> dot_aws_config(profile),
+        ecs_instance_credentials,
+        ec2_instance_credentials
+    ]
+
+    # Loop through our search locations until we get credentials back
+    for f in functions
+        credential_function = f
```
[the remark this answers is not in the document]
mattBrzezinski: I brought this in, but then I realized that `credential_function` is defined above and this seems to introduce it in a different scope and alter the behaviour.

I'll look into this some more if it can be simplfied.

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+    # the instance, Amazon does not document this file however so it's quite unreliable
+    # 5. Check `http://169.254.169.254`; This is a link-local address for metadata,
+    # apparently other cloud providers make this metadata URL available now as well so it's
+    # not guaranteed that you're on an EC2 instance
+    #    Or check a specific endpoint of the instance metadata such as:
+    #         ims_local_hostname = String(HTTP.get("http://169.254.169.254/latest/meta-data/local-hostname").body)
+    #    but with a fast timeout and cache the result.
+    #    See https://docs.aws.amazon.com/en_us/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+    # 6. When checking the UUID, check for little-endian representation,
+    # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html
```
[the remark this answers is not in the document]
mattBrzezinski: Originally when I wrote this back in `AWSCore.jl` it was decided to not include this in the documentation because the end user doesn't care about its implementation.

I think this can persist, as only maintainers will need to know this information.

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+    # all return "Amazon EC2" except the last two
+    sys_vendor = "/sys/devices/virtual/dmi/id/sys_vendor"
+    if _can_read_file(sys_vendor) && _ends_with_ec2(sys_vendor)
+        return true
+    end
+
+    return false
+end
+
+_can_read_file(file_name::String) = return isfile(file_name) && try isreadable(open(file_name, "r")) catch e; false; end
```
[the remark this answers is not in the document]
mattBrzezinski: https://github.com/JuliaCloud/AWS.jl/issues/155

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+_begins_with_ec2(file_name::String) = return uppercase(String(read(file_name, 3))) == "EC2"
+_ends_with_ec2(file_name::String) = return endswith(strip(uppercase(read(file_name, String))), "EC2")
```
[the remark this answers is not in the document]
mattBrzezinski: Yes, the file passed in here is `/sys/devices/virtual/dmi/id/sys_vendor` which from as far as I can tell is small.

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+
+    # This is not guarenteed to work on Windows as RNG can make the UUID begin with EC2 on a
+    # non-EC2 instance
+    if @mock Sys.iswindows()
+        command = `wmic path win32_computersystemproduct get uuid`
+        result = @mock Base.read(command, String)
+        instance_uuid = strip(split(result, "\n")[2])
+
+        return instance_uuid[1:3] == "EC2"
+    end
```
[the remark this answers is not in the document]
mattBrzezinski: I'm not quite understanding the benefits of `@static`, 

> https://github.com/JuliaLang/julia/blob/697e782ab86bfcdd7fd15550241fe162c51d9f98/base/osutils.jl#L3-L13

src/AWS.jl
```diff
@@ -1,235 +1,674 @@
...
-    else
-        ep_scheme = ep[1:(first(s)-1)]
-        ephp = ep[first(s)+3:end]
+function _generate_rest_resource(request_uri::String, args::AbstractDict{String, <:Any})
+    for (k,v) in args
+        if occursin("{$k}", request_uri)
+            request_uri = replace(request_uri, "{$k}" => v)
+        elseif occursin("{$k+}", request_uri)
+            request_uri = replace(request_uri, "{$k+}" => HTTP.escapepath(v))
+        end
```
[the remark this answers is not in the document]
mattBrzezinski: I have not seen an example for this, but it might be around.

I think that this is one of those limitations with the repository where something like this might happen, and until someone reports it we won't know. Unless we start writing tests for everything ourselves.

src/AWSCredentials.jl
```diff
@@ -0,0 +1,550 @@
...
+    println(io,
+        c.user_arn,
+        isempty(c.user_arn) ? "" : " ",
+        "(",
+        c.account_number,
+        isempty(c.account_number) ? "" : ", ",
+        c.access_key_id,
+        isempty(c.secret_key) ? "" : ", $(c.secret_key[1:3])...",
+        isempty(c.token) ? "" : ", $(c.token[1:3])...",
+        c.expiry,
```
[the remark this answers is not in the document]
mattBrzezinski: Only problem with `join` is if you have empty strings it'll just do `, ,`

src/AWSCredentials.jl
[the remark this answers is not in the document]
omus: `@static` allows for conditions to be evaulated at parse time which means you avoid having to call `Sys.iswindows()` each time this function is evaluated.

src/AWS.jl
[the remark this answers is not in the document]
omus: Ok, can we add a comment to mention that that we're aware of this limitation?

src/AWSExceptions.jl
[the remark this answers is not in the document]
omus: Ok, definitely worth leaving a comment about expecting these kinds of inputs. Ideally we'd also have test for both

src/AWSMetadata.jl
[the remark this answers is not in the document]
omus: Just for maintainability. If you move or rename this file it'll be harder to miss as it'll probably be located with other file constants.

src/AWSMetadata.jl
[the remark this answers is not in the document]
omus: A demonstration will clear things up:
```julia
julia> operations = ["a", "b", "c"]
3-element Array{String,1}:
 "a"
 "b"
 "c"

julia> f = stdout
Base.TTY(RawFD(0x0000000e) open, 0 bytes waiting)

julia> print(f, join(operations, "\n"))
a
b
c
julia> join(f, operations, "\n")
a
b
c
```
The `join` function has a method which can write to `IO` directly instead of first having to generate a `String`.

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,575 @@
...
+    req = @mock HTTP.get(url, headers)
+    files = JSON.parse(String(req.body), dicttype=OrderedDict)
+    filter!(f -> endswith(f["name"], ".normal.json"), files)  # Only get ${Service}.normal.json files
+    files = _filter_latest_service_version(files)
+
+    return files
+end
+
+"""
+    _filter_latest_service_version(services::Array{Any})
```
omus, line 31:
```suggestion
    _filter_latest_service_version(services::AbstractArray)
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,575 @@
...
+
+Return a list of all AWS Services and their latest version.
+
+# Arguments
+- `services::Array`: Array of all AWS APIs and their versions
+
+# Returns
+- `Array{Dict}`: Array of the latest AWS Service definitions
```
omus, line 39:
```suggestion
- `services::AbstractArray`: Array of all AWS APIs and their versions

# Returns
- `Vector{OrderedDict}`: Array of the latest AWS Service definitions
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,575 @@
...
+"""
+    _generate_low_level_definitions(services::AbstractArray{OrderedDict})
+
+Get the low-level definitions for all AWS Services.
+
+# Arguments
+- `services::Array{OrderedDict}`: List of AWS Services to generate low-level definitions
+
+# Returns
+- `String[]`: Array of low-level service code to be written into `AWSServices.jl`
```
omus, line 96:
```suggestion
- `services::AbstractArray{OrderedDict}`: List of AWS Services to generate low-level definitions

# Returns
- `Vector{String}`: Array of low-level service code to be written into `AWSServices.jl`
```

src/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,575 @@
...
+# Arguments
+- `service::Dict{String, Any}`: AWS Service to generate the low-level code definition
+
+# Returns
+- `String`: Low-level definition for the AWS Service
+
+# Throws
+- `ProtocolNotFound`: If `service` is using a protocol that's not either `[rest-xml, rest-json, json, query]`
+"""
+function _generate_low_level_definition(service::Dict)
```
omus, line 130:
```suggestion
    _generate_low_level_definition(service::AbstractDict)

Get the low-level definition for an AWS Service.

# Arguments
- `service::AbstractDict`: AWS Service to generate the low-level code definition

# Returns
- `String`: Low-level definition for the AWS Service

# Throws
- `ProtocolNotFound`: If `service` is using a protocol that's not either `[rest-xml, rest-json, json, query]`
"""
function _generate_low_level_definition(service::AbstractDict)
```

test/AWSMetadataUtilities.jl
```diff
@@ -0,0 +1,448 @@
+function _clean_high_level_definition(definition::String)
+    # Required post Julia 1.5 with how triple quoted strings are dealt with.
```
omus, line 2: Is the comment referring to Julia 1.5 and after or release after Julia 1.5?

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+            request_method="HEAD",
+            url="https://s3.us-east-1.amazonaws.com/sample-bucket"
+        )
+
+        apply(Patches._aws_http_request_patch) do
+            Patches._response!()
+            result = AWS.submit_request(aws, request)
+
+            @test result == expected_result_type(Patches.headers)
+            @test typeof(result) <: expected_result_type
```
omus, line 103:
```suggestion
            @test result isa expected_result_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+                    expected_body = xml_dict((Patches.body), expected_body_type)
+
+                    expected_header_type = LittleDict{SubString{String}, SubString{String}}
+
+                    Patches._response!(headers=expected_headers)
+
+                    @testset "body" begin
+                        result = AWS.submit_request(aws, request)
+
+                        @test typeof(result) <: expected_body_type
```
omus, line 196:
```suggestion
                        @test result isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+                    end
+
+                    @testset "body and headers" begin
+                        body, headers = AWS.submit_request(aws, request; return_headers=true)
+
+                        @test body == expected_body
+                        @test typeof(body) <: expected_body_type
+
+                        @test headers == LittleDict(expected_headers)
+                        @test typeof(headers) <: expected_header_type
```
omus, line 207:
```suggestion
                        @test headers isa expected_header_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+
+                        @test typeof(result) <: expected_body_type
+                        @test result == expected_body
+                    end
+
+                    @testset "body and headers" begin
+                        body, headers = AWS.submit_request(aws, request; return_headers=true)
+
+                        @test body == expected_body
+                        @test typeof(body) <: expected_body_type
```
omus, line 204:
```suggestion
                        @test body isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+            expected_headers = Pair["Content-Type"=>"application/xml",]
+
+            apply(Patches._aws_http_request_patch) do
+                Patches._response!(headers=expected_headers,)
+
+                @testset "body" begin
+                    result = AWS.submit_request(aws, request)
+
+                    @test result == expected_body
+                    @test typeof(result) <: expected_body_type
```
omus, line 232:
```suggestion
                    @test result isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+                end
+
+                @testset "body and headers" begin
+                    body, headers = AWS.submit_request(aws, request; return_headers=true)
+
+                    @test body == expected_body
+                    @test typeof(body) <: expected_body_type
```
omus, line 239:
```suggestion
                    @test body isa expected_body_type
```

test/AWS.jl
```diff
@@ -0,0 +1,869 @@
...
+
+                    @test headers == LittleDict(expected_headers)
+                    @test typeof(headers) <: LittleDict{SubString{String}, SubString{String}}
```
omus, line 242:
```suggestion
                    @test headers isa LittleDict{SubString{String}, SubString{String}}
```

src/AWS.jl
```diff
@@ -1,235 +1,663 @@
...
+    return _flatten_query!(Pair{String, String}[], service, query)
+end
 
-        url *= iam[1]
-        resp = Requests.get(url)
-        if resp.status != 200
-            return nothing
+function _flatten_query!(result::Vector{Pair{String, String}}, service::String, query::AbstractDict{String, <:Any}, prefix::String="")
+    for (k, v) in query
+        if typeof(v) <: AbstractDict
```
omus, line 436:
```suggestion
        if v isa AbstractDict
```

src/AWS.jl
```diff
@@ -1,235 +1,663 @@
...
+            _flatten_query!(result, service, v, string(prefix, k, "."))
+        elseif typeof(v) <: AbstractArray
```
omus, line 438:
```suggestion
        elseif v isa AbstractArray
```

src/AWS.jl
```diff
@@ -1,235 +1,663 @@
...
+            for (i, j) in enumerate(v)
+                suffix = service in ("ec2", "sqs") ? "" : ".member"
+                prefix_key = string(prefix, k, suffix, ".", i)
+
+                if typeof(j) <: AbstractDict
```
omus, line 443:
```suggestion
                if j isa AbstractDict
```

test/AWSMetadataUtilities.jl
mattBrzezinski: This is for Julia 1.5 or higher, adjusted it.

PR #149, opened 2020-08-17
Mock getting AWS SDK JS files

mattBrzezinski:
## Problem
Currently we are making unauthorized requests to GitHub when retrieving the [AWS SDK JS]() for generating our Julia definitions off of. Unauthenticated requests are limited to 60 requests per hour, I've run into the rate limitation personally and we are now seeing that with Travis CI.

## Solutions
There are two solutions to this problem, the first is creating authenticated requests. I attempted to configure an OAuth application on GitHub to make authorized requests, it was very simple. The problem comes how do we make the request in Julia.

I first attempted to do run an external command to do `curl -u id:secret https://github.com/aws/aws-sdk-js/tree/master/apis`, this was fine. Except that running this command prints the output into the terminal and causes it to hang. I then looked into using [LibCURL.jl](https://github.com/JuliaWeb/LibCURL.jl), however this seems like overkill for one test.

The other solution is in this merge request, just simply mock the response back from GitHub and continue onwards with testing.

### DIFF
--- a/test/patch.jl
+++ b/test/patch.jl
@@ -50,8 +50,23 @@ function _response!(; version::VersionNumber=version, status::Int64=status, head
     return response
 end
 
-_http_request_patch = @patch function AWS._http_request(request::Request)
+_aws_http_request_patch = @patch function AWS._http_request(request::Request)
     return response
+end
+
+
+aws_sdk_files =
+    """
+    [
+        {"name": "test-2020-01-01.normal.json"},
+        {"name": "test-2020-01-01.min.json"},
+        {"name": "test-2020-01-01.paginators.json"},
+        {"name": "test-2020-01-01.examples.json"}
+    ]
+    """
+
+_http_get_patch = @patch function HTTP.get(url::String, headers)
+    return HTTP.Response(200, aws_sdk_files)
 end
 
 end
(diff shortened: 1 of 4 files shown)


test/patch.jl
```diff
@@ -50,8 +50,23 @@ function _response!(; version::VersionNumber=version, status::Int64=status, head
...
 end
 
+
+aws_sdk_files =
+    """
+    [
+        {"name": "test-2020-01-01.normal.json"},
+        {"name": "test-2020-01-01.min.json"},
+        {"name": "test-2020-01-01.paginators.json"},
+        {"name": "test-2020-01-01.examples.json"}
```
TylerLoewen, line 64:
```suggestion
        {"name": "test-2020-01-01.examples.json"},
```

mattBrzezinski: The JSON specification doesn't allow for trailing commas.

iamed2:
For Authenticated requests: https://github.com/JuliaWeb/GitHub.jl#authentication

But I think the solution here is fine

mattBrzezinski:
> For Authenticated requests: https://github.com/JuliaWeb/GitHub.jl#authentication
> 
> But I think the solution here is fine

Oh interesting I didn't know this package existed!

PR #151, opened 2020-08-25
DNM: Experiment with .gitattributes

omus:
Trying to hide generated code files from the diff by default.


omus:
It works!

> Some generated files are not loaded by default

PR #157, opened 2020-08-30
Properly throw errors if credentials are not found

mattBrzezinski:
# Overview
One overlooked thing which I hadn't considered is if users attempt to use this package without having any AWS Credentials configured on their machine.

An example of this is in my attempt to add this package to the [General Julia Registry](https://github.com/JuliaRegistries/General/pull/20506). The AutoMerge step for the registrator uses Microsoft Azure instances, these internally use the IP address `169.254.169.254`. When looking for AWS Credentials, our last attempt is to use the [instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/instancedata-data-retrieval.html). We expect that if you're not on an AWS instance you will receive a `TimeoutError`. However, running this on an Azure instance will return back a 404 error.

# Solution
Check the returned value from attempting to get instance metadata if it's `nothing`.

# Notes
Added a test to replicate a "clean" instance, one without:

- Environment Variables for AWS Credentials
- No `config` or `credential` files
- Returns back `nothing` when querying the instance metadata endpoints

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -295,7 +295,7 @@ function _ec2_metadata(metadata_endpoint::String)
     try
         request = @mock HTTP.request("GET", "http://169.254.169.254/latest/meta-data/$metadata_endpoint")
 
-        return String(request.body)
+        return request isa Nothing ? nothing : String(request.body)
     catch e
         e isa HTTP.IOError || rethrow(e)
     end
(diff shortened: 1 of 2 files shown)


src/AWSCredentials.jl
```diff
@@ -295,7 +295,7 @@ function _ec2_metadata(metadata_endpoint::String)
     try
         request = @mock HTTP.request("GET", "http://169.254.169.254/latest/meta-data/$metadata_endpoint")
 
-        return String(request.body)
+        return request isa Nothing ? nothing : String(request.body)
```
iamed2, line 298:
```suggestion
        return request === nothing ? nothing : String(request.body)
```

src/AWSCredentials.jl
```diff
@@ -311,6 +311,11 @@ Parse the EC2 metadata to retrieve AWSCredentials.
 """
 function ec2_instance_credentials()
     info = _ec2_metadata("iam/info")
+
+    if info isa Nothing
```
iamed2, line 315:
```suggestion
    if info === nothing
```

PR #158, opened 2020-08-30
Instance Metadata, catch Azure 404s

mattBrzezinski:


iamed2:
Looks like the docstring needs an update regarding error behaviour

PR #163, opened 2020-08-31
Use create-pull-request@v3

mattBrzezinski:
Attempting to test the GH Action to auto generate files from a different repository as to not spam this one with merge requests. It's however running into some issues, which I've logged https://github.com/peter-evans/create-pull-request/issues/534.

It seems like Peter is in Tokyo and it's currently sleeping time there, so I'll check in on the issue later on in the evening. In the meantime these seem to be two useful things to changed:

- Update to `create-pull-request@v3` and use the latest version

### DIFF
--- a/.github/workflows/update_apis.yml
+++ b/.github/workflows/update_apis.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: julia-actions/julia-buildpkg@latest
       - name: UpdateAPI
         # Run against the latest version, since everything merged into master will be tagged.
-        run: sudo julia --project -e 'using AWS; AWS.AWSMetadata.parse_aws_metadata();'
+        run: julia --project -e 'using AWS; AWS.AWSMetadata.parse_aws_metadata();'
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
--- a/.github/workflows/update_apis.yml
+++ b/.github/workflows/update_apis.yml
@@ -13,8 +13,6 @@ jobs:
         os: [ubuntu-latest]
     steps:
       - uses: actions/checkout@v2
-        with:
-          repository: JuliaCloud/AWS.jl
       - uses: julia-actions/setup-julia@latest
         with:
           version: ${{ matrix.julia-version }}
@@ -25,7 +23,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
-         commit-message: AWS API Definitions Updated
-         reviewers: mattBrzezinski
-         title: AWS API Definitions Updated
-         token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: AWS API Definitions Updated
+          reviewers: mattBrzezinski
+          title: AWS API Definitions Updated
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


iamed2:
> Add sudo because writing to local files on the GHA without results in a permissions denied error

Are you attempting to write to the package directory? This could be unrelated to GitHub Actions.

mattBrzezinski:
> > Add sudo because writing to local files on the GHA without results in a permissions denied error
> 
> Are you attempting to write to the package directory? This could be unrelated to GitHub Actions.

Yeah, basically what happens is we pull down the repo. Then run the `parse_aws_metadata()` function which will update the API definitions on the CI. It will then use the `create-pull-request` GHA to make a PR back into this repository with the changes.

The problem comes when files need to be updated, we need to write them so we can `git commit`. But without `sudo` on the CI you'll get an error saying you don't have permissions to write to the local file.

See, [here](https://github.com/mattBrzezinski/JuliaCloud-AWS-State/runs/1052070017?check_suite_focus=true#step:3:325)

iamed2:
Oh I'm sorry, you absolutely do need Pkg.develop here and my Pkg.add suggestion was wrong then

mattBrzezinski:
> Oh I'm sorry, you absolutely do need Pkg.develop here and my Pkg.add suggestion was wrong then

No worries! Maybe I'll just play around with this in my other repo while waiting for a response on https://github.com/peter-evans/create-pull-request/issues/534 and we can just group all changes up in this PR.

mattBrzezinski:
@iamed2 all fixed up now!

.github/workflows/update_apis.yml
```diff
@@ -12,14 +12,20 @@ jobs:
         julia-arch: [x86]
         os: [ubuntu-latest]
     steps:
+      - uses: actions/checkout@v2
+        with:
+          repository: JuliaCloud/AWS.jl
```
peter-evans, line 17: I thought you needed to create a PR in a remote repository, but I now realise you were just doing that for test purposes. Because you are checking out the current repository you don't need to specify `repository` here.

Just this line on it's own is fine:
```
- uses: actions/checkout@v2
```

.github/workflows/update_apis.yml
```diff
@@ -12,14 +12,20 @@ jobs:
...
-        run: julia -e 'import Pkg; Pkg.add(Pkg.PackageSpec(url="https://github.com/JuliaCloud/AWS.jl.git", rev="master")); using AWS; AWS.AWSMetadata.parse_aws_metadata();'
+        run: sudo julia --project -e 'using AWS; AWS.AWSMetadata.parse_aws_metadata();'
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v2
+        uses: peter-evans/create-pull-request@v3
         with:
-         path: "~/.julia/dev/AWS/"
+         commit-message: AWS API Definitions Updated
+         reviewers: mattBrzezinski
+         title: AWS API Definitions Updated
```
peter-evans, line 30: The lines below `with:` are missing one whitespace character for indentation.

mattBrzezinski:
Just tested @peter-evans code changes in my test repo, works just as well! Thank you again for the feedback, it's very much appreciated!

iamed2:
Are you sure you still need `sudo` if you're no longer adding the repo?

mattBrzezinski:
Just tested, `sudo` is not required. I've changed this in `76c98fa`.

PR #164, opened 2020-08-31
Minor fixes

mattBrzezinski:
1. Properly reference the global AWSConfig, in each protocol function
With the changes to the default `AWSConfig`, if you do not call `AWS.global_aws_config()`, and you perform a low-level request you will get an error that the Reference is not assigned.

1.  Remove response_dict_type from args if it's available
If you pass in a specified `response_dict_type` in a REST protocol function it will attempt to iterate over this type which will error. We just need to `delete!` it from `args` after we assign it to the `Request`.

1. Use response_dict_type in Query and JSON services
Missed adding this `Request` parameter for Query and JSON protocol requests.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -500,8 +500,7 @@ function (service::RestXMLService)(
     )
 
     if haskey(args, "response_dict_type")
-        request.response_dict_type = args["response_dict_type"]
-        delete!(args, "response_dict_type")
+        request.response_dict_type = pop!(args, "response_dict_type")
     end
 
     delete!(args, "headers")
@@ -564,8 +563,7 @@ function (service::QueryService)(
     )
 
     if haskey(args, "response_dict_type")
-        request.response_dict_type = args["response_dict_type"]
-        delete!(args, "response_dict_type")
+        request.response_dict_type = pop!(args, "response_dict_type")
     end
 
     request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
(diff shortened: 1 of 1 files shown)


src/AWS.jl
```diff
@@ -501,6 +501,7 @@ function (service::RestXMLService)(
 
     if haskey(args, "response_dict_type")
         request.response_dict_type = args["response_dict_type"]
+        delete!(args, "response_dict_type")
```
iamed2, line 504:
```suggestion
        request.response_dict_type = pop!(args, "response_dict_type")
```

src/AWS.jl
```diff
@@ -562,6 +563,11 @@ function (service::QueryService)(
         return_raw=get(args, "return_raw", false),
     )
 
+    if haskey(args, "response_dict_type")
+        request.response_dict_type = args["response_dict_type"]
+        delete!(args, "response_dict_type")
```
iamed2, line 568:
```suggestion
        request.response_dict_type = pop!(args, "response_dict_type")
```

src/AWS.jl
```diff
@@ -610,6 +616,11 @@ function (service::JSONService)(
...
```
iamed2, line 621:
```suggestion
        request.response_dict_type = pop!(args, "response_dict_type")
```

src/AWS.jl
```diff
@@ -655,6 +666,7 @@ function (service::RestJSONService)(
 
     if haskey(args, "response_dict_type")
         request.response_dict_type = args["response_dict_type"]
+        delete!(args, "response_dict_type")
```
iamed2, line 669:
```suggestion
        request.response_dict_type = pop!(args, "response_dict_type")
```

PR #165, opened 2020-09-01
Cleaned up RestXML and Query responses

mattBrzezinski:
## Overview
While re-writing the code for this package we wanted to make responses back be Dictionaries.
We currently use `XMLDict.xml_dict()`, which appends two metadata pieces of information back, the XML `version` and `encoding`.

```julia
EC2.describe_volume_status()
OrderedCollections.LittleDict{Union{String, Symbol},Any,Array{Union{String, Symbol},1},Array{Any,1}} with 3 entries:
  :version                       => "1.0"
  :encoding                      => "UTF-8"
  "DescribeVolumeStatusResponse" => OrderedCollections.LittleDict{…]
```

This adds an unnecessary level of abstraction as the user would then need to do `result["DescribeVolumeStatusResponse"]` to get the actual AWS information.
This also means that you cannot directly replace `AWSCore.jl` with `AWS.jl` and have a smooth transition over.

## Solution
Replicate what `XMLDict` does and only return back the important information.

```julia
EC2.describe_volume_status()
OrderedCollections.LittleDict{Union{String, Symbol},Any,Array{Union{String, Symbol},1},Array{Any,1}} with 2 entries:
  "requestId"       => "19c981b2-8701-4972-aa53-100272d7cc0a"
  "volumeStatusSet" => OrderedCollections.LittleDict{Union{String, Symbol},Any,Array{Union{String, Symbol},1},Array{Any,1}}("item"=>OrderedCollections.LittleDict{Union{String, Symbol},Any,Array{Union{String, Symbol},…
```

### DIFF
--- a/test/AWS.jl
+++ b/test/AWS.jl
@@ -85,6 +85,11 @@ end
 @testset "submit_request" begin
     aws = AWS.AWSConfig()
 
+    function _expected_body(body::AbstractString, dict_type::Type)
+        parsed = parse_xml(body)
+        return xml_dict(XMLDict.root(parsed.x), dict_type)
+    end
+
     @testset "HEAD request method" begin
         expected_result_type = LittleDict
 
@@ -182,9 +187,9 @@ end
                 end
 
                 @testset "text/xml" begin
-                    expected_headers = Pair["Content-Type"=>"",]
+                    expected_headers = Pair["Content-Type"=>"text/xml",]
                     expected_body_type = LittleDict{Union{Symbol, String}, Any}
-                    expected_body = xml_dict((Patches.body), expected_body_type)
+                    expected_body = _expected_body(Patches.body, expected_body_type)
 
                     expected_header_type = LittleDict{SubString{String}, SubString{String}}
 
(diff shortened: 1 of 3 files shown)


omus:
What happens in the scenario where exception occurs? Is the exception in the response or is it nested in the result?

test/AWS.jl
```diff
@@ -501,11 +506,11 @@ end
         expected_policy_document = JSON.json(expected_policy_document)
 
         response = IAM.create_policy(expected_policy_document, expected_policy_name)
-        policy_arn = policy_arn = response["CreatePolicyResponse"]["CreatePolicyResult"]["Policy"]["Arn"]
+        policy_arn = policy_arn = response["CreatePolicyResult"]["Policy"]["Arn"]
```
omus, line 509:
```suggestion
        policy_arn = response["CreatePolicyResult"]["Policy"]["Arn"]
```

mattBrzezinski:
> What happens in the scenario where exception occurs? Is the exception in the response or is it nested in the result?

Messaged on Slack, but to put it here. From what I've been testing it seems that exceptions are unaffected. I did some off-hand testing with creating an S3 bucket in `us-east-2`, and attempting to list the objects in it while using the region `us-east-1`.

This functioned as expected, and the error message was still returned back in the response. These changes only purge the XML `version` and `encoding`, AFAIK. However I did not overtly test with with a large variety of services, just a few.

PR #170, opened 2020-09-02
High-level definition Compat fix

mattBrzezinski:
## Overview
When attempting to use high-level definitions you get the error:

```julia
julia> using AWS

julia> @service S3
ERROR: LoadError: ArgumentError: Package Compat not found in current path:
- Run `import Pkg; Pkg.add("Compat")` to install the Compat package.
```

Tests were working successfully because we were including `Compat` in `runtests.jl`. It was also working for me locally because I already had `Compat` in my Julia environment.

I was able to replicate this issue by removing `Compat` from my Julia environment, and re-running the commands in the logged issue.

## Solution
Introduce `Compat` in the `module AWS`, and reference it in each high-level definition as `using AWS: Compat`.
To test changes I re-ran the commands with the proposed changes, and `Compat` removed from my Julia environment.

I suspect this has to do with the macro below and what is and is not available: https://github.com/JuliaCloud/AWS.jl/blob/f55efaa72ade5c05dcdf955d05df9c9364a97a75/src/AWS.jl#L94-L101


src/AWSMetadata.jl
```diff
@@ -99,8 +99,8 @@ function _generate_high_level_wrapper(services::AbstractArray{<:OrderedDict})
         open(service_path, "w") do f
             println(f, "# This file is auto-generated by AWSMetadata.jl")
             println(f, "using AWS")
+            println(f, "using AWS: Compat")
```
omus, line 102: You probably want to have all of the Compat exports included here:
```suggestion
            println(f, "using AWS.Compat")
```

src/AWS.jl
```diff
@@ -1,5 +1,6 @@
 module AWS
 
+using Compat
```
omus, line 3: Just include the module as that is all you need here
```suggestion
using Compat: Compat
```

mattBrzezinski:
Just going to regenerated the definitions, and squash them into the existing commit.

src/AWSMetadata.jl
```diff
@@ -99,8 +99,8 @@ function _generate_high_level_wrapper(services::AbstractArray{<:OrderedDict})
         open(service_path, "w") do f
             println(f, "# This file is auto-generated by AWSMetadata.jl")
             println(f, "using AWS")
+            println(f, "using AWS.Compat")
             println(f, "using AWS.AWSServices: $service_name\n")
-            println(f, "using Compat")
             println(f, "using UUIDs")
```
omus, line 104: I think this will also cause you grief later but only when AWS.jl is used within another package.

omus: Yes, this does seem to be an issue:
```
┌ Warning: Package X does not have UUIDs in its dependencies:
│ - If you have X checked out for development and have
│   added UUIDs as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with X
└ Loading UUIDs into X from project dependency, future warnings for X are suppressed.
```

mattBrzezinski: Hmmm I'm not able to get that error message. I've made a minimal example project to play around with this issue:

`Project.toml`
```julia
name = "Playground"
uuid = "9341520b-331d-57cd-baf0-c27f46988a5f"

[deps]
Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"

[compat]
Compat = "3"
julia = "1"

[extras]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"

[targets]
test = ["Test", "UUIDs"]
```

`src/Playground.jl`
```julia
module Playground
using Compat: Compat
export @service

macro service(module_name::Symbol)
    service_name = joinpath(@__DIR__, lowercase(string(module_name)) * ".jl")

    return Expr(:toplevel,
    :(module($(esc(module_name)))
        Base.include($(esc(module_name)), $(esc(service_name)))
     end))
end
end
```

`src/compat_test.jl`
```julia
using UUIDs
using Playground: Compat

println(uuid4())
```

Then from the REPL
```julia
julia> using Playground

julia> @service compat_test
c3728c76-a2d5-438a-b950-dbd1ac574466
Main.compat_test
```

omus: Try calling `@service compat_test` from within `src/Playground.jl`

omus:
```
name = "Playground"
uuid = "..."

[deps]
AWS = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"

[compat]
AWS = "1"
julia = "1"
```
```
module AnotherPackage
using AWS
@service S3
end
```

mattBrzezinski: Ah I see. This only seems to be a problem from within `src/Playground.jl` though?

Creating another file: `external.jl`
```julia
using Playground
@service compat_test
```

```bash
Playground $ julia external.jl
d5481c07-3c5d-4dd8-a853-dc64df63f9d5
```

mattBrzezinski: Just saw your updated comment, yeah playing around I can see that issue popping up...

omus: I believe this restriction is only for packages. Julia wants a package to declare all dependencies it directly uses. As the `@service` macro adds code that uses UUIDs wherever it is expanded that package is now a direct dependency of UUIDs. The reason you probably weren't seeing the failure consistently is that since UUIDs is part of the stdlib things like scripts can use it without it being explicitly added to your environment manifest.

src/AWSMetadata.jl
```diff
@@ -100,8 +100,8 @@ function _generate_high_level_wrapper(services::AbstractArray{<:OrderedDict})
             println(f, "# This file is auto-generated by AWSMetadata.jl")
             println(f, "using AWS")
             println(f, "using AWS.AWSServices: $service_name\n")
```
omus, line 102: The separation between the imports isn't necessary anymore
```suggestion
            println(f, "using AWS.AWSServices: $service_name")
```

omus (commented):
I'd add a test which makes a dummy package to ensure we don't reintroduce this mistake in the future.

test/test_pkg.jl
```diff
@@ -0,0 +1,17 @@
...
+            redirect_stderr(err_io) do 
+                @eval using TestPkg
+            end
+        end
+
+        @test isempty(read(err_file))
+    end
+
+    @test isempty(read(out_file))
+end
```
omus, line 22: You may just want to use `@capture_out` and `@capture_err` from https://github.com/JuliaIO/Suppressor.jl

omus (approved):

test/test_pkg.jl
mattBrzezinski: I saw this, it does make things a bit cleaner so I added it in!

PR #180, opened 2020-09-07
Pass along kwargs for global_aws_config; set_global_aws_config()

mattBrzezinski:
## Overview
By default we are always setting the `AWSConfig.region` to `us-east-1`. Some people don't use this region, we should give them the option to set the default region by passing `kwargs...` along to the underlying `AWSConfig()` call.

Also included a function `set_global_aws_config(config::AWSConfig)` to set the global configuration.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -13,7 +13,7 @@ using UUIDs: UUIDs
 using XMLDict
 
 export @service
-export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_global_aws_config, set_user_agent
+export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_user_agent
 export JSONService, RestJSONService, RestXMLService, QueryService
 
 include("utilities.jl")
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -42,7 +42,7 @@ If one is not set, create one with default configuration options.
 - `AWSConfig`: The global AWS configuration
 """
 function global_aws_config(; kwargs...)
-    if !isassigned(aws_config)
+    if !isassigned(aws_config) || !isempty(kwargs)
         aws_config[] = AWSConfig(; kwargs...)
     end
 
@@ -61,7 +61,7 @@ Set the global AWSConfig.
 # Returns
 - `AWSConfig`: Global AWSConfig
 """
-function set_global_aws_config(config::AWSConfig)
+function global_aws_config(config::AWSConfig)
     return aws_config[] = config
 end
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -51,7 +51,7 @@ end
 
 
 """
-    set_global_aws_config(config::AWSConfig)
+    global_aws_config(config::AWSConfig)
 
 Set the global AWSConfig.
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -13,7 +13,7 @@ using UUIDs: UUIDs
 using XMLDict
 
 export @service
-export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_user_agent
+export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_global_aws_config, set_user_agent
 export JSONService, RestJSONService, RestXMLService, QueryService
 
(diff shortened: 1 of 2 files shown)


mattBrzezinski:
Any thoughts on this @tclements?

tclements:
I really like the `AWS.set_global_aws_config(AWSConfig(; region="us-west-1")` syntax. Much more succinct than needing to specify a `region` for each call.

src/AWS.jl
```diff
@@ -33,19 +33,39 @@ aws_config = Ref{AWSConfig}()
...
+
+# Keywords
+- `kwargs...`: AWSConfig kwargs to be passed along if the global configuration is not already set
 
 # Returns
 - `AWSConfig`: The global AWS configuration
 """
-function global_aws_config()
+function global_aws_config(; kwargs...)
     if !isassigned(aws_config)
```
omus, line 45: Should probably always create a new `AWSConfig` if `kwargs` is passed in:
```suggestion
    if !isassigned(aws_config) || !isempty(kwargs)
```

src/AWS.jl
```diff
@@ -33,19 +33,39 @@ aws_config = Ref{AWSConfig}()
...
+
+Set the global AWSConfig.
+
+# Arguments
+- `config::AWSConfig`: The AWSConfig to set in the global state
+
+# Returns
+- `AWSConfig`: Global AWSConfig
+"""
+function set_global_aws_config(config::AWSConfig)
```
omus, line 64: Doesn't need to be a different function name. I you really want a different function name I would suggest `global_aws_config!`:
```suggestion
function global_aws_config(config::AWSConfig)
```

test/AWS.jl
```diff
@@ -7,6 +7,31 @@ end
...
+        AWS.aws_config[] = AWSConfig()
+    end
+end
+
+@testset "set global aws config" begin
+    test_region = "test region"
+    expected = AWSConfig(region=test_region)
+
+    try
+        AWS.set_global_aws_config(expected)
```
mattBrzezinski, line 26:
```suggestion
        AWS.global_aws_config(expected)
```

test/AWS.jl
```diff
@@ -7,6 +7,31 @@ end
...
+        result = AWS.global_aws_config()
+
+        @test result.region == test_region
+    finally
+        AWS.set_global_aws_config(AWSConfig())
```
mattBrzezinski, line 31:
```suggestion
        AWS.global_aws_config(AWSConfig())
```

mattBrzezinski (commented):
Comment code changes to do them all at once on GH.

src/AWS.jl
```diff
@@ -33,19 +33,39 @@ aws_config = Ref{AWSConfig}()
...
+    if !isassigned(aws_config) || !isempty(kwargs)
+        aws_config[] = AWSConfig(; kwargs...)
     end
 
     return aws_config[]
 end
 
 
+"""
+    set_global_aws_config(config::AWSConfig)
```
omus, line 54:
```suggestion
    global_aws_config(config::AWSConfig)
```

src/AWS.jl
```diff
@@ -13,7 +13,7 @@ using UUIDs: UUIDs
 using XMLDict
 
 export @service
-export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_user_agent
+export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_global_aws_config, set_user_agent
```
omus, line 16:
```suggestion
export _merge, AWSConfig, AWSExceptions, AWSServices, Request, global_aws_config, set_user_agent
```

omus (approved):
Looks good. My only concern is if calling `global_aws_config()` is any slower now that we do the kwargs check. If there is significant slowdown maybe we should just have `global_aws_config()` and `global_aws_config(::AWSConfig)`

mattBrzezinski:
> Looks good. My only concern is if calling `global_aws_config()` is any slower now that we do the kwargs check. If there is significant slowdown maybe we should just have `global_aws_config()` and `global_aws_config(::AWSConfig)`

There is no slow down, both `master` and this PR run in `0.000002 seconds` after warming up the function. (Idk what to call that, I just ran it a few times then `@time`'d it).

Issue #207, opened 2020-10-02
API auto-generation attempting to remove service


mattBrzezinski:
First occurrence of this happening in #204 where workspaces is attempted to be removed.

mattBrzezinski:
Looks like I had missed something here, because there is no low-level definition for `XRay`, although the service still exists.

- [metadata.json XRay](https://github.com/JuliaCloud/AWS.jl/blob/9b2935ff467d63ec58ace41864f9d638a6bbb969/deps/metadata.json#L2)
- [AWSServices.jl missing XRay](https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWSServices.jl#L240)

mattBrzezinski:
Nothing is actually broken with our package. The issue comes from GitHub itself, it only serves the first 1000 files in the directory.

> https://github.com/aws/aws-sdk-js/tree/master/apis

> Sorry, we had to truncate this directory to 1,000 files. 8 entries were omitted from the list.

PR #208, opened 2020-10-02
Use GitHub.jl to get API definitions

mattBrzezinski:
Resolves: #207 

## Problem
We are currently using the GitHub [Get Repository Contents](https://developer.github.com/v3/repos/contents/#get-repository-content) API to get definitions from the [aws/aws-sdk-js/apis](https://github.com/aws/aws-sdk-js/tree/master/apis) directory.

This API request is limited in returning the first `1,000` objects in a directory. AWS has now pushed to `1,008` files in the `/apis/` directory. This results in us now not being able to generate definitions for certain services. In merging some of the auto-generation I only realized now I've removed the low-level wrapper for [XRay](https://github.com/JuliaCloud/AWS.jl/commit/49ac50843dfc8a1e5add23a9fa59ad6e3d6a5dd4#diff-b03ddd764d1b2745ba6d15cfb9c3d844L238) two days ago breaking the calls for that service.

## Solution
GitHub recommends using [Get A Tree](https://developer.github.com/v3/git/trees/#get-a-tree) API instead for retrieving `>1,000` files. According to [this](https://github.community/t/github-get-tree-api-limits-and-recursivity/1300/2) post in the GitHub community this API has a limit of `100,000` objects before the requests are truncated.

I don't expect the `aws-sdk-js` to reach that number any time soon (currently 1,008 files). Instead of handling the case of `truncated=True`, I decided to just throw an error. This way the GH Action for auto-generation will fail and we can handle it if we ever get there.

**Update:** Using the [GitHub.jl](https://github.com/JuliaWeb/GitHub.jl) package to achieve this, since it also allows us to use authenticated requests. The auto-generation is done via GitHub Actions, which provide us with an authentication token! (Handy!)

## Notes
I am currently rate-limited by the GH API for making requests, I've decided to post this merge request up now. I'm fairly confident that my changes are working but I will confirm this before merging.


src/AWSMetadataUtilities.jl
```diff
@@ -16,13 +17,42 @@ Get a list of all AWS service API definition files from the `awsk-sdk-js` GitHub
 - `Array{OrderedDict}`: Array of Dictionaries which contains information about each AWS Service
 """
 function _get_aws_sdk_js_files()
-    headers = ["User-Agent" => "JuliaCloud/AWS.jl"]
-    url = "https://api.github.com/repos/aws/aws-sdk-js/contents/apis"
+    aws_sdk_js_url = "https://api.github.com/repos/aws/aws-sdk-js"
+
+    function _request(url::String)
+        response = HTTP.get(url, headers)
+        return JSON.parse(String(response.body), dicttype=OrderedDict)
```
mattBrzezinski, line 24:
```suggestion

        return JSON.parse(String(response.body), dicttype=OrderedDict)
```

src/AWSMetadataUtilities.jl
```diff
@@ -16,13 +17,43 @@ Get a list of all AWS service API definition files from the `awsk-sdk-js` GitHub
...
+        resp = _request(url_tree)
+
+        return [t for t in resp["tree"] if t["path"] == "apis"][1]["url"]
+    end
+
+    function _get_api_files(url_apis::String)
+        resp = _request(url_apis)
+
+        if resp["truncated"]
+            throw(FileTruncated("There are more API files that GitHub can send."))
```
iamed2, line 46: Maybe this? I think it should be clearer than what's written but I'm not attached to this suggestion
```suggestion
            throw(FileTruncated("GitHub API file limit exceeded"))
```

src/AWSMetadataUtilities.jl
```diff
@@ -16,13 +17,43 @@ Get a list of all AWS service API definition files from the `awsk-sdk-js` GitHub
...
+        end
+
+        files = resp["tree"]
+        filter!(f -> endswith(f["path"], ".normal.json"), files)
+    end
+
+    headers = ["User-Agent" => "JuliaCloud/AWS.jl"]
```
iamed2, line 53: This line should be back where it came from, since it's referenced early on

src/AWSMetadataUtilities.jl
```diff
@@ -95,16 +126,19 @@ Get the low-level definitions for all AWS Services.
 # Returns
 - `Vector{String}`: Array of low-level service code to be written into `AWSServices.jl`
 """
-function _generate_low_level_definitions(services::AbstractArray{OrderedDict})
+function _generate_low_level_definitions(services::AbstractArray)
```
iamed2, line 129:
```suggestion
function _generate_low_level_definitions(services::AbstractArray{<:AbstractDict})
```
Since we use the dict interface below we can require this.

iamed2 (changes requested):

mattBrzezinski:
I need to think about this some more. This solution is working but has two problems:

1) It's so slow needing multiple HTTP calls to get information, it went from around 1 minute to like 5 minutes and it doesn't even complete cause it's rate limited.
2) The rate limit is 60 requests, it burns through them, and if AWS ever updates a bunch of services at the same time it'll eat them all at once.

iamed2:
What about:
1. Shallow-cloning the repo using git
2. Getting the source code from latest release (e.g. https://github.com/aws/aws-sdk-js/releases/tag/v2.766.0)

mattBrzezinski:
> What about:
> 
> 1. Shallow-cloning the repo using git
>    or
> 2. Getting the source code from latest release (e.g. https://github.com/aws/aws-sdk-js/releases/tag/v2.766.0)

I think we'll have to do something like that instead. I'll work on it this weekend when I have some time.

src/AWSMetadataUtilities.jl
```diff
@@ -85,26 +89,28 @@ function _get_service_and_version(filename::String)
...
     for service in services
-        service_name = service["name"]
+        service_name = service["path"]
         @info "Generating low level wrapper for $service_name"
 
         # Get the contents of the ${Service}.normal.json file
-        request = HTTP.get(service["download_url"])
-        service_metadata = JSON.parse(String(request.body))["metadata"]
+        response = JSON.parse(String(base64decode(blob(repo_name, service["sha"]; auth=auth).content)))
+        service_metadata = response["metadata"]
```
iamed2, line 113:
```suggestion
        service_blob = blob(repo_name, service["sha"]; auth=auth)
        service = JSON.parse(String(base64decode(service_blob.content)))
        service_metadata = service["metadata"]
```

line was too long, changed to match _generate_high_level_wrapper

src/AWSMetadataUtilities.jl
```diff
@@ -252,12 +258,12 @@ Get the required and optional parameters for a given operation.
...
 # Arguments
 - `input::String`: The input object for the given operation
-- `shapes::Dict{String, Any}`: Dictionary of all shapes for this AWS Service
+- `shapes::AbstractDict{String, Any}`: Dictionary of all shapes for this AWS Service
 
 # Returns
 - `Tuple(Dict, Dict)`: (required_parameters, optional_parameters)
 """
-function _get_function_parameters(input::String, shapes::Dict)
+function _get_function_parameters(input::String, shapes::AbstractDict)
```
iamed2, line 266:
```suggestion
function _get_function_parameters(input::String, shapes::AbstractDict{String})
```
doesn't hurt to match the docstring

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "1.14.0"
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+GitHub = "bc5e4493-9b4d-5f90-b8aa-2b2bcaad7a26"
```
iamed2, line 10: Add a version in `[compat]` (at least v5 and then maybe older ones as well).

iamed2 (changes requested):
Maybe add a new description of this PR in a PR comment so people aren't confused if they read this PR in the future?

mattBrzezinski:
> Maybe add a new description of this PR in a PR comment so people aren't confused if they read this PR in the future?

Added an updated section on using GitHub.jl, and that the auth token is provided in the GH Action which is what we use for auto-generation.

iamed2 (approved):

Issue #216, opened 2020-10-13
RestXML parse Content as String


mattBrzezinski:
A string can be represented as an array of uint8, currently this does not work since we do not attempt to parse it as such.

https://github.com/JuliaCloud/AWS.jl/blob/bf563a529c7eab755859eda36aa6a6614e0cf268/src/AWS.jl#L519

PR #217, opened 2020-10-13
Allow Array{UInt8} to be passed in for content in RestXMLService

mattBrzezinski:
Resolves: #216

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -155,7 +155,7 @@ Base.@kwdef mutable struct Request
     request_method::String
 
     headers::AbstractDict{String, String}=LittleDict{String, String}()
-    content::String=""
+    content::Union{String, Array{UInt8}}=""
     resource::String=""
     url::String=""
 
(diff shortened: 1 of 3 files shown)


mjram0s:
Is there an issue attached to this?

mjram0s (approved):

mattBrzezinski:
> Is there an issue attached to this?

Forgot to link to it. Updated the top post with a link for it.

mattBrzezinski:
The alternative solution to this is;

```julia
content=String(args, "body", "")
```

I think this is a more clean solution.

mjram0s:
> The alternative solution to this is;
> 
> ```julia
> content=String(args, "body", "")
> ```
> 
> I think this is a more clean solution.

I agree

src/AWS.jl
```diff
@@ -155,7 +155,7 @@ Base.@kwdef mutable struct Request
     request_method::String
 
     headers::AbstractDict{String, String}=LittleDict{String, String}()
-    content::String=""
+    content::Union{String, Array{UInt8}}=""
```
iamed2, line 158:
```suggestion
    content::Union{String, Vector{UInt8}}=""
```

I don't think it's necessary to support nd arrays here, and keeping the type concrete will help the compiler.

iamed2 (approved):

Issue #223, opened 2020-10-17
S3: Spaces in paths cause HttpVersionNotSupported


christopher-dG:
Doing operations on S3 paths with spaces in them causes an HttpVersionNotSupported error (or at least, the `get_object` and `head_object` that I've tried). Not sure if this is a bug in the XML package or this one, but I figured I'd report it where I found it.

```jl
julia> using AWS: @service

julia> @service S3
 Main.S3

shell>  aws s3 mb s3://cdg-scratch
make_bucket: cdg-scratch

julia> write("x.txt", "hi")
 2

shell> aws s3 cp x.txt s3://cdg-scratch/x.txt
upload: ./x.txt to s3://cdg-scratch/x.txt

# File exists, works fine
julia> S3.get_object("cdg-scratch", "x.txt")
 "hi"

# File does not exist, error is correct
julia> S3.get_object("cdg-scratch", "noexist.txt")
ERROR: AWS.AWSExceptions.AWSException("NoSuchKey", "The specified key does not exist.", EzXML.Document(EzXML.Node(<DOCUMENT_NODE@0x0000000003bdd430>)), HTTP.ExceptionRequest.StatusError(404, "GET", "/cdg-scratch/noexist.txt", HTTP.Messages.Response:
"""
HTTP/1.1 404 Not Found
x-amz-request-id: D2657756A30DAA45
x-amz-id-2: n9lRYt4gZmIvpq1b0s3Fun9dmw7s7vjgsFqXMrIrNpTTm74I6Kdmi+YHA2kafp1TfTiVG/fKId4=
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Sat, 17 Oct 2020 04:21:35 GMT
Server: AmazonS3

<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Key>noexist.txt</Key><RequestId>D2657756A30DAA45</RequestId><HostId>n9lRYt4gZmIvpq1b0s3Fun9dmw7s7vjgsFqXMrIrNpTTm74I6Kdmi+YHA2kafp1TfTiVG/fKId4=</HostId></Error>"""))
Stacktrace:
[...]

PR #226, opened 2020-10-19
Handle spacing in S3 files

mattBrzezinski:
## Problem
This is also a problem in [AWSCore.jl](https://github.com/JuliaCloud/AWSCore.jl), I'll port a fix for that some time soon. But currently you cannot get an object with a space in its name.

The issue roots itself in the URL not being encoded.

```julia
using AWS
@service S3

S3.get_object("/mattbr-test-bucket", "hello world")

> ERROR: AWS.AWSExceptions.AWSException("HttpVersionNotSupported", "The HTTP version specified is not supported."
```

## Solution
Resolves: #223 

We cannot do an `HTTP.escapeuri()` call because we do not want to escape `/` characters. This is a bit problematic since in the high-level wrapper we have,

https://github.com/JuliaCloud/AWS.jl/blob/6dcae89cc64f93d8135c7c50a98771bd7b70d535/src/services/s3.jl#L760-L761

Which passes in `/$(Bucket)/$(Key)` as the URI. Instead what we can do is replace the space character with `%20` to encode it.

### DIFF
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -7,9 +7,9 @@ S3.create_bucket(bucket_name)
     # https://github.com/JuliaCloud/AWS.jl/issues/223
     body = "Hello World!"
     file_name = "contains spaces"
-    S3.put_object(bucket_name, file_name, Dict("body" => body))
 
     try
+        S3.put_object(bucket_name, file_name, Dict("body" => body))
         resp = S3.get_object(bucket_name, file_name)
 
         @test String(resp) == body
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -1,0 +1,15 @@
+@service S3
+
+bucket_name = "aws-jl-test-issues---" * _now_formatted()
+S3.create_bucket(bucket_name)
+
+@testset "issue 223" begin
+    # https://github.com/JuliaCloud/AWS.jl/issues/223
+    body = "Hello World!"
+    file_name = "contains spaces"
+    S3.put_object(bucket_name, file_name, Dict("body" => body))
+
+    resp = S3.get_object(bucket_name, file_name)
+
+    @test String(resp) == body
+end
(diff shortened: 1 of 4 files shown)


test/issues.jl
```diff
@@ -0,0 +1,15 @@
...
+@testset "issue 223" begin
+    # https://github.com/JuliaCloud/AWS.jl/issues/223
+    body = "Hello World!"
+    file_name = "contains spaces"
+    S3.put_object(bucket_name, file_name, Dict("body" => body))
+
+    resp = S3.get_object(bucket_name, file_name)
+
+    @test String(resp) == body
+end
```
mattBrzezinski, line 15:
```suggestion
end

```

test/issues.jl
```diff
@@ -0,0 +1,15 @@
+@service S3
+
+bucket_name = "aws-jl-test-issues---" * _now_formatted()
+S3.create_bucket(bucket_name)
```
iamed2, line 4: This bucket needs to be deleted

mattBrzezinski: Whoopsies, good catch!

mjram0s (approved):

test/issues.jl
```diff
@@ -0,0 +1,19 @@
...
+
+@testset "issue 223" begin
+    # https://github.com/JuliaCloud/AWS.jl/issues/223
+    body = "Hello World!"
+    file_name = "contains spaces"
+    S3.put_object(bucket_name, file_name, Dict("body" => body))
```
iamed2, line 10: You'll need to put this in the `try` as well

mattBrzezinski: I can move it in, but I don't necessarily think that needs to be in the `try` since we're not testing this functionality but rather retrieving the object.

iamed2: But `put_object` could break for unrelated reasons, and then we would have random buckets lying around.

mattBrzezinski: Fair point.

iamed2 (approved):

Issue #168, opened 2020-09-02
Issues with docstrings for high-level services


omus:
- [ ] Docstring function name doesn't match actual function name ([`CreateConnectorProfile` vs. `create_connector_profile`](https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/appflow.jl#L8))
- [ ] Docstrings aren't actually [attached to any function/method](https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/appflow.jl#L21). I'd suggest attaching it to the function in general (e.g. `function create_connector_profile end`)
- [ ] Docstring function name cannot be [called without arguments](https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/appflow.jl#L8). Just remove the `()`
- [ ] Parameters should probably also be in [snake_case instead of CamelCase](https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/appflow.jl#L14)?

mattBrzezinski:
I'll tack this in as well, but a line character limit would be nice to impose here. Currently if parameter docstr is say 500 character, they'll all sit on the same line. We should limit this to 140, break on the next word and continue the remainder of the docstring on the next line.

mattBrzezinski:
And since we're touching on docstrs, add [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl) to the package too.

mattBrzezinski:
Handle lists as well.

https://github.com/JuliaCloud/AWS.jl/pull/171#discussion_r482984692

PR #229, opened 2020-10-23
Attach docstrings to high level service functions

c42f:
Fixes the second point in #168 which allows users to access the docstrings in the REPL.

I tested this locally in an ad-hoc way by locally modifying deps/metadata.jl and running `julia --project -e 'using AWS; AWS.AWSMetadata.parse_aws_metadata();'`

It seems that the real tests are oddly whitespace insensitive.


mattBrzezinski:
This is lovely, thanks for the merge request! The tests are whitespace insensitive because I found it to be PITA to modify the tests when whitespace changes really didn't matter.

One test is currently failing locally for me, going to quickly fix that merge it in and then these should be good to go!

mattBrzezinski:
Tested the changes locally to see how the docstrs get generated, look good to me! Thanks again :)

c42f:
Awesome, thanks Matt.

Is there a way to force the whole set of services to get updated? In principle the automated update bot should invalidate the cache based not just on the hashes of the service json definitions, but on the hash of AWSMetadata.jl and AWSMetadataUtilities.jl as well.

mattBrzezinski:
That's one thing which I should have requested to go along side this PR, but I can make a new one for it.
Usually with breaking changes, I would regenerate all service definitions by wiping the `deps/metadata.json` file to just `{}`. This would force everything to be re-created.

We could modify the bot to just ignore the contents of the `deps/metadata.json` file, and just re-generate all definitions. This would definitely increase the runtime but when the merge request is made you'd always have the proper version. I guess runtime complexity doesn't matter much in this case.

I've logged an issue for this change https://github.com/JuliaCloud/AWS.jl/issues/234, thanks for bringing it up. It's definitely a change which makes sense to have!

PR #235, opened 2020-10-26
make URL's a little more hackable

ExpandingMan:
This adds the `host` field to `AWSConfig` which allows users to override the generated host URL for all services.  This is crucial for software that provides API's that mimic AWS services such as min.io.

I used the extremely awkward name `endpoint` as the keyword argument.  This is to maintain consistency with the AWS CLI which allows overriding this argument with `--endpoint-url`.  Of course this name makes no sense because it only gives the host typically, I'd be happy to change it but I'm not sure what the best convention is here.  It's already a bit obscure.

I still have to add some unit tests, in the meantime, let me know what you think.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -450,12 +450,22 @@ function _generate_rest_resource(request_uri::String, args::AbstractDict{String,
     return request_uri
 end
 
-function _generate_service_url(region::String, service::String, resource::String)
+"""
+    service_url(aws::AWSConfig, service="")
+
+Generate the URL for the service.  If the `host` field of the `AWSConfig` is set, will return this
+value unmodified.
+"""
+function service_url(aws::AWSConfig, service::String="")
+    isempty(aws.host) || return aws.host
+
     SERVICE_HOST = "amazonaws.com"
     regionless_services = ("iam", "route53")
 
-    if service in regionless_services || (service == "sdb" && region == "us-east-1")
-        region = ""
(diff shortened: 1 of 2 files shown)


meggart:
I think this is related to #212 which would solve your use case as well

mattBrzezinski:
Sorry I'm quite busy with other work to really review all these PRs. Around the end of the month I'll have more time for GitHub work.

As @meggart mentioned #212 will make things more abstract and allow for more composability. I think that's more favourable as a long term solution. Unless this is immediately a blocking issue, then I have no problems pushing this through to help ease your workflow!

ExpandingMan:
Thanks.  My only concern with #212 is that it is perhaps a bit elaborate for wanting to change something as simple as the URL string.  (On the other hand, this probably isn't a particularly common use case.)  One option would be to just add some convenience functions for the implementation in #212 if it's not done already.

I'd be happy to make another PR to make this intended use case "easy" (not that it's particularly difficult in @meggart 's example) after #212 is merged.

In the mean time, it'd be nice if you wanted to merge and tag this PR in the meantime (i.e. until it is superceded by #212), but it's not urgent for me (I probably need to stick to AWSCore.jl for a while anyway in my code), so no pressure.  If you'd prefer to close this, that'd be fine with me.

mattBrzezinski:
https://github.com/JuliaCloud/AWS.jl/pull/274 went in, I believe this resolves this?

ExpandingMan:
Yup, thanks.

PR #257, opened 2020-12-02
Use WebIdentityTokens to create credentials

### LINKED ISSUE #244: AWSCredentials support for AssumeRoleWithWebIdentity mechanism
I'm trying to use AWS.jl from within a K8s pod that is running with a [IAM-role-linked ServiceAccount](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (see also: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/). Via this mechanism, it's super easy to provision credentials to an individual pod so that commands in the pod are automatically run via the specified role (basically similar to EC2 instance-attached roles, but at a per-pod level). A bunch of the other SDKs already support this out-of-the-box without any required configuration.

Is there a way to for us to use AWS.jl with this mechanism? Naively, it seems like commands fail with permissions errors; inspecting more deeply, it looks like constructing a new `AWSCredentials` instance skips past the pod IAM role settings and just grabs the underlying node's role (i.e. the underlying EC2 instance's attached role). Reading the [AWSCredentials code](https://github.com/JuliaCloud/AWS.jl/blob/7d26e4fa3a8e0a57f3535ae7d06cc635aeabf20c/src/AWSCredentials.jl), it looks like it doesn't automatically read either the `AWS_WEB_IDENTITY_TOKEN_FILE` environmental variable or the `web_identity_token_file` setting from `~/.aws/config` when searching for credentials. 

If AWS.jl adds support for this (and/or if there's a nice workaround we can use), Beacon Biosignals can start using AWS.jl, which would be a big win for us 🙂 We're currently locked to using PyCall w/ `boto3` because of this mechanism, which can have annoying interactions with async Julia code.

additional references:

- https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-web_identity_token_file.html
- https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc
- https://github.com/aws/aws-sdk-go/issues/3101 (a similiar issue for the AWS Go SDK)

mattBrzezinski:
### Resolves #244 

These changes use existing credentials to assume temporary credentials via a [Web Identity File](https://docs.aws.amazon.com/credref/latest/refdocs/setting-global-web_identity_token_file.html) using [AssumeRoleWithWebIdentity](https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/sts.jl#L48).

The code was taken from the issue itself, with a slight modification to ensure that the appropriate env vars are set before proceeding.

### DIFF
--- a/test/AWSCredentials.jl
+++ b/test/AWSCredentials.jl
@@ -462,17 +462,35 @@ end
         end
     end
 
+    @testset "Web Identity File" begin
+        mktempdir() do dir
+            web_identity_file = joinpath(dir, "web_identity")
+            write(web_identity_file, "foobar")
+
+            withenv(
+                "AWS_ROLE_ARN" => "foobar",
+                "AWS_ROLE_SESSION_NAME" => Patches.web_sesh_token,
+                "AWS_WEB_IDENTITY_TOKEN_FILE" => web_identity_file,
+            ) do
+                apply(Patches._web_identity_patch) do
+                    result = credentials_from_webtoken()
+
+                    @test result.access_key_id == Patches.web_access_key
+                    @test result.secret_key == Patches.web_secret_key
+                    @test result.token == Patches.web_sesh_token
+                    @test result.renew == credentials_from_webtoken
+                end
+            end
+        end
+    end
+
+    @testset "Web Identity File -- Exception" begin
+        @test_throws WebIdentityVarsNotSet credentials_from_webtoken()
+    end
+
     @testset "Credentials Not Found" begin
         _http_request_patch = @patch function HTTP.request(method::String, url::String)
             return nothing
-        end
-
-        _cred_file_patch = @patch function dot_aws_credentials_file()
-            return ""
-        end
-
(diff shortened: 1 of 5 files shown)


fchorney (approved):
Code LGTM

test/AWSCredentials.jl
```diff
@@ -462,17 +462,35 @@ end
...
+                "AWS_ROLE_SESSION_NAME" => Patches.web_sesh_token,
+                "AWS_WEB_IDENTITY_TOKEN_FILE" => web_identity_file,
+            ) do
+                apply(Patches._web_identity_patch) do
+                    result = credentials_from_webtoken()
+
+                    @test result.access_key_id == Patches.web_access_key
+                    @test result.secret_key == Patches.web_secret_key
+                    @test result.token == Patches.web_sesh_token
+                    @test result.renew == credentials_from_webtoken
```
SimonDanisch, line 481: Maybe add the test for refresh as well?

PR #260, opened 2020-12-04
Add webtoken auth to precedence list

sean-bennett112:
This should fix the issues we're still having from #244.

Also:

- removes `profile` from the webtoken parameters since it's not really used. I can add this back in if necessary, I was just following the convention from the EC2 and ECS credential functions.
- returns `nothing` if the keys don't exist, since we want it to continue down the precedence list.
- explicitly passes `nothing` in for `creds`. This caused an infinite loop as `credentials_from_webtoken` called `AWSConfig` which would call `AWSCredentials`, etc.

There's probably a cleaner way to handle that last point, if anyone has opinions!

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -152,6 +152,7 @@ function AWSCredentials(; profile=nothing)
         env_var_credentials,
         () -> dot_aws_credentials(profile),
         () -> dot_aws_config(profile),
+        credentials_from_webtoken,
         ecs_instance_credentials,
         ec2_instance_credentials
     ]
@@ -453,7 +454,7 @@ function dot_aws_config(profile=nothing)
 end
 
 
-function credentials_from_webtoken(profile=nothing)
+function credentials_from_webtoken()
     token_role_arn = "AWS_ROLE_ARN"
     token_role_session = "AWS_ROLE_SESSION_NAME"
     token_web_identity = "AWS_WEB_IDENTITY_TOKEN_FILE"
@@ -464,9 +465,7 @@ function credentials_from_webtoken(profile=nothing)
         haskey(ENV, token_web_identity)
 
     if !has_all_keys
-        throw(WebIdentityVarsNotSet(
-            "You need to set $(token_role_arn), $(token_role_session), $(token_web_identity) environment variables"
-        ))
+        return nothing
     end
 
     role_arn = ENV[token_role_arn]
@@ -480,7 +479,7 @@ function credentials_from_webtoken(profile=nothing)
             "RoleSessionName" => role_session,
             "WebIdentityToken" => web_identity
         );
-        aws_config=AWSConfig(profile=profile)
+        aws_config=AWSConfig(creds=nothing)
     )
 
     role_creds = resp["AssumeRoleWithWebIdentityResult"]["Credentials"]
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -472,7 +472,7 @@ function credentials_from_webtoken()
     role_session = ENV[token_role_session]
     web_identity = read(ENV["AWS_WEB_IDENTITY_TOKEN_FILE"], String)
 
-    resp = AWSServices.sts(
+    resp = @mock AWSServices.sts(
         "AssumeRoleWithWebIdentity",
         Dict(
             "RoleArn" => role_arn, 
--- a/src/AWSExceptions.jl
+++ b/src/AWSExceptions.jl
@@ -5,7 +5,7 @@ using JSON
 using XMLDict
 using XMLDict: XMLDictElement
 
-export AWSException, ProtocolNotDefined, InvalidFileName, NoCredentials, WebIdentityVarsNotSet
+export AWSException, ProtocolNotDefined, InvalidFileName, NoCredentials
 
 struct ProtocolNotDefined <: Exception
     message::String
--- a/src/AWSExceptions.jl
+++ b/src/AWSExceptions.jl
@@ -22,11 +22,6 @@ struct NoCredentials <: Exception
 end
 Base.show(io::IO, e::NoCredentials) = println(io, e.message)
 
-struct WebIdentityVarsNotSet <: Exception
-    message::String
-end
-Base.show(io::IO, e::WebIdentityVarsNotSet) = println(io, e.message)
-
 struct AWSException <: Exception
     code::String
     message::String
--- a/test/AWSCredentials.jl
+++ b/test/AWSCredentials.jl
@@ -463,6 +463,8 @@ end
     end
 
     @testset "Web Identity File" begin
+        @test credentials_from_webtoken() == nothing
+
         mktempdir() do dir
             web_identity_file = joinpath(dir, "web_identity")
             write(web_identity_file, "foobar")
@@ -479,13 +481,18 @@ end
                     @test result.secret_key == Patches.web_secret_key
                     @test result.token == Patches.web_sesh_token
                     @test result.renew == credentials_from_webtoken
+                    expiry = result.expiry
+
+                    result = check_credentials(result)
+
+                    @test result.access_key_id == Patches.web_access_key
+                    @test result.secret_key == Patches.web_secret_key
+                    @test result.token == Patches.web_sesh_token
+                    @test result.renew == credentials_from_webtoken
+                    @test expiry != result.expiry
                 end
             end
         end
-    end
-
-    @testset "Web Identity File -- Exception" begin
-        @test_throws WebIdentityVarsNotSet credentials_from_webtoken()
     end
 
     @testset "Credentials Not Found" begin
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,7 +1,7 @@
 using AWS
 using AWS: AWSCredentials
 using AWS: AWSServices
-using AWS.AWSExceptions: AWSException, NoCredentials, WebIdentityVarsNotSet
+using AWS.AWSExceptions: AWSException, NoCredentials
 using AWS.AWSMetadataUtilities: _clean_documentation, _filter_latest_service_version,
     _generate_low_level_definition, _generate_high_level_definition, _generate_high_level_definitions,
     _get_aws_sdk_js_files, _get_service_and_version, _get_function_parameters, _clean_uri, _format_function_name,


src/AWSCredentials.jl
```diff
@@ -464,9 +465,7 @@ function credentials_from_webtoken(profile=nothing)
         haskey(ENV, token_web_identity)
 
     if !has_all_keys
-        throw(WebIdentityVarsNotSet(
```
mattBrzezinski, line 467: Can you remove this from the `src/AWSExceptions.jl` file?

src/AWSCredentials.jl
```diff
@@ -480,7 +479,7 @@ function credentials_from_webtoken(profile=nothing)
             "RoleSessionName" => role_session,
             "WebIdentityToken" => web_identity
         );
-        aws_config=AWSConfig(profile=profile)
+        aws_config=AWSConfig(creds=nothing)
```
mattBrzezinski, line 482: I guess there isn't really a simple way of testing this out is there... (all this functionality that is).

mattBrzezinski (commented):
Since I can't think of a simple way of testing this out, without creating some identity provider and jumping through hoops.

Are you able to confirm that this testing works with this branch and changes in it?

src/AWSCredentials.jl
sean-bennett112: Done!

src/AWSCredentials.jl
sean-bennett112: Yea, unfortunately it's a little involved. What we ended up doing is:

Modifying the STS call to include `DurationSeconds` of `900`(shortest the docs say it accepts).

Grabbing the auth token from one of our EKS clusters via:

```julia
token = chomp(read(`kubectl exec -t $(pod) -it -- julia -e 'write(stdout, read(ENV["AWS_WEB_IDENTITY_TOKEN_FILE"]))'`, String))
```

Then running auth in a loop checking that the credentials calls work, and correctly renews after expiration:

```julia
creds = AWS.global_aws_config()
expires_in = round(creds.credentials.expiry - now(UTC), Minute)
@show expires_in
@assert expires_in > Minute(10)

# Check that force renewal works
AWS.check_credentials(creds.credentials; force_refresh=true)

# Sleep until after expiration to confirm renewal works
for i in 1:2
    # Wait for 9 minutes so we skip past the <5 minutes renewal period
    sleep(540)
    AWS.check_credentials(creds.credentials)
    @info "creds: $(creds)"
    local expires_in = round(creds.credentials.expiry - now(UTC), Minute)
    @show expires_in
end
```

If you have a good idea on how to automate that testing, I'd be happy to hear it! :)

Also if there's any other manual testing you'd like me to do before you're willing to merge it, happy to take those steps.

mattBrzezinski: Yeah my main point is the `WebIdentityToken` itself, and getting a real response back from STS.

Just to confirm that this is in fact using the `credentials_from_webtoken` function;

```julia
using AWS
config = AWS.global_aws_config()

config.credentials.renew == AWS.credentials_from_webtoken
```

sean-bennett112: Oh I see. Yes, I was able to confirm that, and I just double-checked it just in case. We also checked an S3 bucket that I wouldn't have access to without switching to that role.

mattBrzezinski (approved):

mattBrzezinski:
Ah just missed removing `WebIdentityVarsNotSet` in:
- `/test/runtests.jl`

Then this test needs to be modified:
- https://github.com/JuliaCloud/AWS.jl/blob/1e012b4024ffe215c9b17b113df7944ef12ad1eb/test/AWSCredentials.jl#L487

And may as well implement the suggestion from:
- https://github.com/JuliaCloud/AWS.jl/pull/257#discussion_r535986554

sean-bennett112:
Ah yea, missed the other occurrences of the exception. I removed those, and made sure the STS call is now mockable. Also added the refresh test in.

I had trouble running all the unit tests locally, but I ran what I could. If there's a good way of running all the unit tests locally, I'd be happy to test further.

mattBrzezinski:
> I had trouble running all the unit tests locally, but I ran what I could. If there's a good way of running all the unit tests locally, I'd be happy to test further.

The tests should work locally, what errors were you encountering?

sean-bennett112:
Generally everything would hang once I hit `ec2_instance_credentials`, specifically [here](https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWSCredentials.jl#L299). For example, [here](https://github.com/JuliaCloud/AWS.jl/blob/master/test/runtests.jl#L29), `AWSConfig` calls `AWSCredentials`, which loops through till it checks for ec2 credentials, and hangs.

I could get around that by running `julia --profile=. test/runtests.jl` with my AWS environment variables explicitly passed in, but then I get a bunch of test failures with `Access Denied` and `403 Forbidden`. Here's the full test run in that case:

```
Test Summary:                      | Pass  Error  Total
AWS.jl                             |  254     11    265
  service module                   |    1             1
  global config, kwargs            |    1             1
  set global aws config            |    1             1
  set user agent                   |    2             2
  sign                             |    9             9
  submit_request                   |   33            33
  _generate_rest_resource          |    1             1
  _generate_service_url            |    4             4
  _return_headers                  |    3             3
  _flatten_query                   |    2             2
  json                             |           2      2
    high-level secrets manager     |           1      1
    low-level secrets manager      |           1      1
  query                            |           4      4
    high-level iam                 |           1      1
    low-level iam                  |           1      1
    high-level sqs                 |           1      1
    low-level sqs                  |           1      1
  rest-xml                         |    2      2      4
    high-level s3                  |    1      1      2
    low-level s3                   |    1      1      2
  rest-json                        |           2      2
    high-level glacier             |           1      1
    low-level glacier              |           1      1
  Load Credentials                 |    5             5
  AWSCredentials                   |   67            67
  Retrieving AWS Credentials       |   37            37
  AWSException                     |   40            40
  _get_aws_sdk_js_files            |    2             2
  _filter_latest_service_version   |    5             5
  _get_service_and_version         |    4             4
  _generate_low_level_definitions  |    5             5
  _generate_low_level_definition   |    6             6
  _clean_documentation             |    1             1
  _clean_uri                       |    5             5
  _format_function_name            |    4             4
  _get_function_parameters         |    8             8
  _generate_high_level_definitions |    2             2
  _generate_high_level_definition  |    4             4
```

mattBrzezinski:
> I could get around that by running `julia --profile=. test/runtests.jl` with my AWS environment variables explicitly passed in, but then I get a bunch of test failures with `Access Denied` and `403 Forbidden`. Here's the full test run in that case:

The way I run these tests is:

```julia
julia --project
] test
```

Do you have full Admin permissions? At the very least you'll need all the permissions in [here](https://github.com/JuliaCloud/AWS.jl/blob/master/test/resources/aws_jl_test.yaml#L11)

sean-bennett112:
Ok, switching from my (limited) work account to my personal account did the trick. Strangely, if I run:

```
julia --project
] test
```

Then all the tests pass. However, if I instead do `julia --project=. runtests.jl`, then I still get a single failure:

```
WARNING: replacing module S3.
Path `/home/sean/projects/beacon-biosignals/.../dev/AWS/test/resources/TestPkg` exists and looks like the correct package. Using existing path.
  Resolving package versions...
AWS.jl: Error During Test at /home/sean/projects/beacon-biosignals/.../dev/AWS/test/runtests.jl:35
  Got exception outside of a @test
  LoadError: KeyError: key "AWS" not found
  Stacktrace:
   [1] _all(::Pkg.Types.var"#37#40"{Dict{String,Bool}}, ::Dict{String,UUID}, ::Colon) at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:?
   [2] destructure(::Dict{UUID,Pkg.Types.PackageEntry}) at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:?
   [3] write_env(::Pkg.Types.EnvCache; update_undo::Bool) at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:?
   [4] write_env at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/Types.jl:1366 [inlined]
   [5] develop(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}, ::Array{UUID,1}; preserve::Pkg.Types.PreserveLevel, platform::Pkg.BinaryPlatforms.Linux) at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/Operations.jl:1154
   [6] develop(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; shared::Bool, preserve::Pkg.Types.PreserveLevel, platform::Pkg.BinaryPlatforms.Linux, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:134
   [7] develop(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:93
   [8] #develop#13 at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:67 [inlined]
   [9] develop at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:67 [inlined]
   [10] #develop#11 at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:65 [inlined]
   [11] develop at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:65 [inlined]
   [12] develop(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::String, mode::Pkg.Types.PackageMode, subdir::Nothing, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /build/source/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:82
   [13] top-level scope at /home/sean/projects/beacon-biosignals/.../dev/AWS/test/test_pkg.jl:4
   [14] include(::String) at ./client.jl:457
   [15] top-level scope at /home/sean/projects/beacon-biosignals/.../dev/AWS/test/runtests.jl:41
   [16] top-level scope at /build/source/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
   [17] top-level scope at /home/sean/projects/beacon-biosignals/.../dev/AWS/test/runtests.jl:36
   [18] include(::Function, ::Module, ::String) at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:? (repeats 2 times)
   [19] exec_options(::Base.JLOptions) at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:?
   [20] _start() at /nix/store/abzhybmpf41fbd32l7wsla9c87hhhlz9-julia-1.5.0/lib/julia/sys.so:?
  in expression starting at /home/sean/projects/beacon-biosignals/.../dev/AWS/test/test_pkg.jl:3
  
Test Summary:                      | Pass  Error  Total
AWS.jl                             |  296      1    297
  service module                   |    1             1
  global config, kwargs            |    1             1
  set global aws config            |    1             1
  set user agent                   |    2             2
  sign                             |    9             9
  submit_request                   |   33            33
  _generate_rest_resource          |    1             1
  _generate_service_url            |    4             4
  _return_headers                  |    3             3
  _flatten_query                   |    2             2
  json                             |    4             4
  query                            |   12            12
  rest-xml                         |   14            14
  rest-json                        |   10            10
  Load Credentials                 |    5             5
  AWSCredentials                   |   67            67
  Retrieving AWS Credentials       |   37            37
  AWSException                     |   40            40
  _get_aws_sdk_js_files            |    2             2
  _filter_latest_service_version   |    5             5
  _get_service_and_version         |    4             4
  _generate_low_level_definitions  |    5             5
  _generate_low_level_definition   |    6             6
  _clean_documentation             |    1             1
  _clean_uri                       |    5             5
  _format_function_name            |    4             4
  _get_function_parameters         |    8             8
  _generate_high_level_definitions |    2             2
  _generate_high_level_definition  |    4             4
  issue 223                        |    1             1
  issue 227                        |    3             3
```

I'm assuming this is due to some differences in how those two commands behave, where I thought they'd be the same.

PR #263, opened 2020-12-09
updated HTTP compat

ExpandingMan:
This package was causing HTTP to be held back.

HTTP 0.9 seems unlikely to break anything, but I don't really have a way of testing this.


mattBrzezinski:
Ran the tests using `HTTP@0.9` and it's fine!

mattBrzezinski (approved):

ExpandingMan:
great, thanks!  A tag would be appreciated when you get a chance

PR #265, opened 2020-12-15
make AWS_ROLE_SESSION_NAME optional

kolia:
AWS docs indicate that it should be optional.
If unspecified, a default is set and warning logged.

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -478,7 +478,8 @@ function credentials_from_webtoken()
 
     role_arn = ENV[token_role_arn]
     role_session = get(ENV, token_role_session) do
-        default_role_session = replace("default-AWS.jl-role-session-$(now())", ":" => "-")
+        timestamp = Dates.format(now(UTC), dateformat"yyyymmdd\THHMMSS\Z")
+        default_role_session = "default-AWS.jl-role-session-$timestamp" 
         @warn "Consider setting (optional) AWS_ROLE_SESSION_NAME environment variable, defaulting to $default_role_session"
         return default_role_session
     end
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -477,7 +477,7 @@ function credentials_from_webtoken()
     end
 
     role_arn = ENV[token_role_arn]
-    role_session = get!(ENV, token_role_session) do
+    role_session = get(ENV, token_role_session) do
         default_role_session = replace("default-AWS.jl-role-session-$(now())", ":" => "-")
         @warn "Consider setting (optional) AWS_ROLE_SESSION_NAME environment variable, defaulting to $default_role_session"
         return default_role_session
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -470,7 +470,6 @@ function credentials_from_webtoken()
 
     has_all_keys =
         haskey(ENV, token_role_arn) &&
-        haskey(ENV, token_role_session) &&
         haskey(ENV, token_web_identity)
 
     if !has_all_keys
@@ -478,7 +477,11 @@ function credentials_from_webtoken()
     end
 
     role_arn = ENV[token_role_arn]
-    role_session = ENV[token_role_session]
+    role_session = get!(ENV, token_role_session) do
+        default_role_session = replace("default-AWS.jl-role-session-$(now())", ":" => "-")
+        @warn "Consider setting (optional) AWS_ROLE_SESSION_NAME environment variable, defaulting to $default_role_session"
+        return default_role_session
+    end
     web_identity = read(ENV["AWS_WEB_IDENTITY_TOKEN_FILE"], String)
 
     resp = @mock AWSServices.sts(
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -494,12 +494,13 @@ function credentials_from_webtoken()
     )
 
     role_creds = resp["AssumeRoleWithWebIdentityResult"]["Credentials"]
+    assumed_role_user = resp["AssumeRoleWithWebIdentityResult"]["AssumedRoleUser"]
 
     return AWSCredentials(
         role_creds["AccessKeyId"],
         role_creds["SecretAccessKey"],
         role_creds["SessionToken"],
-        role_creds["AssumedRoleUser"]["Arn"];
+        assumed_role_user["Arn"];
         expiry=DateTime(rstrip(role_creds["Expiration"], 'Z')),
         renew=credentials_from_webtoken
     )
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -463,34 +463,32 @@ function dot_aws_config(profile=nothing)
 end
 
 
+# https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc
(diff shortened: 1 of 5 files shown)


mattBrzezinski (approved):

src/AWSCredentials.jl
```diff
@@ -461,15 +461,18 @@ function credentials_from_webtoken()
...
-        haskey(ENV, token_role_session) &&
         haskey(ENV, token_web_identity)
 
     if !has_all_keys
         return nothing
     end
 
     role_arn = ENV[token_role_arn]
-    role_session = ENV[token_role_session]
+    role_session = get!(ENV, token_role_session) do
```
omus, line 471: Do we need to permanently modify the `ENV`?
```suggestion
    role_session = get(ENV, token_role_session) do
```

kolia: > Do we need to permanently modify the `ENV`?

No need really, but also once this has run you don't have a trace of what the generated session name ended up being outside of logs. What's worst, having no trace or setting an ENV var? Can't think of a third way.

I guess it's best not to pollute ENV.

mattBrzezinski:
Bump.

kolia:
Is the build failure on julia 1.3 a blocker? Seems to have to do with auth expiration times not changing as expected, I'm not understanding what that has to do with this PR or with the julia version.

src/AWSCredentials.jl
```diff
@@ -461,15 +461,18 @@ function credentials_from_webtoken()
...
         haskey(ENV, token_web_identity)
 
     if !has_all_keys
         return nothing
     end
 
     role_arn = ENV[token_role_arn]
-    role_session = ENV[token_role_session]
+    role_session = get(ENV, token_role_session) do
+        default_role_session = replace("default-AWS.jl-role-session-$(now())", ":" => "-")
```
omus, line 472: You can just specify a date format to avoid having to replace characters. I'd recommend sticking to a ISO8601 format so I switched to using a UTC format which doesn't use special characters:
```suggestion
        timestamp = Dates.format(now(UTC), dateformat"yyyymmdd\THHMMSSZ")
        default_role_session = "default-AWS.jl-role-session-$timestamp" 
```
Note: this module will need to have `using Dates` for this to work

omus, line 472: You can just specify a date format to avoid having to replace characters. I'd recommend sticking to a ISO8601 format so I switched to using a UTC format which doesn't use special characters:
```suggestion
        timestamp = Dates.format(now(UTC), dateformat"yyyymmdd\THHMMSSZ")
        default_role_session = "default-AWS.jl-role-session-$timestamp" 
```
Note: this module will need to have `using Dates` for this to work

kolia:
The module is already `using Dates`, I just manually confirmed that role-based auth works with and without setting a session name.

omus:
bors try

omus:
After running into this issue myself I realized that the warning would get old very quickly. In order to improve the situation I added the role name into the default session name. Using this and automatically restricting the role session name to 64 characters (the session name limit) I think the default is definitely good enough that we don't need to promote having users specify their own role names.

omus:
bors try

omus:
bors try

omus:
I've validated this behaviour works correctly in a live AWS session

omus:
@mattBrzezinski do you want to give this another look over?

mattBrzezinski:
> @mattBrzezinski do you want to give this another look over?

Yeah I can take a look at this tonight / tomorrow morning and merge it in afterwards!

src/AWSCredentials.jl
```diff
@@ -463,40 +463,44 @@ function dot_aws_config(profile=nothing)
...
     end
 
     role_arn = ENV[token_role_arn]
-    role_session = ENV[token_role_session]
     web_identity = read(ENV["AWS_WEB_IDENTITY_TOKEN_FILE"], String)
+    role_session = get(ENV, token_role_session) do
+        _role_session_name(
+            "AWS.jl-role-",
+            basename(role_arn),
+            "-" * Dates.format(@mock(now(UTC)), dateformat"yyyymmdd\THHMMSS\Z"),
```
mattBrzezinski, line 482: > Dates.format(@mock(now(UTC)), dateformat"yyyymmdd\THHMMSS\Z")

Maybe we pull this out as an internal function, a similar piece of code already exists in [test/runtests.jl](https://github.com/JuliaCloud/AWS.jl/blob/68d2955446347b70dca7e23d04d56c375bf45993/test/runtests.jl#L32)

test/AWSCredentials.jl
```diff
@@ -468,30 +473,62 @@ end
...
+                    @test result.user_arn == role_arn * "/" * session_name
                     @test result.renew == credentials_from_webtoken
                     @test expiry != result.expiry
                 end
             end
+
+            session_name = "AWS.jl-role-foobar-20210101T000000Z"
+            patches = [
+                patch
+                @patch Dates.now(::Type{UTC}) = DateTime(2021)
```
mattBrzezinski, line 519: Is there a missing comma here? I'm not familiar with the syntax.

omus: This is just creating a vector by using newlines as the delimiter:
```julia
julia> [
          1
          2
       ]
2-element Vector{Int64}:
 1
 2

julia> [
          1
          2
       ]
2-element Vector{Int64}:
 1
 2
```
Normally I would just use commas but the `@patch` macro captures that as part of the expression and I need to use parenthesis to avoid that. This seems like a cleaner way to do it

src/AWSCredentials.jl
omus: I'll look into that. Do you happen to know why that test function uses `lowercase`?

test/AWSCredentials.jl
mattBrzezinski: Ah makes sense! Thanks for the explanation :)

src/AWSCredentials.jl
omus: I'm not sure it's worth it to refactor this into a separate function. The `_now_formatted` function includes milliseconds (which we could use here) and there are several different `DateFormat`s used throughout the repo already:
```
src/AWS.jl:224:    query["Expires"] = Dates.format(time + Dates.Minute(2), dateformat"yyyy-mm-dd\THH:MM:SS\Z")
src/AWS.jl:246:    date = Dates.format(time, dateformat"yyyymmdd")
src/AWS.jl:247:    datetime = Dates.format(time, dateformat"yyyymmdd\THHMMSS\Z")
src/AWSCredentials.jl:482:            "-" * Dates.format(@mock(now(UTC)), dateformat"yyyymmdd\THHMMSS\Z"),
test/AWS.jl:49:    date = Dates.format(time, dateformat"yyyymmdd")
test/AWS.jl:88:        expected_x_amz_date = Dates.format(time, dateformat"yyyymmdd\THHMMSS\Z")
test/runtests.jl:33:    return lowercase(Dates.format(now(Dates.UTC), dateformat"yyyymmdd\THHMMSSsss\Z"))
```
As some of these formats are required to be specific to follow spec and others are just there to make identifiers unique I think there isn't a point in extracting this date formatting code. I could see some further refactoring of the `_role_session_name` to automatically add a timestamp suffix and possibly use it with other AWS services that support names

mattBrzezinski: Makes sense!

PR #285, opened 2021-02-17
Fix special characters in S3 filenames

mattBrzezinski:
# Problem
AWS S3 allows for a lot of characters in file names. These aren't always limited to `a-z A-Z 0-9`, but also some special characters like `-` and `_`. Or more problematic ones, like `*`, `(`, `'`.

I've already handled one special casing previous (a space), however starting to see a few more appearing. This PR will handle a few more, but not all.

# Solution
The fix here is to replace known problematic characters. In [AWSCore](https://github.com/JuliaCloud/AWSCore.jl/blob/1eb00b46ceac09712429ff923bc1a8cfcf21c1e2/src/AWSCore.jl#L333) this was a bit easier, because the file path was provided as an extra argument. However, in this iteration we are just providing the `request_uri`.

So simply doing `HTTP.escapeuri(request_uri)` would work, except that it will replace `/` with `%2F`, and it will completely break S3 usage. There are a few ways to handle these situations outside of this proposed change:

1) Ignore it, users can manually escape URIs themselves (not nice)
2) Call `HTTP.escapeuri(request_uri)` then replace all instances of `%2F` with `/` (seems more dangerous)
3) Completely re-work `AWS.jl` and create a new major version release (I don't have time for this, nor the want)

# Notes
- [Relevant Stackoverflow Post](https://stackoverflow.com/questions/7116450/what-are-valid-s3-key-names-that-can-be-accessed-via-the-s3-rest-api)

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -330,6 +330,38 @@ function _http_request(request::Request)
     end
 end
 
+
+"""
+    _clean_uri(uri::AbstractString)
+
+Escape special AWS S3 characters properly.
+
+AWS S3 allows for various special characters in file names, these characters are not being
+properly escaped before we make the requests.
+
+We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
+are used in the filepathing for sub-directories.
+
+# Arguments
+- `uri::AbstractString`: URI to to cleaned
+
+# Returns
+- `String`: URI with characters escaped
+"""
+function _clean_uri(uri::AbstractString)
+    chars_to_clean = [
+        ' ' => "%20",
+        '!' => "%21",
+        ''' => "%27",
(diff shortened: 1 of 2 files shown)


nicoleepp (approved):

mattBrzezinski:
bors try

nickrobinson251:
bump :) it'd help me to switch for AWSCore.jl -> AWS.jl if there were a release with this fix

mattBrzezinski:
> bump :) it'd help me to switch from AWSCore.jl -> AWS.jl if there were a release with this fix

Was leaving this in case anyone else wanted to comment.
Again, bors is struggling, going to merge and tag a release now.

nickrobinson251:
thanks a lot for the quick fix!

PR #286, opened 2021-02-17
Throw a better error when no creds are found

### LINKED ISSUE #284: Unclear error if key env vars aren't set: type Nothing has no field renew
If neither `AWS_ACCESS_KEY_ID` nor `AWS_SECRET_ACCESS_KEY` are set, the lack of auth credentials isn't caught gracefully when trying `AWSConfig`

It also takes ~20 seconds before erroring.
When the keys are set it returns promptly.

```julia
julia> using AWS

julia> AWS.AWSConfig(region = "us-east-2")
ERROR: type Nothing has no field renew
Stacktrace:
 [1] setproperty!(x::Nothing, f::Symbol, v::Function)
   @ Base ./Base.jl:34
 [2] AWSCredentials(; profile::Nothing)
   @ AWS ~/.julia/packages/AWS/4QyI4/src/AWSCredentials.jl:167
 [3] top-level scope
   @ REPL[8]:1
```

AWS.jl v1.25.1
Julia v1.5 and v1.7

mattBrzezinski:
Fixes #284

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -146,8 +146,8 @@ function AWSCredentials(; profile=nothing)
         profile = get(ENV, "AWS_PROFILE", get(ENV, "AWS_DEFAULT_PROFILE", nothing))
     end
 
-    # Define our search options, expected to be callable with no arguments. Should return
-    # `nothing` when credentials are not able to be located
+    # Define our search options, expected to be callable with no arguments.
+    # Throw NoCredentials if none are found
     functions = [
         env_var_credentials,
         () -> dot_aws_credentials(profile),
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -164,12 +164,13 @@ function AWSCredentials(; profile=nothing, throw_cred_error=true)
         creds === nothing || break
     end
 
-    # If credentials are nothing, default to throwing an error
-    creds === nothing && throw_cred_error && throw(NoCredentials("Can't find AWS credentials!"))
-
-    # Otherwise if credentials are nothing, return nothing
+    # If credentials are nothing, default to throwing an error, otherwise return nothing
     if creds === nothing
-        return nothing
+        if throw_cred_error
+            throw(NoCredentials("Can't find AWS credentials!"))
+        else
+            return nothing
+        end
     end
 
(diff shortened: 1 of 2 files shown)


morris25 (approved):

mattBrzezinski:
bors try

omus:
Documentation needs to be updated as this function no longer can return `nothing`. You may want to add a keyword so people can avoid having to catch the exception and just depend on `nothing` being returned.

mattBrzezinski:
> You may want to add a keyword so people can avoid having to catch the exception and just depend on `nothing` being returned.

I'm not able to think of a situation where you would attempt to get a credential object, and be okay with the result being `nothing`? I think the only time where you would want to hit this code is to know you have not setup credentials in some form on the machine that's running this, in which case you would want the error thrown to make debugging easier.

If you're trying to do some credential-less AWS calls using this package, you would be now using the `AbstractAWSConfig` to create the functionality yourself.

mattBrzezinski:
bors try

mattBrzezinski:
bors try

mattBrzezinski:
bors try-

mattBrzezinski:
bors try

src/AWSCredentials.jl
```diff
@@ -164,6 +164,14 @@ function AWSCredentials(; profile=nothing)
         creds === nothing || break
     end
 
+    # If credentials are nothing, default to throwing an error
+    creds === nothing && throw_cred_error && throw(NoCredentials("Can't find AWS credentials!"))
+
+    # Otherwise if credentials are nothing, return nothing
+    if creds === nothing
+        return nothing
+    end
```
omus, line 173:
```suggestion
    # If credentials are nothing, default to throwing an error, otherwise return nothing
    if creds === nothing
        if throw_cred_error
            throw(NoCredentials("Can't find AWS credentials!"))
        else
            return nothing
        end
    end
```

mattBrzezinski:
bors try

mattBrzezinski:
Unless someone wants to take it over in case of 💥 , I'll merge it on Monday (2021-02-22) and tag it, as I'm heading offline and am OOO tomorrow.

omus (approved):

PR #290, opened 2021-02-19
Improve accuracy and style of generated docstrings

### LINKED ISSUE #168: Issues with docstrings for high-level services

nickrobinson251:
- closes #168 and #301

Updates the docstring function name to match the actual function name, and removes the `()` (which implied all the functions took no args)
  - I also tried changing the optional parameter names to be snake_case too (as suggest by #168) but that causes issues and doesn't have any user-facing benefit, so let's not do that. 
  - I regenerated the pages `src/services/` pages so the effect of the changes in `src/AWSMetadataUtilities.jl` are visible


src/services/accessanalyzer.jl
```diff
@@ -5,274 +5,274 @@ using AWS.Compat
...
+- `analyzer_arn`: The Amazon resource name (ARN) of the analyzer.
+- `rule_name`: The name of the rule to apply.
 
 # Optional Parameters
-- `clientToken`: A client token.
+- `client_token`: A client token.
 """
-apply_archive_rule(analyzerArn, ruleName; aws_config::AbstractAWSConfig=global_aws_config()) = accessanalyzer("PUT", "/archive-rule", Dict{String, Any}("analyzerArn"=>analyzerArn, "ruleName"=>ruleName, "clientToken"=>string(uuid4())); aws_config=aws_config)
-apply_archive_rule(analyzerArn, ruleName, args::AbstractDict{String, <:Any}; aws_config::AbstractAWSConfig=global_aws_config()) = accessanalyzer("PUT", "/archive-rule", Dict{String, Any}(mergewith(_merge, Dict{String, Any}("analyzerArn"=>analyzerArn, "ruleName"=>ruleName, "clientToken"=>string(uuid4())), args)); aws_config=aws_config)
+apply_archive_rule(analyzer_arn, rule_name; aws_config::AbstractAWSConfig=global_aws_config()) = accessanalyzer("PUT", "/archive-rule", Dict{String, Any}("analyzer_arn"=>analyzer_arn, "rule_name"=>rule_name, "client_token"=>string(uuid4())); aws_config=aws_config)
```
nickrobinson251, line 19: oh, we probably don't want the keys in the dict to change, e.g. 
```diff
- Dict{String, Any}("analyzerArn"=>analyzerArn, "ruleName"=>ruleName
+ Dict{String, Any}("analyzer_arn"=>analyzer_arn, "rule_name"=>rule_name`
```

nickrobinson251 (commented):
Re the suggestions in #168

-  Docstring function name doesn't match actual function name (CreateConnectorProfile vs. create_connector_profile)
   - fixed in this PR
- Docstrings aren't actually attached to any function/method. I'd suggest attaching it to the function in general (e.g. function create_connector_profile end)
    - already fixed by #229 
- Docstring function name cannot be called without arguments. Just remove the ()
   - fixed in this PR
- Parameters should probably also be in snake_case instead of CamelCase?
   -  Let's just not do this (i'll update this PR to not make this change afterall as it causes issues; see comment below)

nickrobinson251:
Okay i pushed commits that fix #301 too

nickrobinson251:
I think this is done.

Here's an example of what docstrings looked like before this PR:

Screenshot 2021-03-02 at 14 46 40

And how that docstrings looks after this PR:
Screenshot 2021-03-02 at 14 46 00

src/AWSMetadataUtilities.jl
```diff
@@ -567,34 +569,43 @@ function _generate_high_level_definition(
...
         end
 
         # Add in the optional parameters if applicable
         if !isempty(optional_parameters)
-            operation_definition *= "# Optional Parameters\n"
+            operation_definition *= """
+                # Optional Parameters
+                Optional parameters can be passed as a `params::Dict{String,<:Any}`. Valid keys are:
+                """
+
```
mattBrzezinski, line 601:
```suggestion
```

src/AWSMetadataUtilities.jl
```diff
@@ -567,34 +569,43 @@ function _generate_high_level_definition(
...
 
             for (optional_key, optional_value) in optional_parameters
-                operation_definition *= "- `$optional_key`: $(optional_value["documentation"])\n"
+                operation_definition *= "- `\"$optional_key\"`: $(optional_value["documentation"])\n"
             end
         end
 
-        operation_definition *= repeat('"', 3)
+        operation_definition *= "\"\"\""
```
mattBrzezinski, line 608: I would revert this back, I think it's a bit easier to read in the original form

```suggestion
        operation_definition *= repeat('"', 3)
```

src/AWSMetadataUtilities.jl
```diff
@@ -567,34 +569,43 @@ function _generate_high_level_definition(
     - `String`: Docstring for the function
     """
     function _generate_docstring(function_name, documentation, required_parameters, optional_parameters)
+        function_name = _format_name(function_name)
+        args = join((_format_name(key) for (key, val) in required_parameters), ", ")
+        maybejoin = isempty(args) ? "" : ", "
         operation_definition = """
-            $(repeat('"', 3))
-                $function_name()
+            \"\"\"
```
mattBrzezinski, line 576:
```suggestion
            $(repeat('"', 3))
```

src/services/accessanalyzer.jl
mattBrzezinski: These parameters need to be in camelCase because that's what AWS wants.

https://github.com/aws/aws-sdk-js/blob/master/apis/accessanalyzer-2019-11-01.normal.json#L778-L799

mattBrzezinski: We could make the Julia functions `snake_case`, but I don't see too much value in spending time doing that.

src/AWSMetadataUtilities.jl
nickrobinson251: this confused both me and my editor 

but i can change it back if you feel strongly?

mattBrzezinski: Oh interesting, which editor are you using?
I'm not too too set on the `repeat()`, I find that reading quotes in strings to be a bit difficult at times to read, especially when modifying it.

nickrobinson251: actually i think it wasn't my editor that was confused by it, but JuliaFormatter (which means probably CSTParser)

nickrobinson251: yeah, on AWS.jl `master`, `using JuliaFormatter; format("src/AWSMetadataUtilities.jl")` errors (due to this `repeat` syntax afaict)

mattBrzezinski: I think that should be an issue logged w/ JuliaFormatter as it's valid?

nickrobinson251: true

mattBrzezinski:
bors try

mattBrzezinski:
These tests just need to be changed to represent the new changes in the high-level definitions,

https://github.com/JuliaCloud/AWS.jl/blob/7b282554bd59944bf254d8b450c1244592a2782b/test/AWSMetadataUtilities.jl#L285

mattBrzezinski:
bors try

nickrobinson251:
rebased on master after #291 was merged.
hopefully gtg now

mattBrzezinski:
bors try

Issue #283, opened 2021-02-12
bors callback not working


mattBrzezinski:
# Problem
`bors` runs CIs to test changes, but the callback does not occur to actually merge the changes.
See https://github.com/JuliaCloud/AWS.jl/pull/281 as an example.

mattBrzezinski:
Closed in #295

PR #292, opened 2021-02-20
Delete unused pirated method

### LINKED ISSUE #205: Type piracy of Base.merge
I'm not sure this is even used.
https://github.com/JuliaCloud/AWS.jl/blob/master/src/utilities.jl#L5

nickrobinson251:
- close #205 
- this is never used directly in either src/ or tests/, and AFAICT never hit at all, so i think we can just delete it


mattBrzezinski:
This came out of generating the high-level wrappers and is needed for `mergewith()` in this section of code:

https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWSMetadataUtilities.jl#L477-L503

nickrobinson251:
> is needed for mergewith() in this section

Sorry, i don't understand. How is it needed for `mergewith()`?

mattBrzezinski:
> > is needed for mergewith() in this section
> 
> Sorry, i don't understand. How is it needed for `mergewith()`?

I can't find it in, but it's used when merging dictionaries.

> https://github.com/JuliaLang/julia/blob/master/base/abstractdict.jl

I *think* you can probably replace these `mergewith()` calls with a `merge(b, a)` but would need to check how the code gets generated. The problematic case here I believe was merging, multiple dictionaries together and if they have the same keys which take priority, although looking at the REPL it's easily done with:

> merge(c,b,a)

I can spend more time looking into why these functions were created later on if need be.

nickrobinson251:
As far as i've been able to tell, this all works with this method deleted. But i could be missing something. I believe `mergewith(::Function, args...)` used to be spelled `merge(::Function, args...)`, so maybe we need `using Compat: mergewith` instead of this method?

Did CI pass? I don't understand how bors works, unfortunately. How do i see CI status?

mattBrzezinski:
> Did CI pass? I don't understand how bors works, unfortunately. How do i see CI status?

I will re-trigger it, we would need to test the regeneration of high-level wrappers, and run the CI on them to confirm they work before merging.

A 30 second crash course on `bors`:
- `bors` was created to prevent [semantic conflicts](https://bors.tech/essay/2017/02/02/pitch/), however... it has another nice feature which we're taking advantage of,
- `bors` limits the triggering of CI by a pre-defined user group, this is mostly the `JuliaCloud` org, with a few other users
- We do this because Invenia hosts the AWS account which is used for testing, we wouldn't want anyone to create a merge request where in `runtests.jl` you spawn 10k EC2 instances and have them run or other malicious reasonings
- `bors` just forked into a special testing environment `staging, trying`
- `bors try` will just test run the CI and *should* report back if it passes all tests or not, this callback is currently broken #283 
- `bors r+` will test the CI, **and** if it all passes automatically merge the branch into `master`

List of all commands are [here](https://bors.tech/documentation/), but not only people on the list can trigger these commands.

mattBrzezinski:
bors try

mattBrzezinski:
Looks like a timing issue, will try again

mattBrzezinski:
bors try

mattBrzezinski:
Just needs a rebase for `bors` to run again

mattBrzezinski:
bors try

mattBrzezinski (approved):

PR #293, opened 2021-02-22
Handle '=' in S3 filenames

nickrobinson251:
- Same as https://github.com/JuliaCloud/AWS.jl/pull/285 but adds `'='` to the list, as this is a valid character in S3 filenames, but if unescaped interferes with the GET / PUT request syntax

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -332,37 +332,6 @@ end
 
 
 """
-    _clean_s3_uri(uri::AbstractString)
-
-Escape special AWS S3 characters properly.
-
-AWS S3 allows for various special characters in file names, these characters are not being
-properly escaped before we make the requests.
-
-We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
-are used in the filepathing for sub-directories.
-
-# Arguments
-- `uri::AbstractString`: URI to to cleaned
-
-# Returns
-- `String`: URI with characters escaped
-"""
-function _clean_s3_uri(uri::AbstractString)
-    chars_to_clean = (
-        ' ' => "%20",
-        '!' => "%21",
-        ''' => "%27",
-        '(' => "%28",
-        ')' => "%29",
-        '*' => "%2A",
-        '=' => "%3D"
-    )
-    return reduce(replace, chars_to_clean, init=uri)
-end
-
-
-"""
     submit_request(aws::AbstractAWSConfig, request::Request; return_headers::Bool=false)
 
 Submit the request to AWS.
@@ -396,9 +365,6 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
 
     request.headers["User-Agent"] = user_agent[]
     request.headers["Host"] = HTTP.URI(request.url).host
-    if request.service == "s3"
-        request.url = _clean_s3_uri(request.url)
-    end
 
     @repeat 3 try
         credentials(aws) === nothing || sign!(aws, request)
@@ -558,6 +524,36 @@ function (service::RestXMLService)(
     request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
 )
+    """
+        _clean_s3_uri(uri::AbstractString)
+
+    Escape special AWS S3 characters properly.
+
+    AWS S3 allows for various special characters in file names, these characters are not being
+    properly escaped before we make the requests.
+
+    We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
+    are used in the filepathing for sub-directories.
+
+    # Arguments
+    - `uri::AbstractString`: URI to to cleaned
+
+    # Returns
+    - `String`: URI with characters escaped
+    """
+    function _clean_s3_uri(uri::AbstractString)
+        chars_to_clean = (
+            ' ' => "%20",
+            '!' => "%21",
+            ''' => "%27",
+            '(' => "%28",
+            ')' => "%29",
+            '*' => "%2A",
+            '=' => "%3D"
+        )
+        return reduce(replace, chars_to_clean, init=uri)
+    end
+
     request = Request(
         service=service.name,
         api_version=service.api_version,
@@ -578,6 +574,9 @@ function (service::RestXMLService)(
     delete!(args, "body")
     return_headers = _return_headers(args)
 
+    if request.service == "s3"
+        request_uri = _clean_s3_uri(request_uri)
+    end
     request.resource = _generate_rest_resource(request_uri, args)
     query_str = HTTP.escapeuri(args)
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -332,7 +332,7 @@ end
 
 
 """
-    _clean_uri(uri::AbstractString)
+    _clean_s3_uri(uri::AbstractString)
 
 Escape special AWS S3 characters properly.
 
@@ -348,17 +348,16 @@ are used in the filepathing for sub-directories.
 # Returns
 - `String`: URI with characters escaped
 """
-function _clean_uri(uri::AbstractString)
-    chars_to_clean = [
+function _clean_s3_uri(uri::AbstractString)
+    chars_to_clean = (
         ' ' => "%20",
         '!' => "%21",
         ''' => "%27",
         '(' => "%28",
         ')' => "%29",
         '*' => "%2A",
-        '=' => "%3D",
-    ]
-
+        '=' => "%3D"
+    )
     return reduce(replace, chars_to_clean, init=uri)
 end
 
@@ -397,7 +396,9 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
 
     request.headers["User-Agent"] = user_agent[]
     request.headers["Host"] = HTTP.URI(request.url).host
-    request.url = _clean_uri(request.url)
+    if request.service == "s3"
+        request.url = _clean_s3_uri(request.url)
+    end
 
     @repeat 3 try
         credentials(aws) === nothing || sign!(aws, request)
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -355,7 +355,8 @@ function _clean_uri(uri::AbstractString)
         ''' => "%27",
         '(' => "%28",
         ')' => "%29",
-        '*' => "%2A"
+        '*' => "%2A",
+        '=' => "%3D",
     ]
 
     return reduce(replace, chars_to_clean, init=uri)
--- a/test/AWS.jl
+++ b/test/AWS.jl
@@ -743,7 +743,7 @@ end
 
     @testset "low-level s3" begin
         bucket_name = "aws-jl-test---" * _now_formatted()
-        file_name = "*)('! .txt"  # Special characters which S3 allows
+        file_name = "*)=('! .txt"  # Special characters which S3 allows
 
         function _bucket_exists(bucket_name)
             try


mattBrzezinski:
bors try

nickrobinson251:
CI failure look unrelated. I think it's because of the `operation=add` in the glacier POST request for `add_tags_to_vault` where the `=` should _not_ be cleaned 
```
add_tags_to_vault(accountId, vaultName; aws_config::AbstractAWSConfig=global_aws_config()) = glacier("POST", "/$(accountId)/vaults/$(va    ultName)/tags?operation=add"; aws_config=aws_config)
add_tags_to_vault(accountId, vaultName, args::AbstractDict{String, <:Any}; aws_config::AbstractAWSConfig=global_aws_config()) = glacier
    ("POST", "/$(accountId)/vaults/$(vaultName)/tags?operation=add", args; aws_config=aws_config)
```
unfortunately this is exactly the structure we _do_ want to escape for S3 GET/PUT requests (e.g. we want the `Key` cleaned in`s3("GET", "/$(Bucket)/$(Key)"; aws_config=aws_config)`)

I guess we should special-case S3 somehow? What do you think @mattBrzezinski ?

mattBrzezinski:
> I guess we should special-case S3 somehow? What do you think @mattBrzezinski ?

This begins to open the flood gates to spaghetti. We would basically need to start special casing services in the function which calls the HTTP request. As of `v1.25.2` we support all of the safe characters which are defined by AWS:

> https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

As much as it would be nice to allow for more support, I would begin pushing back against this as it begins to impact the code base.

nickrobinson251:
as far as i can tell the `_clean_uri` function added in  #285 only applies to S3 anyway, so we just need an `if` statement to avoid needlesly running it on other services, and we already have an `if "s3" ...` statement in `sign_aws4!` so i think this pattern is fine. Let me know what you think after the last commit

mattBrzezinski:
> as far as i can tell the `_clean_uri` function added in #285 only applies to S3 anyway, so we just need an `if` statement to avoid needlesly running it on other services, and we already have an `if "s3" ...` statement in `sign_aws4!` so i think this pattern is fine. Let me know what you think after the last commit

In its current state #285 applies to all AWS services. One problematic thing about limiting this to certain services is managing the relationship between like-wise ones. In this instance S3 and Glacier, which I *think* support both the same safe characters ( need to double check). This becomes an issue if a user archives a file into Glacier with an equal sign in it, then attempts to access it via this package, which they would be unable to do so. However, that file exists in S3, it would work perfectly fine.

It's a bit of a problem when making a package to call AWS as generically as possible, since AWS doesn't have a standard definition for its APIs, its up to the specific teams owning the products to come up with some reasoning. There are some places in this codebase where we absolutely have to special case things such as glacier requiring a specific header in its requests.

> https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWSServices.jl#L111

But I would prefer to avoid this stuff as much as possible. I think that enabling all safe characters which AWS defines is an appropriate solution, as it doesn't clutter up the code too much (for now). I think that going down the non-required special casing for services can lead to a monolith of `if` statements. If end users really really want to use these special characters in their files, they still are able to, but they must escape them when making the request themselves.

e.g.
```julia
S3.get_object("my-bucket", "my=file")  # Will not work
S3.get_object("my-bucket", "my%3Dfile")  # Will work
```

If we really would want to allow for these characters we would need to switch back to an `AWSCore.jl` style of function where you pass in the `resource` which can then be `HTTP.escapeuri()` to clean it up,

> https://github.com/JuliaCloud/AWSCore.jl/blob/master/src/Services.jl#L2782

But this would involve re-writing our low-level functors, high-level definitions to follow suit, and a new major release:

> https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWS.jl#L555

src/AWS.jl
```diff
@@ -396,7 +396,9 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
 
     request.headers["User-Agent"] = user_agent[]
     request.headers["Host"] = HTTP.URI(request.url).host
-    request.url = _clean_uri(request.url)
+    if contains(request.service, "s3")
```
mattBrzezinski, line 399: I think this should just be a straight check, I am not familiar with S3 Control or S3 Outposts, but I'm not sure if we would want to proceed with this cleansing on those services.

nickrobinson251: yeah, fair -- will change :)

mattBrzezinski:
bors try

iamed2:
I feel like there's some way to do this generically. Like clearly we know when we have semantic `=` (just to separate query params). Where do the URIs come from?

mattBrzezinski:
> I feel like there's some way to do this generically. Like clearly we know when we have semantic `=` (just to separate query params). Where do the URIs come from?

URIs come in two places

1) High-level, `$(Bucket)/$(Key)`, which then goes down to the lowl-level

https://github.com/JuliaCloud/AWS.jl/blob/master/src/services/s3.jl#L746

2) Low-level

https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWS.jl#L556

I guess we could move this `_clean_uri()` call into each of the functors, or at the very least just `RestXML`.

mattBrzezinski:
@nickrobinson251 needs a rebase to be able to run `bors`

src/AWS.jl
```diff
@@ -348,16 +348,16 @@ are used in the filepathing for sub-directories.
 # Returns
 - `String`: URI with characters escaped
 """
-function _clean_uri(uri::AbstractString)
-    chars_to_clean = [
+function _clean_s3_uri(uri::AbstractString)
```
mattBrzezinski, line 351: Since this is now targeting S3 specifically, we should probably move it into the `RestXMLService` functor as a nested function, and call it on the `request_uri` parameter.

mattBrzezinski:
bors try

mattBrzezinski:
bors try-

mattBrzezinski:
bors try

src/AWS.jl
nickrobinson251: > move it into the RestXMLService functor as a nested function

You mean define the function inside there too? like this
```diff
diff --git a/src/AWS.jl b/src/AWS.jl
index 5efeb2e..60b4d53 100644
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -331,37 +331,6 @@ function _http_request(request::Request)
 end
 
 
-"""
-    _clean_s3_uri(uri::AbstractString)
-
-Escape special AWS S3 characters properly.
-
-AWS S3 allows for various special characters in file names, these characters are not being
-properly escaped before we make the requests.
-
-We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
-are used in the filepathing for sub-directories.
-
-# Arguments
-- `uri::AbstractString`: URI to to cleaned
-
-# Returns
-- `String`: URI with characters escaped
-"""
-function _clean_s3_uri(uri::AbstractString)
-    chars_to_clean = (
-        ' ' => "%20",
-        '!' => "%21",
-        ''' => "%27",
-        '(' => "%28",
-        ')' => "%29",
-        '*' => "%2A",
-        '=' => "%3D"
-    )
-    return reduce(replace, chars_to_clean, init=uri)
-end
-
-
 """
     submit_request(aws::AbstractAWSConfig, request::Request; return_headers::Bool=false)
 
@@ -396,9 +365,6 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
 
     request.headers["User-Agent"] = user_agent[]
     request.headers["Host"] = HTTP.URI(request.url).host
-    if request.service == "s3"
-        request.url = _clean_s3_uri(request.url)
-    end
 
     @repeat 3 try
         credentials(aws) === nothing || sign!(aws, request)
@@ -718,6 +684,36 @@ function (service::RestJSONService)(
     request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, String}();
     aws_config::AbstractAWSConfig=global_aws_config(),
 )
+    """
+        _clean_s3_uri(uri::AbstractString)
+
+    Escape special AWS S3 characters properly.
+
+    AWS S3 allows for various special characters in file names, these characters are not being
+    properly escaped before we make the requests.
+
+    We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
+    are used in the filepathing for sub-directories.
+
+    # Arguments
+    - `uri::AbstractString`: URI to to cleaned
+
+    # Returns
+    - `String`: URI with characters escaped
+    """
+    function _clean_s3_uri(uri::AbstractString)
+        chars_to_clean = (
+            ' ' => "%20",
+            '!' => "%21",
+            ''' => "%27",
+            '(' => "%28",
+            ')' => "%29",
+            '*' => "%2A",
+            '=' => "%3D"
+        )
+        return reduce(replace, chars_to_clean, init=uri)
+    end
+
     return_headers = _return_headers(args)
 
     request = Request(
@@ -737,6 +733,9 @@ function (service::RestJSONService)(
     end
 
     request.url = generate_service_url(aws_config, request.service, request.resource)
+    if request.service == "s3"
+        request.url = _clean_s3_uri(request.url)
+    end
 
     if !isempty(service.service_specific_headers)
         merge!(request.headers, service.service_specific_headers)
```

or did you just mean move the call site?

mattBrzezinski: Yes to the `git diff` shown above.

iamed2:
> URIs come in two places

Oh I meant when generating the services

mattBrzezinski:
> > URIs come in two places
> 
> Oh I meant when generating the services

They are defined as,

https://github.com/aws/aws-sdk-js/blob/master/apis/s3-2006-03-01.normal.json#L660

nickrobinson251:
> when generating the services

my understanding is that we grab them from the `aws-sdk-js` repo (e.g. here's the glacier `add_tags_to_vault` request: https://github.com/aws/aws-sdk-js/blob/54f8555bd94d33a1754a44a35286f1d9e31c28a3/apis/glacier-2012-06-01.min.json#L69) as part of [`_generate_high_level_wrapper`](https://github.com/JuliaCloud/AWS.jl/blob/5cd2aec9c30381f80b423b7e5ae5970feb2465f8/src/AWSMetadata.jl#L55-L70) 

which calls `_generate_high_level_definitions` which in turn calls `_generate_rest_operation_defintion`, and [that fixes up the syntax to be valid in Julia](https://github.com/JuliaCloud/AWS.jl/blob/5cd2aec9c30381f80b423b7e5ae5970feb2465f8/src/AWSMetadataUtilities.jl#L448-L451)

so maybe we could do something at that stage?

iamed2:
Ah yeah, so the `=` is explicitly provided for Glacier; could we not make it so every interpolated value is escaped, rather than the whole URI? (with no copy if there's no necessary change)

mattBrzezinski:
> which calls `_generate_high_level_definitions` which in turn calls `_generate_rest_operation_defintion`, and [that fixes up the syntax to be valid in Julia](https://github.com/JuliaCloud/AWS.jl/blob/5cd2aec9c30381f80b423b7e5ae5970feb2465f8/src/AWSMetadataUtilities.jl#L448-L451)
> 
> so maybe we could do something at that stage?

This issue needs to be fixed at the low-level wrapper stage. We *could* wrap the result of your link in `HTTP.escapeuri()` but if a user attempt to use the low-level definition to make a request it will fail, it would also be a lot of redundant function calls.

mattBrzezinski:
> Ah yeah, so the `=` is explicitly provided for Glacier; could we not make it so every interpolated value is escaped, rather than the whole URI? (with no copy if there's no necessary change)

In @nickrobinson251 commit we moved the replacement code inside of `RestXMLService` functor, which runs on the `request_uri` before any query strings are appended to it.

AFAIK this is only affecting S3 (and possibly Glacier). If we find other service types which this is affecting, we can extract this now nested function, and call it in other functors.

mattBrzezinski:
The outline of the problem:

- We need to replace special characters in S3 (and presumably for Glacier as well) calls, other services are unknown
- This needs to be done before queries are appended to URIs because these special characters do *not* want to be replaced in the query section
- This must be done in the low-level definition form as the high level calls the low-level code, and there's no point in doing this twice

I'm happy to have this cleanse code live only in `RestXMLService` currently, if this issue crops up elsewhere we can extract the now nested function and call it in the other service functors as well.

To test, I would imagine if you want to add a `tag` to a resource and input some interesting characters you may be able to reproduce this in other services.

nickrobinson251:
Cool :) Just let me know what more, if anything, needs doing in the PR, and what can be moved to an issue

nickrobinson251:
on the more general issue

> This must be done in the low-level definition form

I think this is the bit i don't understand. For the low-level API, don't users have to provide there own request in full? So there it makes sense that it's left to the user to provide a valid request (with appropriate syntax). 

e.g. if i call `s3("GET", "/bucket/key=value/") then i would find it surprising behaviour if this automatically got transformed into `"/bucket/key%3Dvalue/"`, likewise if `glacier("POST", "vaults/key=value/tags?operation=add")` became `"vaults/key%3Dvalue/tags?operation=add"` (even though the latter examples are the correct syntax).
 
Do you think that the low-level API should transform `glacier("POST", "vaults/key=value/tags?operation=add")` -> `glacier("POST", "vaults/key%3Dvalue/tags?operation=add")`?

Or am i misunderstanding something? (very possible, as i'm new to AWS.jl)

mattBrzezinski:
> I think this is the bit i don't understand. For the low-level API, don't users have to provide there own request in full? So there it makes sense that it's left to the user to provide a valid request (with appropriate syntax).

Yes. For example doing a get object request users would call this as:

```julia
using AWS.AWSServices: s3
s3("GET", "/my-bucket/my=file")
```

However in the case above the `=` character is not being escaped properly in the underlying HTTP request to AWS, and will fail. Currently if a user wants this request to work they would need to call it as:

```julia
s3("GET", "/my-bucket/my%3Dfile")
```

The high-level wrapper just wraps around this low-level version to provide a more convenient function for the end user. In this example, it's relatively easy to know how to make a get object S3 request. But some other APIs endpoints are a bit more confusing, this is where the high-level wrapper fits in. It abstracts some of this away so all you need to know is:

1) The service you want to use
2) The endpoint you wish to call
3) The parameters for the endpoint

So all you need to do now is,

```julia
using AWS
@service S3

S3.get_object(bucket, key)
```

It's also nice in the REPL because you can do things such as,

```julia
julia> S3.
abort_multipart_upload                          get_bucket_logging                               put_bucket_accelerate_configuration
complete_multipart_upload                       get_bucket_metrics_configuration                 put_bucket_acl
copy_object                                     get_bucket_notification                          put_bucket_analytics_configuration
create_bucket                                   get_bucket_notification_configuration            put_bucket_cors
create_multipart_upload                         get_bucket_ownership_controls                    put_bucket_encryption
...

help?> S3.abort_multipart_upload
  AbortMultipartUpload()

  This operation aborts a multipart upload. After a multipart upload is aborted, no additional parts can be uploaded using that upload ID. The storage consumed by any
  previously uploaded parts will be freed. However, if any part uploads are currently in progress, those part uploads might or might not succeed. As a result, it
  might be necessary to abort a given multipart upload multiple times in order to completely free all storage consumed by all parts. To verify that all parts have
  been removed, so you don't get charged for the part storage, you should call the ListParts operation and ensure that the parts list is empty. For information about
  permissions required to use the multipart upload API, see Multipart Upload API and Permissions. The following operations are related to AbortMultipartUpload:
  CreateMultipartUpload UploadPart CompleteMultipartUpload ListParts ListMultipartUploads 

  Required Parameters
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

    •    Bucket: The bucket name to which the upload was taking place. When using this API with an access point, you must direct requests to the access point
        hostname. The access point hostname takes the form AccessPointName-AccountId.s3-accesspoint.Region.amazonaws.com. When using this operation with an access
        point through the AWS SDKs, you provide the access point ARN in place of the bucket name. For more information about access point ARNs, see Using Access
        Points in the Amazon Simple Storage Service Developer Guide. When using this API with Amazon S3 on Outposts, you must direct requests to the S3 on
        Outposts hostname. The S3 on Outposts hostname takes the form AccessPointName-AccountId.outpostID.s3-outposts.Region.amazonaws.com. When using this
        operation using S3 on Outposts through the AWS SDKs, you provide the Outposts bucket ARN in place of the bucket name. For more information about S3 on
        Outposts ARNs, see Using S3 on Outposts in the Amazon Simple Storage Service Developer Guide.

    •    Key: Key of the object for which the multipart upload was initiated.

    •    uploadId: Upload ID that identifies the multipart upload.

  Optional Parameters
  ≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡≡

    •    x-amz-expected-bucket-owner: The account id of the expected bucket owner. If the bucket is owned by a different account, the request will fail with an
        HTTP 403 (Access Denied) error.

    •    x-amz-request-payer: 
```

As well as other nice things like peek a function's parameters in Visual Studio Code, etc. Doing a call such as the following does *work*:

> s3("GET", "/my-bucket/my%3Dfile")

But this feels bad, and end users shouldn't need to escape their own special characters. So instead we can do this for them with the `_clean_uri()` function. By cleaning for them they can do the first call, and have it succeed.

> e.g. if i call `s3("GET", "/bucket/key=value/")` then i would find it surprising behaviour if this automatically got transformed into `"/bucket/key%3Dvalue/"`, likewise if `glacier("POST", "vaults/key=value/tags?operation=add")` became `"vaults/key%3Dvalue/tags?operation=add"` (even though the latter examples are the correct syntax).

Unless you're debugging or looking into the implementation you won't see this, it's only used when making the request to AWS. For example:

```julia
using AWS
@service S3

S3.put_object("mattbr-test-bucket", "my%3Dfile", Dict("body"=>"hello world!"))
> UInt8[]

resp = S3.list_objects("mattbr-test-bucket")
[r["Key"] for r in resp["Contents"]]
> 24-element Array{String,1}:
>  "my=file"
```

nickrobinson251:
i agree with all that. I'm just a bit confused about the low-level use-case for automatic escaping. 

It makes complete sense to me for the high-level API to handle this stuff automatically
e.g. for `add_tags_to_vault("123", "vaultname=value")` to "just work", for example via having generated code that looks like
```
add_tags_to_vault(accountId, vaultName; aws_config::AbstractAWSConfig=global_aws_config()) = glacier("POST", "/$(_clean_uri(accountId))/vaults/$(_clean_uri(vaultName))/tags?operation=add"; aws_config=aws_config)
```
and so the code that get's called (via the high-level API generating the write low-level API call) is `glacier("POST", "/123/vaults/vaultname%3Dvalue/tags?operation=add")`

But if as a user i call the low-level API directly, e.g. with `glacier("POST", "/123/vaults/vaultname=value/tags?operation=add")`, then that's what i'd expect to be call (and _not_ `glacier("POST", "/123/vaults/vaultname%3Dvalue/tags?operation=add")`)

I can't tell if you agree with that or not :)

nickrobinson251:
Anyway, this seems to be a bigger discussion than this PR so I'd be happy to move it to an issue, or leave it to @mattBrzezinski and @iamed2 to figure out 😂 

Are there more changes you'd like to see in this PR?

Thanks for the help :)

mattBrzezinski:
> i agree with all that. I'm just a bit confused about the low-level use-case for automatic escaping.
> 
> It makes complete sense to me for the high-level API to handle this stuff automatically
> e.g. for `add_tags_to_vault("123", "vaultname=value")` to "just work", for example via having generated code that looks like
> 
> ```
> add_tags_to_vault(accountId, vaultName; aws_config::AbstractAWSConfig=global_aws_config()) = glacier("POST", "/$(_clean_uri(accountId))/vaults/$(_clean_uri(vaultName))/tags?operation=add"; aws_config=aws_config)
> ```
> 
> and so the code that get's called (via the high-level API generating the write low-level API call) is `glacier("POST", "/123/vaults/vaultname%3Dvalue/tags?operation=add")`
> 
> But if as a user i call the low-level API directly, e.g. with `glacier("POST", "/123/vaults/vaultname=value/tags?operation=add")`, then that's what i'd expect to be call (and _not_ `glacier("POST", "/123/vaults/vaultname%3Dvalue/tags?operation=add")`)
> 
> I can't tell if you agree with that or not :)

I slid a bit into the cabbage patch in my last post.

Looking more into it, AWS Glacier flat out doesn't let you create a vault with these special characters so you won't run into this issue.

```
Glacier.create_vault("-", "mattbr=vault")

> AWS.AWSExceptions.AWSException("400", "Invalid vault name: mattbr=vault=name. Valid vault names may consist only of a-z, A-Z, 0-9 \".\", \"-\", and \"_\". and have length limit of 255", ...
```

Adding `tags` with special characters also does not matter because this is a `RestJSONService` and things such as `=` are capture in the request' body which is all JSON and AWS is happy about it. So this won't be problematic at all.

I am happy to move forward with these changes, I'll give Eric a chance to comment if they'd like otherwise I'll proceed forward with merging later tonight / tomorrow morning.

nickrobinson251:
having moved the location of the `_clean_s3_uri` call (https://github.com/JuliaCloud/AWS.jl/pull/293/commits/b01d71d0cffe99ed38587922125da156783a9974), this no longer works as expected. in particular `list_objects` seems to be returning objects that should be excluded by a `"prefix"` arg. I don't yet understand what yet is going on.

nickrobinson251:
with the changes in this PR, `S3.get_object(bucket, key)` now works as expected when `key` contain `=`, e.g. "my=file", but `S3.list_objects(bucket, Dict("prefix" => key))` no longer works as expected (whatever the `key` is). 😭 

I suspect this is to do with when `key` gets passed to the low-level API
`get_objects(bucket, key) = s3("GET", "/$(bucket)/$(key)")` 
whereas
`list_objects(bucket, args) = s3("GET", "/$(bucket)", args)` (with the `key` being in `args`, and i'm not sure when this gets populated into the request)

ooof this stuff is fiddly

src/AWS.jl
```diff
@@ -588,6 +586,9 @@ function (service::RestXMLService)(
     end
 
     request.url = generate_service_url(aws_config, request.service, request.resource)
+    if request.service == "s3"
+        request.url = _clean_s3_uri(request.url)
```
mattBrzezinski, line 590: Note that `request_uri` would be something like `/my-bucket/my=file`, while the full URL would be something like `http://s3.us-east-1.amazonaws.com/my-bucket/my=file/?prefix=foobar` or something like that. We only want to scrub the URI otherwise we would be cleaning `prefix=foobar` which will break.

You would want this instead, also might be worthwhile to turn it into a `!` function.

```suggestion
        request_uri = _clean_s3_uri(request_uri)
```

Then line should also be moved above line 577 (in this commit) where the `request.resource` is set.

nickrobinson251: Thanks! 

> might be worthwhile to turn it into a `!` function

I don't think this is possible, since a `String` is not mutable

mattBrzezinski: Oh right, :4head:

mattBrzezinski:
bors try

mattBrzezinski:
bors-

mattBrzezinski:
bors try-

mattBrzezinski:
bors try

mattBrzezinski (approved):

mattBrzezinski:
Lets 🚢 🇮🇹

mattBrzezinski:
Could you rebase on `master`? @nickrobinson251 I think that `bors` is now fixed.

mattBrzezinski:
Looks like a timing issue. Trying again.

PR #298, opened 2021-02-25
use URIs.escapepath instead of custom S3 escaping

### LINKED ISSUE #297: Some characters are not properly URI escaped in S3 URIs
I discovered that some S3 paths which are fine using the AWS CLI result in
errors using AWSS3.jl, and I think this is the right place to report because
this is where the escaping happens.

Specifically, I found that any path with a `+` in it would cause an
`AWS.AWSExceptions.AWSException` with `"SignatureDoesNotMatch", "The request
signature we calculated does not match the signature you provided. Check your
key and signing method."`.  Manually escaping the `+` would cause the request to
return the contents of the object with `+` in its key.

The custom URI escaping rules used here don't seem to reflect the most recent
AWS guidelines about what characters are okay, which "might require special
handling", and which are actively to be avoided (see
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html).

`+` is on the list of characters that "might require special handling", but it
also seems to be automatically escaped somewhere in AWS's pipeline leading to
the signature mismatch.  I've also confirmed informally that a key containing
all of those characters surrounds a roundtrip when they're URI encoded and show
up in the correct unencoded form with `aws s3 ls ...`.

I think the fix is pretty simple but wanted to report the issue here first; I'll
open a PR soon.

kleinschmidt:
Fixes #297

The current custom implementation of URI escaping misses many characters that
are valid in S3 keys
https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

URIs.jl now includes an `escapepath` function that leaves `/` characters
un-escaped and is suitable for escaping S3 keys.  This PR replaces the custom
escaping function with that one, and changes the "weird filename" in the tests
to use all the characters that are either explicitly allowed or "might require
special handling in code" according to the current AWS docs (linked above).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
--- a/Project.toml
+++ b/Project.toml
@@ -30,6 +30,7 @@ MbedTLS = "0.6, 0.7, 1"
 Mocking = "0.7"
 OrderedCollections = "1"
 Retry = "0.3, 0.4"
+URIs = "1"
 XMLDict = "0.3, 0.4"
 julia = "1"
 
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -4,6 +4,7 @@ using Compat: Compat
 using Base64
 using Dates
 using HTTP
+using URIs
 using MbedTLS
 using Mocking
 using OrderedCollections: LittleDict, OrderedDict
@@ -524,36 +525,6 @@ function (service::RestXMLService)(
     request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
 )
-    """
-        _clean_s3_uri(uri::AbstractString)
-
-    Escape special AWS S3 characters properly.
-
-    AWS S3 allows for various special characters in file names, these characters are not being
-    properly escaped before we make the requests.
-
-    We cannot call `HTTP.escapeuri(request.uri)` because this will escape `/` characters which
-    are used in the filepathing for sub-directories.
-
-    # Arguments
-    - `uri::AbstractString`: URI to to cleaned
-
-    # Returns
-    - `String`: URI with characters escaped
-    """
-    function _clean_s3_uri(uri::AbstractString)
-        chars_to_clean = (
-            ' ' => "%20",
-            '!' => "%21",
-            ''' => "%27",
-            '(' => "%28",
-            ')' => "%29",
-            '*' => "%2A",
-            '=' => "%3D"
-        )
-        return reduce(replace, chars_to_clean, init=uri)
-    end
-
     request = Request(
         service=service.name,
         api_version=service.api_version,
@@ -575,7 +546,7 @@ function (service::RestXMLService)(
     return_headers = _return_headers(args)
 
     if request.service == "s3"
-        request_uri = _clean_s3_uri(request_uri)
+        request_uri = URIs.escapepath(request_uri)
     end
     request.resource = _generate_rest_resource(request_uri, args)
     query_str = HTTP.escapeuri(args)
--- a/test/AWS.jl
+++ b/test/AWS.jl
@@ -743,7 +743,10 @@ end
 
     @testset "low-level s3" begin
         bucket_name = "aws-jl-test---" * _now_formatted()
-        file_name = "*)=('! .txt"  # Special characters which S3 allows
+        # Special characters which S3 explicitly allows or does not explicitly
+        # suggest avoiding "might require special handling"
+        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+        file_name = "!/-_.*'()&\$@=;:+ ,?.txt"
 
         function _bucket_exists(bucket_name)
             try


ararslan:
bors try

Project.toml
```diff
@@ -16,6 +16,7 @@ Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
```
mattBrzezinski, line 19: Just needs a compat statement

kleinschmidt: just gotta find out when that function was introduced :)

mattBrzezinski:
Brilliant, open source software at it's finest! 😄
Thanks for the PR, this all LGTM I'll let the `bors try` go through and then `r+` afterwards.

ararslan:
The bors try finished but the builds failed for a seemingly unrelated reason, unless the escaping is changing something it shouldn't be?

kleinschmidt:
It looks like there migth be something about the escaping in the bucket name in that test:

```
  LoadError: AWSException("NoSuchBucket", "The specified bucket does not exist", EzXML.Document(EzXML.Node(<DOCUMENT_NODE@0x0000000007a29470>)), HTTP.ExceptionRequest.StatusError(404, "GET", "/anewbucket%3Flist-type%3D2", HTTP.Messages.Response:
```

mattBrzezinski:
> The bors try finished but the builds failed for a seemingly unrelated reason, unless the escaping is changing something it shouldn't be?

> It looks like there migth be something about the escaping in the bucket name in that test:
> 
> ```
>   LoadError: AWSException("NoSuchBucket", "The specified bucket does not exist", EzXML.Document(EzXML.Node(<DOCUMENT_NODE@0x0000000007a29470>)), HTTP.ExceptionRequest.StatusError(404, "GET", "/anewbucket%3Flist-type%3D2", HTTP.Messages.Response:
> ```

These are unrelated to your changes. It's around:

> https://github.com/JuliaCloud/AWS.jl/blob/master/test/minio.jl#L26

Specifically that multiple runners going in parallel, some kill that bucket and fail others in contention.
I can get a PR for this tomorrow morning fixing it, merged in, and we can re-run this PR afterwards.

I'm currently unavailable, unless you'd wish to make the changes in here yourself. All that needs to happen is naming of the bucket like:

> https://github.com/JuliaCloud/AWS.jl/blob/master/test/AWS.jl#L693

kleinschmidt:
No huge rush, tomorrow is fine by me.

nickrobinson251:
this is so much nicer -- thanks!

kleinschmidt:
Okay I think this might actually be my fault:

https://github.com/JuliaCloud/AWS.jl/runs/1974898387#step:10:127

From what I can tell it's trying to delete a bucket by POSTing a request to `/$(bucket)?delete` but the `?` is getting escaped to `%3F`.  ~~I dug around a bit and can't figure out where that query is being added;~~ The `?delete` is being added in the test code itself:

https://github.com/JuliaCloud/AWS.jl/blob/master/test/AWS.jl#L792

It doesn't look like there's any way for a request URI to be generated like this _except_ by this manual method, so I'm not sure what the test is trying to do or whether it should be changed (it is in the "low level" section after all so maybe that's intentional?).  If the test is important than I can add some logic to first split off the query string and escape just the path part but I want to make sure that's necessary.

nickrobinson251:
`AWSServices.s3("POST", "/$bucket_name?delete", Dict("body" => body))` looks to me like a pretty standard "low-level" call that we want to be valid, but maybe @mattBrzezinski can correct me on that.

I think this is something we ran into when the `_clean_uri` step was applied to all services (not just S3), e.g. [with Glacier here](https://github.com/JuliaCloud/AWS.jl/pull/293#issuecomment-783437310), and we "worked around" it by only applying it to S3 (none of the high-level `S3` functions make a low-level `AWSServices.s3` call with special chars like `?`). 

If we can magically escape low-level calls correctly that'd be cool. 

This is what @mattBrzezinski and I were discussing in #293 (e.g. https://github.com/JuliaCloud/AWS.jl/pull/293#issuecomment-783653553 and https://github.com/JuliaCloud/AWS.jl/pull/293#issuecomment-783677503). I was unsure magically escaping low-level calls was desireable, but i think Matt thinks it is desireable. We just need to make it possible (but knowing where to split the string into bits that need escaping and bits that don't)

kleinschmidt:
The thing that got me slightly tripped up there is that there's a `<Delete>` in the body of the request too; I know next to nothing about interacting with AWS so I have no idea whether the query and the body are both necessary there.

mattBrzezinski:
> Okay I think this might actually be my fault:
> 
> https://github.com/JuliaCloud/AWS.jl/runs/1974898387#step:10:127

Ah okay, I didn't see that job failure. Yeah this is where things are a bit problematic, and were discussed a bunch in #293

> From what I can tell it's trying to delete a bucket by POSTing a request to `/$(bucket)?delete` but the `?` is getting escaped to `%3F`. ~I dug around a bit and can't figure out where that query is being added;~ The `?delete` is being added in the test code itself:
> 
> https://github.com/JuliaCloud/AWS.jl/blob/master/test/AWS.jl#L792

This is actually deleting a file from an S3 bucket. The specification comes from [AWS](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html)

```
POST /?delete HTTP/1.1
Host: Bucket.s3.amazonaws.com
x-amz-mfa: MFA
x-amz-request-payer: RequestPayer
x-amz-bypass-governance-retention: BypassGovernanceRetention
x-amz-expected-bucket-owner: ExpectedBucketOwner
<Delete xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
   <Object>
      <Key>string</Key>
      <VersionId>string</VersionId>
   </Object>
   ...
   <Quiet>boolean</Quiet>
</Delete>
```

> It doesn't look like there's any way for a request URI to be generated like this _except_ by this manual method, so I'm not sure what the test is trying to do or whether it should be changed (it is in the "low level" section after all so maybe that's intentional?). If the test is important than I can add some logic to first split off the query string and escape just the path part but I want to make sure that's necessary.

I think the going this path of splitting the query off will be needed, because a lot of other S3 endpoints are using some form of querying, as one example;

https://github.com/JuliaCloud/AWS.jl/blob/b4ce295505e240d17111d6af6ced84676d3b454a/src/services/s3.jl#L404

mattBrzezinski:
This is definitely something maybe overlooked or unforeseen when creating this package, but it all comes back to these special characters which AWS allows for and supporting them in this package. It's a bit of a trade off between having these high-level functions, not over complicating the package in breaking them down into a lower level, and allowing for a smooth transition from using `AWSCore.jl -> AWS.jl`.

While AWS does allow for things such as `$, &, +`, I again would push back a bit on using them in your filenames. As we've seen in #293 sometimes it's not up to the end user and someone further upstream is following this format, but when possible I would recommend only using safe characters as defined by AWS.

I don't have a problem adding into this list of characters to clean a `'+' => "%2B"`, as we've already set down this path, it's the most simple solution, and unblocks your work. The proper solution to cleanly support these extra characters involves quite a bit of work however.

mattBrzezinski:
> Okay. So sounds like may be better to create a new PR that's more narrowly scoped that just adds `+` escape to the existing escaping code?

Probably easier than re-writing it all in this branch / PR! I'll keep my eye open for it and get it through for you ASAP.

kleinschmidt:
One of the frustrating things here is that AWS is pretty vague on what "safe characters" are.  There are at least three relevant categories described by the docs: there are "definitely safe" characters (`!'()*-./_a-zA-Z0-9`), "may require special handling... likely need to be URL encoded or referenced as HEX." (` \$&+,:;=?@`), and "characters to avoid" (`\{}^%`[]<>~#|`).  Currently, we escape some (but not all) of the non alphanumeric "definitely safe" (`!'()*`) and some (but not all) of the "may require special handling" (` =`).

kleinschmidt:
And, again, the AWS cli is automatically escaping these characters too.  I have no idea whether that's specific to the S3 subcommand or general though.

mattBrzezinski:
> One of the frustrating things here is that AWS is pretty vague on what "safe characters" are. There are at least three relevant categories described by the docs: there are "definitely safe" characters (`!'()*-./_a-zA-Z0-9`), "may require special handling... likely need to be URL encoded or referenced as HEX." (` $&+,:;=?@`), and "characters to avoid" (`` \{}^%``[]<>~#| ``). Currently, we escape some (but not all) of the non alphanumeric "definitely safe" (`!'()*`) and some (but not all) of the "may require special handling" (` =`).

For certain characters we do *not* need to be explicit in escaping them. Referring to [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html) webpage. This package full supports all characters in the safe section as well as spaces and equal signs.

We *should* be able to safely support all characters in the ones which require special handling, however I would want to take this on a case-by-case basis as they come up as some maybe more problematic than others.

```julia
S3.put_object("mattbr-test-bucket", "0aA/!-_.*'().txt", Dict("body"=>"hello world!"))
UInt8[]
```

mattBrzezinski:
Closing in favour of #299

PR #317, opened 2021-03-16
S3: really call version 2 in `list_objects_v2`

### LINKED ISSUE #316: S3.list_objects_v2 doesn't contain a continuation token
To paginate buckets I wanted to make use of the `ContinuationToken` but `NextContinuationToken` it isn't contained in the response from S3 although 
`"IsTruncated"="true"`
is returned.

According to https://docs.aws.amazon.com/cli/latest/reference/s3api/list-objects-v2.html 

> NextContinuationToken -> (string)
> NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed. The next list requests to Amazon S3 can be continued with this NextContinuationToken . NextContinuationToken is obfuscated and is not a real key

Am I missing something obvious?

laborg:
Fixes #316 by only using S3 escape rules for the path.

- With this I can properly use the continuation token for paginated results.
- I don't know enough about minio to know if the added tests are suitable to detect a regression. 
- URIs.jl has been added as an dependency, but I feel that this will sooner or later be the case anyway.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -11,6 +11,7 @@ using Retry
 using Sockets
 using UUIDs: UUIDs
 using XMLDict
+import URIs
 
 export @service
 export _merge, AbstractAWSConfig, AWSConfig, AWSExceptions, AWSServices, Request
@@ -576,7 +577,9 @@ function (service::RestXMLService)(
     return_headers = _return_headers(args)
 
     if request.service == "s3"
-        request_uri = _clean_s3_uri(request_uri)
+        u = URIs.URI(request_uri)
+        cleaned_uri = URIs.URI(u;path=_clean_s3_uri(u.path))
+        request_uri = string(cleaned_uri)
     end
     request.resource = _generate_rest_resource(request_uri, args)
     query_str = HTTP.escapeuri(args)
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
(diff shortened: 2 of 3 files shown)


mattBrzezinski:
bors try

laborg:
bors try

mattBrzezinski:
bors try

src/AWS.jl
```diff
@@ -576,7 +577,9 @@ function (service::RestXMLService)(
     return_headers = _return_headers(args)
 
     if request.service == "s3"
-        request_uri = _clean_s3_uri(request_uri)
+        u = URIs.URI(request_uri)
```
mattBrzezinski, line 580: Can we put this functionality into `_clean_s3_ur()` then add in another test case where the passed in `request_uri` is something like:

> "/test-bucket/*)=('! +.txt?list-objects=v2"

I would expect (I've already tested this) to return back:

> "/test-bucket/%2A%29%3D%28%27%21%20%2B.txt?list-objects=v2"

mattBrzezinski (commented):
LGTM, just a few minor things, thank you for making this PR!

src/AWS.jl
laborg: I had to move `_clean_s3_uri` to the top level to make it callable in the test set, hope that is okay?

mattBrzezinski: Yep that's fine!

test/AWS.jl
```diff
@@ -457,6 +457,23 @@ end
...
+    @test AWS._clean_s3_uri(uri) == expected_uri
+    @test URIs.unescapeuri(expected_uri) == uri
+
+    # make sure that other parts of the uri aren't changed by `_clean_s3_uri`
+    for uri in ("https://julialang.org",
+            "http://julialang.org",
+            "http://julialang.org:8080",
+            "/onlypath",
+            "/path?query=  +99",
+            "/anchor?query=yes#anchor1")
```
mattBrzezinski, line 472:
```suggestion
    for uri in (
            "https://julialang.org",
            "http://julialang.org",
            "http://julialang.org:8080",
            "/onlypath",
            "/path?query=  +99",
            "/anchor?query=yes#anchor1"
    )
```

mattBrzezinski (commented):
Two minor changes and it's good to go!

laborg:
Thanks for steering this PR. I really like your work on AWS.jl.

mattBrzezinski (approved):

PR #321, opened 2021-03-18
fix streaming interface for S3

laborg:
I had problems when trying to stream the data of large files in S3 into local files and discovered a bug as well as an suboptimal implementation.

- `response_stream` expects a `Base.BufferStream` but within `get_objects` this is then turned into a query part, and of course this doesn't work well.
- Most of the Julia IO stuff excepts <:IO because this can easily be provided with `open()`. Having a more general type above `Base.BufferStream` allows for such flexibility.


src/AWS.jl
```diff
@@ -575,8 +575,10 @@ function (service::RestXMLService)(
         request.response_dict_type = pop!(args, "response_dict_type")
     end
 
-    delete!(args, "headers")
-    delete!(args, "body")
+    for key in ("headers", "body", "response_stream", "return_stream", "return_raw")
+        delete!(args, key)
+    end
```
mattBrzezinski, line 580: Looking at this some more, we can do one even better. When making the Request object above and setting these parameters we can just do this for each of the keys:

> pop!(args, "headers", [])

test/minio.jl
```diff
@@ -32,6 +32,17 @@ S3.put_object("anewbucket","foo/baz",Dict("body"=>"a secondnested object"))
...
+# Fix: BufferStream can be generalized to IO
+# v1.29.1: ERROR: MethodError: Cannot `convert` an object of type IOStream to an object of type Base.BufferStream
+mktemp() do f, io
+    S3.get_object("anewbucket","myobject", Dict("response_stream"=>io))
+    @test read(f, String) == "Hi from Minio"
+end
+
+# Bug: `response_stream` is pushed into the query part...
+# v1.29.1: ERROR: MethodError: no method matching iterate(::Base.BufferStream)
+S3.get_object("anewbucket","myobject", Dict("response_stream"=>Base.BufferStream(), "return_stream"=>true))
```
mattBrzezinski, line 44:
```suggestion
S3.get_object("anewbucket", "myobject", Dict("response_stream"=>Base.BufferStream(), "return_stream"=>true))
```

test/minio.jl
```diff
@@ -32,6 +32,17 @@ S3.put_object("anewbucket","foo/baz",Dict("body"=>"a secondnested object"))
 # Test retrieving an object
 @test S3.get_object("anewbucket","myobject") |> String == "Hi from Minio"
 
+# Fix: BufferStream can be generalized to IO
```
mattBrzezinski, line 35: These can probably get moved into the `test/AWS.jl` file as they're general tests and not Min.IO specific. Although with that said some general tests need to be in place for Min.IO to check it's functionality works too.

src/AWS.jl
laborg: Yes, that looks better.

test/minio.jl
laborg: I've reshuffled this a bit: 
 - A "streaming" test is now in AWS.jl and minio.jl, as both should handle this case properly.
 - A regression test in issues.jl concerning the problem where the `response_stream` was pushed in to the query.

test/issues.jl
```diff
@@ -49,3 +49,24 @@ end
         @test_throws NoCredentials Lambda.list_functions(;aws_config=AWSConfig(creds=nothing))
     end
 end
+
+@testset "issue streaming" begin
```
mattBrzezinski, line 53: Would you mind cutting an issue explaining the problem then titling this test set `issue XYZ`?

src/AWS.jl
```diff
@@ -563,20 +563,15 @@ function (service::RestXMLService)(
         service=service.name,
         api_version=service.api_version,
         request_method=request_method,
-        content=get(args, "body", ""),
-        headers=LittleDict{String, String}(get(args, "headers", [])),
-        return_stream=get(args, "return_stream", false),
-        http_options=get(args, "http_options", []),
-        response_stream=get(args, "response_stream", nothing),
-        return_raw=get(args, "return_raw", false),
+        content=pop!(args, "body", ""),
```
mattBrzezinski, line 566: A bit out of scope, but do you mind changing the RESTJSON, JSON and Query functors as well to this `pop!` style? I think they should be consistent throughout.

laborg: I've refactored the code a little to get rid of some duplications.

src/AWS.jl
```diff
@@ -465,15 +477,7 @@ function _generate_rest_resource(request_uri::String, args::AbstractDict{String,
...
-function _return_headers(args::AbstractDict{String, <:Any})
-    return_headers = get(args, "return_headers", false)
-
-    if return_headers
-        delete!(args, "return_headers")
-    end
-
-    return return_headers
-end
+_return_headers(args::AbstractDict{String, <:Any}) = pop!(args, "return_headers", false)
```
mattBrzezinski, line 480: We can just scrap this function and call `pop!` where applicable

src/AWS.jl
mattBrzezinski: Awesome thank you!

mattBrzezinski (commented):
bors try

test/issues.jl
```diff
@@ -49,3 +49,24 @@ end
         @test_throws NoCredentials Lambda.list_functions(;aws_config=AWSConfig(creds=nothing))
     end
 end
+
+@testset "issue 324" begin
+    body = "Hello World!"
+    file_name = "streaming.bin"
+
+    try
+        S3.put_object(bucket_name, file_name, Dict("body" => body))
```
mattBrzezinski, line 58: Ah looks like when I made this I created and deleted the bucket after every test.
Probably to ensure that no artifacts get stuck and left over.

You'll just need to preface this line with:

```suggestion
        S3.create_bucket(bucket_name)
        S3.put_object(bucket_name, file_name, Dict("body" => body))
```

laborg: done

mattBrzezinski:
bors try

src/AWS.jl
laborg: I removed the `_return_headers` test, because it was basically just testing the three arg `pop!()` call method.

mattBrzezinski:
Ugh... LittleDicts do not implement `pop!` with a default value. I'll make these changes in `OrderedCollections`, feel free to revert that change and I can get around to putting it back in.

mattBrzezinski:
Oh it already looks like I was trying to do this in the past...

> https://github.com/JuliaCollections/OrderedCollections.jl/pull/59

laborg:
Do you see a problem with the proposed `_pop!` as a fall back for now?

src/AWS.jl
```diff
@@ -722,32 +703,21 @@ function (service::RestJSONService)(
...
     aws_config::AbstractAWSConfig=global_aws_config(),
 )
-    return_headers = _return_headers(args)
+    return_headers = _pop!(args, "return_headers", false)
 
-    request = Request(
-        service=service.name,
-        api_version=service.api_version,
+    request = Request(;
+        _extract_common_kw_args(service,args)...,
```
mattBrzezinski, line 709:
```suggestion
        _extract_common_kw_args(service, args)...,
```

src/AWS.jl
```diff
@@ -162,12 +162,36 @@ Base.@kwdef mutable struct Request
...
     response_dict_type::Type{<:AbstractDict}=LittleDict
 end
 
+# use this until the threed arg pop! is available for LittleDict
+# https://github.com/JuliaCollections/OrderedCollections.jl/pull/59
+function _pop!(dict::AbstractDict{String, <: Any},kw,default)
+    if haskey(dict,kw)
+        val = dict[kw]
+        delete!(dict,kw)
+        return val
```
mattBrzezinski, line 177:
```suggestion

        return val
```

mattBrzezinski:
bors try

mattBrzezinski (commented):
Two minor white spacing things.

mattBrzezinski:
> Do you see a problem with the proposed `_pop!` as a fall back for now?

No it makes sense, I don't see this being problematic.

mattBrzezinski:
bors try

mattBrzezinski:
bors try

laborg:
Anything else left to do?

mattBrzezinski (approved):
bors r+

Issue #335, opened 2021-04-06
DNSError: Unknown node or service


mattBrzezinski:
Beginning to see this crop up with a few services, using `AWS@1.33.1`;

```julia
using AWS
@service Detective

Detective.list_graphs()
ERROR: DNSError: detective.us-east-1.amazonaws.com, unknown node or service (EAI_NONAME)
```

Most likely need to check if this is a regionless service, or what the URL is expecting here.

phelipe:
Is this issue solved? I still have this problem in IOT services.

```julia
Iot.describe_endpoint()
ERROR: DNSError: execute-api.us-east-1.amazonaws.com, unknown node or service (EAI_NONAME)
Stacktrace:
  [1] getalladdrinfo(host::String)
    @ Sockets /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:113
  [2] getalladdrinfo
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Sockets/src/addrinfo.jl:122 [inlined]
  [3] getconnection(::Type{Sockets.TCPSocket}, host::SubString{String}, port::String; keepalive::Bool, connect_timeout::Int64, readtimeout::Int64, kw::Base.Iterators.Pairs{Symbol, Union{Nothing, Bool}, Tuple{Symbol, Symbol, Symbol}, NamedTuple{(:require_ssl_verification, :iofunction, :response_stream), Tuple{Bool, Nothing, Nothing}}})
    @ HTTP.ConnectionPool ~/.julia/packages/HTTP/cxgat/src/ConnectionPool.jl:657
  [4] #getconnection#29
    @ ~/.julia/packages/HTTP/cxgat/src/ConnectionPool.jl:726 [inlined]
  [5] newconnection(pod::HTTP.ConnectionPool.Pod, T::Type, host::SubString{String}, port::SubString{String}, pipeline_limit::Int64, require_ssl_verification::Bool, idle_timeout::Int64; kw::Base.Iterators.Pairs{Symbol, Nothing, Tuple{Symbol, Symbol}, NamedTuple{(:iofunction, :response_stream), Tuple{Nothing, Nothing}}})
    @ HTTP.ConnectionPool ~/.julia/packages/HTTP/cxgat/src/ConnectionPool.jl:624
[...]

PR #336, opened 2021-04-06
Use signingName when possible

mattBrzezinski:
### Overview

We were always using the `service["endpointPrefix"]` for the URL when making requests to AWS. This works in most cases because the `endpointPrefix` matches what is expected in the request. However in certain situations this value was not what AWS expected for certain services such as SageMaker.

This PR changes functionality by seeing if `signingName` is available (not in all service definitions), if so use that otherwise default back to the `endpointPrefix`.

I've tested the new URL endpoints for SageMaker, ECR and they are working properly now. There is another issue coming up now, see #335


src/AWSMetadataUtilities.jl
```diff
@@ -136,7 +136,7 @@ Get the low-level definition for an AWS Service.
 """
 function _generate_low_level_definition(service::AbstractDict)
     protocol = service["protocol"]
-    service_name = service["endpointPrefix"]
+    service_name = haskey(service, "signingName") ? service["signingName"] : service["endpointPrefix"]
```
mjram0s, line 139: Should add a test

mattBrzezinski: Done!

mjram0s (approved):

PR #342, opened 2021-04-15
isa Any

mattBrzezinski:
This has apparently always been incorrect, we want to be using `Dict{String, <:Any}` not `Dict{String, Any}`

- https://github.com/JuliaCloud/AWS.jl/blob/v1.0.0/src/AWSMetadataUtilities.jl#L493


iamed2:
How about just `AbstractDict{String}`? Equivalent and shorter.

mattBrzezinski:
bors try

mjram0s (approved):

Issue #224, opened 2020-10-17
S3: force get_object to return Vector{UInt8}?


christopher-dG:
The return type of `get_object` seems to depend on what type of file you're looking for. In the case of strings vs bytes it's not so bad, but IMO getting a Dict or Vector for a JSON file is not great, since you lose all formatting (and who says I want it parsed anyways?). Is this behaviour intended? Is there a way to always get a byte vector?

```jl
julia> using AWS: @service

julia> @service S3;

julia> S3.get_object("cdg-scratch", "x.txt")
 "hi"

julia> S3.get_object("cdg-scratch", "x.png")
 1245160-element Array{UInt8,1}:
 ⋮

julia> S3.get_object("cdg-scratch", "x.json")  # {}
 OrderedCollections.LittleDict{String,Any,Array{String,1},Array{Any,1}}()

julia> S3.get_object("cdg-scratch", "y.json")  # []
 Any[]
```

mattBrzezinski:
This functions the same way as `AWSCore.jl` does, you do have some control over how the response is returned;

https://github.com/JuliaCloud/AWS.jl/blob/6dcae89cc64f93d8135c7c50a98771bd7b70d535/src/AWS.jl#L166

Maybe we change the `response_dict_type::Type{<:AbstractDict}` to `response_type::Type` and then give users free reign over how they want things returned, defaulting to what already exists?

christopher-dG:
IMO the "right" default is to return the raw file contents, but it's less disruptive to default to the existing method. But I think changing `response_dict_type` -> `response_type` is a good move!

iamed2:
Boto3 returns a dict with metadata and the readable stream under the 'Body' key. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object

What's the source of the difference here?

iamed2:
Found the difference: https://github.com/boto/botocore/blob/develop/botocore/data/s3/2006-03-01/service-2.json#L3860

Boto3 respects the input shape and output shape specifications, while AWS.jl applies a heuristic to determine how to return data. In the long term, we can also use the shape specifications. In the short term, I think AWSS3.jl should be made able to provide the interface that you want, and AWS.jl should just need to be configurable to facilitate that.

c42f:
Yeah, I really don't think AWS.jl is the right place to have fancy heuristics for the return type. A higher layer should deal with that.

[...]

PR #346, opened 2021-04-22
RFC: Return Raw more explicitly

mattBrzezinski:
## Overview
In #224 it was brought up that the default behavior for making a request should be to return a `Vector{UInt8}` instead of applying heuristics to parse responses back. Applying heuristics makes viewing responses a lot more simple and stops users from needing to parse themselves, however you do end up losing the format of the file.

One of the tenets when creating this iteration of the package was to allow for the easiest interfacing with AWS. Currently you can get raw responses back (as `Vector{UInt8}`), but you need to pass in `return_raw=true` with every request you make. This does not make sense in a large scale application to constantly pass it along.

## Proposed Changes
I think that these changes are a good compromise between the two. It allows for users to set a globally set a variable `set_return_raw(true)` to always return back raw responses. I've also added in functionality upon loading the package to check for an environment variable `AWS_JL_RAW` and set this global variable based off of it. The environment variables main purpose is to quickly use raw responses with EC2/ECS instances.

These changes are backwards compatible, as I do not want to tag `AWS@2` yet, so users can continue to use `Request.return_raw`. However by accessing this property or attempting to manually set it a warning will be output warning that this functionality will be removed in the next major version.

## Other Changes
As I was working on this I came across a couple minor things:

- Moved the `struct Request` into its own file, as `src/AWS.jl` is getting large
- Low-level service functions kwarg `aws` did not match that of high-level functions `aws_config`
- `const` globally to UPPERCASE

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,7 @@ Mocking = "78c3b35d-d492-501b-9361-3d52fe80e533"
 OrderedCollections = "bac558e1-5e72-5ebc-8fee-abe8a469f55d"
 Retry = "20febd7b-183b-5ae2-ac4a-720e7ce64774"
 Sockets = "6462fe0b-24de-5631-8697-dd941f90decc"
+Suppressor = "fd094767-a336-5f1f-9728-57cf17d0bbfb"
 URIs = "5c2747f8-b7ea-4ff2-ba2e-563bfd36b1d4"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
@@ -30,6 +31,7 @@ MbedTLS = "0.6, 0.7, 1"
 Mocking = "0.7"
 OrderedCollections = "1"
 Retry = "0.3, 0.4"
+Suppressor = "0.2"
 URIs = "1"
 XMLDict = "0.3, 0.4"
 julia = "1"
@@ -42,4 +44,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [targets]
-test = ["Pkg", "Random", "Suppressor", "Test", "UUIDs"]
+test = ["Pkg", "Random", "Test", "UUIDs"]
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -9,16 +9,19 @@ using Mocking
 using OrderedCollections: LittleDict, OrderedDict
 using Retry
 using Sockets
+using Suppressor
 using UUIDs: UUIDs
 using XMLDict
 import URIs
 
 export @service
 export _merge, AbstractAWSConfig, AWSConfig, AWSExceptions, AWSServices, Request
-export global_aws_config, generate_service_url, set_user_agent, sign!, sign_aws2!, sign_aws4!
+export global_aws_config, set_user_agent, set_return_raw
+export generate_service_url, sign!, sign_aws2!, sign_aws4!
 export JSONService, RestJSONService, RestXMLService, QueryService
 
 include("utilities.jl")
+include("request.jl")
 include("AWSExceptions.jl")
 include("AWSCredentials.jl")
 include("AWSConfig.jl")
@@ -28,27 +31,32 @@ include("AWSMetadata.jl")
 using ..AWSExceptions
 using ..AWSExceptions: AWSException
 
-const user_agent = Ref("AWS.jl/1.0.0")
-const aws_config = Ref{AbstractAWSConfig}()
+const ENV_RETURN_RAW = "AWS_JL_RAW"
+
+const USER_AGENT = Ref("AWS.jl/1.0.0")
+const AWS_CONFIG = Ref{AbstractAWSConfig}()
+const RETURN_RAW = Ref{Bool}(false)
+
+
+function __init__()
+    if haskey(ENV, ENV_RETURN_RAW)
+        set_return_raw(ENV[ENV_RETURN_RAW])
+    end
+end
+
 
 """
     global_aws_config()
 
 Retrieve the global AWS configuration.
 If one is not set, create one with default configuration options.
-
-# Keywords
-- `kwargs...`: AWSConfig kwargs to be passed along if the global configuration is not already set
-
-# Returns
-- `AWSConfig`: The global AWS configuration
 """
 function global_aws_config(; kwargs...)
-    if !isassigned(aws_config) || !isempty(kwargs)
-        aws_config[] = AWSConfig(; kwargs...)
+    if !isassigned(AWS_CONFIG) || !isempty(kwargs)
+        AWS_CONFIG[] = AWSConfig(; kwargs...)
     end
 
-    return aws_config[]
+    return AWS_CONFIG[]
 end
 
 
@@ -56,15 +64,9 @@ end
     global_aws_config(config::AbstractAWSConfig)
 
 Set the global AWSConfig.
-
-# Arguments
-- `config::AWSConfig`: The AWSConfig to set in the global state
-
-# Returns
-- `AWSConfig`: Global AWSConfig
 """
 function global_aws_config(config::AbstractAWSConfig)
-    return aws_config[] = config
+    return AWS_CONFIG[] = config
 end
 
 
@@ -72,14 +74,17 @@ end
     set_user_agent(new_user_agent::String)
 
 Set the global user agent when making HTTP requests.
-
-# Arguments
-- `new_user_agent::String`: User agent to set when making HTTP requests
-
-# Return
-- `String`: The global user agent
 """
-set_user_agent(new_user_agent::String) = return user_agent[] = new_user_agent
+set_user_agent(new_user_agent::String) = return USER_AGENT[] = new_user_agent
+
+
+
+"""
+    set_return_raw(b::Bool)
+
+Return raw requests from AWS.
+"""
+set_return_raw(b::Bool) = return RETURN_RAW[] = b
 
 
 """
@@ -151,26 +156,9 @@ end
 
 RestJSONService(name::String, api_version::String) = RestJSONService(name, api_version, LittleDict{String, String}())
 
-Base.@kwdef mutable struct Request
-    service::String
-    api_version::String
-    request_method::String
-
-    headers::AbstractDict{String, String}=LittleDict{String, String}()
-    content::Union{String, Vector{UInt8}}=""
-    resource::String=""
-    url::String=""
-
-    return_stream::Bool=false
-    response_stream::Union{<:IO, Nothing}=nothing
-    http_options::Array{String}=[]
-    return_raw::Bool=false
-    response_dict_type::Type{<:AbstractDict}=LittleDict
-end
-
 # use this until the threed arg pop! is available for LittleDict
 # https://github.com/JuliaCollections/OrderedCollections.jl/pull/59
-function _pop!(dict::AbstractDict{String, <: Any},kw,default)
+function _pop!(dict::AbstractDict{String, <: Any}, kw, default)
     if haskey(dict,kw)
         val = dict[kw]
         delete!(dict,kw)
@@ -389,7 +377,7 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
         "PriorRequestNotComplete"
     ]
 
-    request.headers["User-Agent"] = user_agent[]
+    request.headers["User-Agent"] = USER_AGENT[]
     request.headers["Host"] = HTTP.URI(request.url).host
 
     @repeat 3 try
@@ -445,8 +433,10 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
     end
 
     # Return raw data if requested...
-    if request.return_raw
-        return (return_headers ? (response.body, response.headers) : response.body)
+    @suppress begin
+        if RETURN_RAW[] || request.return_raw
+            return (return_headers ? (response.body, response.headers) : response.body)
+        end
     end
 
     # Parse response data according to mimetype...
@@ -554,7 +544,7 @@ end
 """
     (service::RestXMLService)(
         request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, String}();
-        aws::AbstractAWSConfig=aws_config
+        aws_config::AbstractAWSConfig=global_aws_config()
     )
 
 Perform a RestXML request to AWS.
@@ -565,7 +555,7 @@ Perform a RestXML request to AWS.
 - `args::AbstractDict{String, <:Any}`: Additional arguments to be included in the request
 
 # Keywords
-- `aws::AbstractAWSConfig`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
+- `aws_config::AbstractAWSConfig=global_aws_config()`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
 
 # Returns
 - `Tuple or Dict`: If `return_headers` is passed in through `args` a Tuple containing the Headers and Response will be returned, otherwise just a Dict
@@ -605,7 +595,7 @@ end
 """
     (service::QueryService)(
         operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}();
-        aws::AbstractAWSConfig=aws_config
+        aws_config::AbstractAWSConfig=global_aws_config()
     )
 
 Perform a Query request to AWS.
@@ -615,7 +605,7 @@ Perform a Query request to AWS.
 - `args::AbstractDict{String, <:Any}`: Additional arguments to be included in the request
 
 # Keywords
-- `aws::AbstractAWSConfig`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
+- `aws_config::AbstractAWSConfig=glboal_aws_config()`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
 
 # Returns
 - `Tuple or Dict`: If `return_headers` is passed in through `args` a Tuple containing the Headers and Response will be returned, otherwise just a Dict
@@ -646,7 +636,7 @@ end
 """
     (service::JSONService)(
         operation::String, args::AbstractDict{String, <:Any}=Dict{String, Any}();
-        aws::AbstractAWSConfig=aws_config
+        aws_config::AbstractAWSConfig=global_aws_config()
     )
 
 Perform a JSON request to AWS.
@@ -656,7 +646,7 @@ Perform a JSON request to AWS.
 - `args::AbstractDict{String, <:Any}`: Additional arguments to be included in the request
 
 # Keywords
-- `aws::AbstractAWSConfig`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
+- `aws_config::AbstractAWSConfig=global_aws_config()`: AWSConfig containing credentials and other information for fulfilling the request, default value is the global configuration
 
 # Returns
 - `Tuple or Dict`: If `return_headers` is passed in through `args` a Tuple containing the Headers and Response will be returned, otherwise just a Dict
@@ -685,7 +675,7 @@ end
 """
     (service::RestJSONService)(
         request_method::String, request_uri::String, args::AbstractDict{String, <:Any}=Dict{String, String}();
-        aws::AbstractAWSConfig=aws_config
+        aws_config::AbstractAWSConfig=global_aws_config()
     )
 
 Perform a RestJSON request to AWS.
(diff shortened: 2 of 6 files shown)


mattBrzezinski:
cc: @christopher-dG, @iamed2, @c42f

oxinabox:
Won't this break if I have two packages my main project uses, one of which sets return raw false and the other true?

Also doesn't this make every thing type unstable? Which isn't in this case performance critical.


Changing return type based on a global variable is weird.
Am I misunderstanding this PR entirely?

mattBrzezinski:
> Won't this break if I have two packages my main project uses, one of which sets return raw false and the other true?

If you're using `AWS.jl` and `AWSS3.jl` and you explicitly set `ENV["AWS_JL_RAW"]=true` everything will return raw. Otherwise you would default to current functionality, but if you wanted some to be raw you would then need to pass along `return_raw=true` to the requests which you want raw.

> Also doesn't this make every thing type unstable? Which isn't in this case performance critical.

Currently your responses come back as `LittleDict` (or any `AbstractDict` you set) by default
OR `Vector{UInt8}` if you pass in `return_raw=true`
OR a stream if you pass in `response_stream=IOStream`

> Changing return type based on a global variable is weird.
> Am I misunderstanding this PR entirely?

There seems to be a community of people who want things returned raw. Personally I don't thing this is be best default behavior as I think this package should be convenience first. I don't think that users should need to convert responses to Strings, then parse based off of request types into either XML or JSON parsers every time.

However, I also don't think it's reasonable for those who want raw returns to need to pass in `return_raw=true` to every single request that they are making.

> Can you split the other changes into a seperate PRs and then rebase this so it is easier to review?

Yes, I can do that tomorrow.

ericphanson:
I'm also not really sold on an environmental variable / global config determining the response type; it seems confusing/surprising/unusual and easy to get out of sync (between the env variable and the internal AWS state), and like it could be the source of confusing subtle bugs. But I also really strongly feel that one should be getting raw bytes always, unless maybe they are using a very high level convenience function (and that it's even worth just cutting an AWS v2).

> I don't think that users should need to convert responses to Strings, then parse based off of request types into either XML or JSON parsers every time.

Right, converting to a string is kind of annoying and wasteful-- but JSON parsers like JSON3 accept raw byte buffers for a reason! (As does CSV.jl, Arrow.jl, etc)

> Personally I don't thing this is be best default behavior as I think this package should be convenience first. 

I think this must depend a lot on how you are using the package. For me, the heuristics are very inconvenient since I will say do `CSV.File(read(S3Path(...))))` to read a CSV into memory. Then I will look at a JSON file and call `JSON3.read(read(S3Path(...)))` and then get an error about `JSON3.read` not understanding a `LittleDict`. I also get errors like `ERROR: MethodError: no method matching Vector{UInt8}(::OrderedCollections.LittleDict{Union{String, Symbol}, Any, Vector{Union{String, Symbol}}, Vector{Any}})` coming from within AWS/AWSS3 where it seems some internal function assumed it would get raw bytes. It's also inconsistent; `read` on local paths always gives a byte vector, whereas on S3Paths it depends on what the file extension is.

I think most users are not sending off single requests and manually inspecting the result. To me it feels like that's something useful when building/debugging a package like this, but not very common for general use-cases. But I also don't really know how other folks outside of Beacon use AWS.jl.

Personally, I really think functions like `s3_get` should serve as a stable primitive you can build on, and they aren't that if they are sometimes returning dicts and sometimes returning bytes.

mattBrzezinski:
> I'm also not really sold on an environmental variable / global config determining the response type; it seems confusing/surprising/unusual and easy to get out of sync (between the env variable and the internal AWS state), and like it could be the source of confusing subtle bugs. But I also really strongly feel that one should be getting raw bytes always, unless maybe they are using a very high level convenience function (and that it's even worth just cutting an AWS v2).
> 
> > I don't think that users should need to convert responses to Strings, then parse based off of request types into either XML or JSON parsers every time.
> 
> Right, converting to a string is kind of annoying and wasteful-- but JSON parsers like JSON3 accept raw byte buffers for a reason! (As does CSV.jl, Arrow.jl, etc)
> 
> > Personally I don't thing this is be best default behavior as I think this package should be convenience first.
> 
> I think this must depend a lot on how you are using the package. For me, the heuristics are very inconvenient since I will say do `CSV.File(read(S3Path(...))))` to read a CSV into memory. Then I will look at a JSON file and call `JSON3.read(read(S3Path(...)))` and then get an error about `JSON3.read` not understanding a `LittleDict`. I also get errors like `ERROR: MethodError: no method matching Vector{UInt8}(::OrderedCollections.LittleDict{Union{String, Symbol}, Any, Vector{Union{String, Symbol}}, Vector{Any}})` coming from within AWS/AWSS3 where it seems some internal function assumed it would get raw bytes. It's also inconsistent; `read` on local paths always gives a byte vector, whereas on S3Paths it depends on what the file extension is.
> 
> I think most users are not sending off single requests and manually inspecting the result. To me it feels like that's something useful when building/debugging a package like this, but not very common for general use-cases. But I also don't really know how other folks outside of Beacon use AWS.jl.
> 
> Personally, I really think functions like `s3_get` should serve as a stable primitive you can build on, and they aren't that if they are sometimes returning dicts and sometimes returning bytes.
> 
> edit: @christopher-dG pointed out to me that the raw byte stuff is probably mostly important for s3 getobject, not various other API calls. I think the important point is basically type stability: the api function you choose (and possibly the type of the input) should determine the output type. So if it's a function that can always return a dict, then that's fine, but s3 getobject can return anything so the common unsurprising denominator is a vector of bytes.

Another alternative thinking about this is to possibly move the heuristics up to only the high-level functions. That way there can be true convenience functions. I feel like if anything most of this discussion revolves around S3 its self and less of other services. Not sure if @omus has any thoughts on this, I feel like he has a good eye for these things.

c42f:
> Won't this break if I have two packages my main project uses, one of which sets return raw false and the other true?

This is exactly my concern. A global setting which modifies the return type into something with a *different abstract interface* makes the entire API unreliable for use in reusable libraries, as those libraries can no longer be composed together.

To avoid this problem it's critical that such a setting is *lexically scoped* to the module. Either explicitly in `return_raw`, or perhaps at the module level (could contextural configuration like this be folded into the expansion of `@service`?)

> Another alternative thinking about this is to possibly move the heuristics up to only the high-level functions. That way there can be true convenience functions. I feel like if anything most of this discussion revolves around S3 its self and less of other services.

This sounds right to me; both that there should be a low-level interface where users can opt out of the parsing machinery, and that this behavior is especially unintuitive for S3 moreso than other services. Many of the AWS services always provide structured data in a certain format according to their APIs and internal data model of the service itself. So in those cases libraries can rely on the interface of the data which is returned and parsed according to content type to always be consistent.

Unfortunately, for S3 blobs we've got uploader tools defaulting to guessing the mime type during upload based on unreliable heuristics. In particular, I only just learned that `awscli` tries to guess it and needs `--no-guess-mime-type` to disable that, which might arguably be the root cause of my surprise here. Even if this is arguably a bad default, I'm sure it's well entrenched in uploader tools at this point.

Personally I'd be in favor of:
* Having an official low-level API where raw content is returned.
* Having a high-level API where structured data is returned, as currently
* Potentially disabling this (by default) specifically for `S3.get_object`, as a workaround for file uploaders with bad defaults :-(
* As an extension, allow the way that requests get parsed in the high-level API to be customized.  Either with an explicitly passed context of some kind, or by having this context set per-module as part of the `@service` expansion.

ericphanson:
@c42f, when you write

> Potentially disabling this (by default) specifically for S3.get_object

do you mean keep the heuristics return for `get_object`, or disable them?

mattBrzezinski:
I personally don't like the idea of disabling this for just S3, I think that might create confusion down the line and special casing services is something in general that I want to avoid where possible.

Based off of the feedback I'm currently thinking of making these changes:

Low-Level requests will always return back raw responses, users if they choose to use these will need to know *how* to perform the requests by knowing the REST request type, and endpoint. You can find these by looking at the high-level code wrappers, [example](https://github.com/JuliaCloud/AWS.jl/blob/d9b5a5ef703d72dba256a6651be53b69c7fe2dd5/src/services/s3.jl#L1728). 
```julia
using AWS.AWSServices : s3
bucket = "my-bucket"
file = "my-file"

s3("GET", "/$(Bucket)/$(Key)"; aws_config=aws_config)
```

High-level functions will function as they currently do applying heuristics to the responses, *AND* if you pass in `return_raw=true` will give back a raw response still allowing to use these convenience functions and getting raw responses back.

```julia
using AWS
@service S3

S3.get_object("my-bucket", "my-file")
```  

Any thoughts? Disagreements?

iamed2:
I'm highly against a global or environment variable, for the reasons covered above.

It would make sense to me to have a response type that can be used to construct many different structures, like Python's `requests.Response` which has `raw`, `json`, and `text` properties. This would just delay the parsing that's already happening. 

I don't think any low level code should return the raw bytes of the body. You're throwing away header and type information to save one field lookup or function call.

mattBrzezinski:
I think the actual solution here it to create response objects based off of the JS API definitions, see [this](https://github.com/aws/aws-sdk-js/blob/c59435fe2c8391a39cbc9cd896db5d5e4bb7e82d/apis/s3-2006-03-01.normal.json#L4574-L4769).

omus:
> It would make sense to me to have a response type that can be used to construct many different structures, like Python's `requests.Response` which has `raw`, `json`, and `text` properties. This would just delay the parsing that's already happening.

I agree this seems to be the right approach.

c42f:
> do you mean keep the heuristics return for get_object, or disable them?

I meant to disable them for `GetObject` as a concession to questionable heuristics in other tools. This isn't at all satisfying though. Hopefully we can do better.

> I think the actual solution here it to create response objects based off of the JS API definitions

This sounds pretty useful. Is the idea that
* For every `$operation`, you'd look at `out_shape = operations.$operation.output.shape`, then look up `shapes.$out_shape`, and that tells you how to generate members (in a struct or dict-like thing) for the response of that operation
* The response stores a raw body internally
* Provide some convenient way(s) to get at the `payload` (which might come from the body, but might be from a header). Possibly as properties of the response object as suggested

What does the JS API do with the output shape?

mattBrzezinski:
I've spent some time playing around with various AWS services and their requests, and it seems that the `S3.get_object()` request is special in how it returns back data. All other services return back some form of structured data about the request. For example:

### JSON Service
```julia
using AWS
@service ECR

args = Dict{String, Any}("return_raw"=>true)
r = ECR.describe_repositories()

String(copy(r))
"{\"repositories\":[{\"createdAt\":1.601315527E9,\"encryptionConfiguration\":{\"encryptionType\":\"AES256\"},\"imageScanningConfiguration\":{\"scanOnPush\":false},\"imageTagMutability\":\"MUTABLE\",\"registryId\":\"{REDACTED}\",\"repositoryArn\":\"arn:aws:ecr:us-east-1:{REDACTED}:repository/mattbr-workplace\",\"repositoryName\":\"mattbr-workplace\",\"repositoryUri\":\"{REDACTED}.dkr.ecr.us-east-1.amazonaws.com/mattbr-workplace\"}]}"
```

### REST XML Service
```julia
using AWS
@service S3

args = Dict{String, Any}("return_raw"=>true)
r = S3.list_buckets(args)

String(copy(r))
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<ListAllMyBucketsResult xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\"><Owner><ID>my-id</ID><DisplayName>matt.brzezinski</DisplayName></Owner><Buckets><Bucket><Name>my-bucket-name</Name><CreationDate>2020-05-12T20:42:25.000Z</CreationDate></Bucket></Buckets></ListAllMyBucketsResult>"
```

### Query / EC2 Service
```julia
using AWS
@service EC2

args = Dict{String, Any}("return_raw"=>true)
r = EC2.describe_regions(args)

String(copy(r))
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<DescribeRegionsResponse xmlns=\"http://ec2.amazonaws.com/doc/2016-11-15/\">\n    <requestId>ba313eae-2f47-41ec-bbe8-5565bc835691</requestId>\n    <regionInfo>\n        <item>\n            <regionName>eu-north-1</regionName>\n            <regionEndpoint>ec2.eu-north-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-south-1</regionName>\n            <regionEndpoint>ec2.ap-south-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>eu-west-3</regionName>\n            <regionEndpoint>ec2.eu-west-3.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>eu-west-2</regionName>\n            <regionEndpoint>ec2.eu-west-2.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>eu-west-1</regionName>\n            <regionEndpoint>ec2.eu-west-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-northeast-3</regionName>\n            <regionEndpoint>ec2.ap-northeast-3.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-northeast-2</regionName>\n            <regionEndpoint>ec2.ap-northeast-2.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-northeast-1</regionName>\n            <regionEndpoint>ec2.ap-northeast-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>sa-east-1</regionName>\n            <regionEndpoint>ec2.sa-east-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ca-central-1</regionName>\n            <regionEndpoint>ec2.ca-central-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-southeast-1</regionName>\n            <regionEndpoint>ec2.ap-southeast-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>ap-southeast-2</regionName>\n            <regionEndpoint>ec2.ap-southeast-2.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>eu-central-1</regionName>\n            <regionEndpoint>ec2.eu-central-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>us-east-1</regionName>\n            <regionEndpoint>ec2.us-east-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>us-east-2</regionName>\n            <regionEndpoint>ec2.us-east-2.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>us-west-1</regionName>\n            <regionEndpoint>ec2.us-west-1.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n        <item>\n            <regionName>us-west-2</regionName>\n            <regionEndpoint>ec2.us-west-2.amazonaws.com</regionEndpoint>\n            <optInStatus>opt-in-not-required</optInStatus>\n        </item>\n    </regionInfo>\n</DescribeRegionsResponse>"
```

However, the body of the HTTP request when performing a `S3.get_object()` call is purely the contents of the file. I'm guessing the [Body](https://github.com/aws/aws-sdk-js/blob/82a02c317e2b71afec4eb379292855a07cbdfdd7/apis/s3-2006-03-01.normal.json#L4576) shape is special to AWS. I am unable to find usages of this `Body` in other services, or similar functionality.

My new proposition is to:
- Have `S3.get_object()` return back a `Vector{UInt8}`
- Add functionality in [JSON.jl](https://github.com/JuliaIO/JSON.jl), [XMLDict.jl](https://github.com/JuliaCloud/XMLDict.jl), to allow parsing via `Vector{UInt8}`, [JSON3.jl](https://github.com/quinnj/JSON3.jl) already supports this.

Since this functionality has been around since `AWS@1`, I would consier this a breaking change. I do not particularily want to tag `AWS@2` just for this, and instead would want to tack in the idea of using `ResponseObjects` outlined above. This is also nice because if removes the need for the `return_headers` argument, as we can place all information in the response object itself. I don't think that creating an output object for every services functions output makes much sense, but instead continue using dictionaries.

In regards to `S3.get_object()` this would result in a functional change for the user, where you would need to now access the file output as:

```julia
using AWS
@service S3

response = S3.get_object("mattbr-test-bucket", "foobar")
response["Body"]
```

This would then make it consistent with all other responses for services.

iamed2:
`Body` is defined in s3-2006-03-01.normal.json (line 1676) as type `"blob"`. I don't see `get_object` as special in any way. It seems to me like all the information on how to construct an object/dict given the HTTP response is present in that JSON file.

mattBrzezinski:
> `Body` is defined in s3-2006-03-01.normal.json (line 1676) as type `"blob"`. I don't see `get_object` as special in any way. It seems to me like all the information on how to construct an object/dict given the HTTP response is present in that JSON file.

It's "special" in that blob comes back without a key, it just dumps back the file contents. All-non blob values come back as something like, { 'Key': 'Value' }. I would expect this to come back as; `{ 'Body': {CONTENTS_OF_FILE} }` or `{ 'blob': {CONTENTS_OF_FILE} }`.

At least then its consistent with how other operations perform, although more inconvenient possibly for the end users.

iamed2:
I think this just indicates that we can't do eager parsing without checking first that we should parse.

BTW, this is not S3-specific, nor is it specific to things called "Body". It is the `shape` type `blob`. I made a JQ search that determines which operations return this kind of data: 

```
jq '(.shapes | to_entries | map(select(.value.type=="blob")) | map(.key)) as $blobshapes | (.shapes | to_entries | map(select(.value | has("payload"))) | map(select(.value.payload as $payload | $blobshapes | index($payload))) | map(.key)) as $blobresponses | (.operations | to_entries | map(select(.value.output.shape as $outputshape | $blobresponses | index($outputshape))) | map(.key)) as $operations | if ($operations | length)==0 then empty else {service: .metadata.uid, operations: $operations} end' *.normal.json
```

Here are the results:

```
{
  "service": "ebs-2019-11-02",
  "operations": [
    "GetSnapshotBlock"
  ]
}
{
  "service": "kinesis-video-archived-media-2017-09-30",
  "operations": [
    "GetClip",
    "GetMediaForFragmentList"
  ]
}
{
  "service": "kinesis-video-media-2017-09-30",
  "operations": [
    "GetMedia"
  ]
}
{
  "service": "location-2020-11-19",
  "operations": [
    "GetMapGlyphs",
    "GetMapSprites",
    "GetMapStyleDescriptor",
    "GetMapTile"
  ]
}
{
  "service": "polly-2016-06-10",
  "operations": [
    "SynthesizeSpeech"
  ]
}
{
  "service": "s3-2006-03-01",
  "operations": [
    "GetObject",
    "GetObjectTorrent"
  ]
}
{
  "service": "schemas-2019-12-02",
  "operations": [
    "GetCodeBindingSource"
  ]
}
```

femtomc:
Just a comment -- but the fact that things try to convert and then throw opaque errors like: 

image

is an absolutely horrendous experience for the user. When I see this, I have literally no idea what happened. I then have to go digging through this repository and the `AWSS3` repository to learn about this conversion business.

It is not an effective pathway for debugging.

mattBrzezinski:
> Just a comment -- but the fact that things try to convert and then throw opaque errors like:
> 
> is an absolutely horrendous experience for the user. When I see this, I have literally no idea what happened. I then have to go digging through this repository and the `AWSS3` repository to learn about this conversion business.
> 
> It is not an effective pathway for debugging.
> 
> AFAICT, this is thrown for many different types of errors on the side of AWS. It's really quite bad.

Yeah, this would be problematic in any version of `AWSS3.jl` when getting JSON files, and I'm sure others. Moving to a standardized return would resolve it.

mattBrzezinski:
Seems like we have a good idea of a path forward with this issue. I'm going to close this PR, and at some point make the necessary changes. Since they are breaking I've decided to create an issue for `AWS@2.0`,

- https://github.com/JuliaCloud/AWS.jl/issues/348

I didn't want to re-tag everyone in there, so please voice your breaking changes in there and / or making respective issues for them.

femtomc:
> > Just a comment -- but the fact that things try to convert and then throw opaque errors like:
> > is an absolutely horrendous experience for the user. When I see this, I have literally no idea what happened. I then have to go digging through this repository and the `AWSS3` repository to learn about this conversion business.
> > It is not an effective pathway for debugging.
> > AFAICT, this is thrown for many different types of errors on the side of AWS. It's really quite bad.
> 
> Yeah, this would be problematic in any version of `AWSS3.jl` when getting JSON files, and I'm sure others. Moving to a standardized return would resolve it.

I was not getting a JSON file, I was actually get an AWS error that I wasn't in the correct region (e.g. `aws_global_config` not set for correct region).

Issue #287, opened 2021-02-18
FR: support `credential_source = Ec2InstanceMetadata` in `~/.aws/config`


ericphanson:
I've an EC2 instance with a role (say role X), which is allowed to assume a role Y. I have a `~/.aws/config` that looks like
```
[profile role-Y]
role_arn = arn:aws:iam::NUMBER:role/Y
credential_source = Ec2InstanceMetadata
```
I can run commands using the AWS CLI within the EC2 instance like
```
aws --profile role-Y s3 cp ...
```
that work fine (and correctly give access denied without the `--profile` flag). However, if I do
```julia
using AWS
role_Y_config = AWSConfig(profile="role-Y")
write(S3Path(...; config=role_Y_config), ...)
```
then I get access denied. The issue seems to be that AWSConfig does not detect the credentials in the profile in `~/.aws/config` and instead falls back to the overall EC2 creds; in particular, I suspect if I had the creds in that `~/.aws/config` directly instead of 
`credential_source = Ec2InstanceMetadata` it might work.

mattBrzezinski:
I think is would be a good feature to have. Looking at the [documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) there are three parameters which `credential_source` can be set to:

* Environment – Specifies that the AWS CLI is to retrieve source credentials from environment variables.
* Ec2InstanceMetadata – Specifies that the AWS CLI is to use the IAM role attached to the EC2 instance profile to get source credentials.
* EcsContainer – Specifies that the AWS CLI is to use the IAM role attached to the ECS container as source credentials.

We would just need a mapping the values to their functions, and then if `credential_source` is set, parsing it and setting it to the `renew` field:

```julia
credential_source_mapping = Dict(
  "Environment" => env_var_credentials,
  "Ec2InstanceMetadata" => ec2_instance_credentials,
  "EcsContainer" => ecs_instance_credentials
)
```

ericphanson:
Cool, thanks for taking a look! I think that's not quite enough here, because for me `AWS.ecs_instance_credentials()` returns the credentials associated to the role assigned to the EC2 instance ("role X"), but **not** the credentials associated to the role that I want to assume using the profile ("role Y").

In case that doesn't make sense, going back to the config,
```
[profile role-Y]
role_arn = arn:aws:iam::NUMBER:role/Y
credential_source = Ec2InstanceMetadata
```
[...]

PR #349, opened 2021-04-28
Improve `ec2_instance_metadata` and create `ec2_instance_region`

omus:
Replaces the `_ec2_metadata` with the more generalized `ec2_instance_metadata`. The main change here is using HTTP.jl's `connect_timeout` which lets this function fail fast if running in and environment where the EC2 metadata is not available.

Additionally I added a `ec2_instance_region` function which determines the region of an instance when executed on the instance. I originally was planning on using this function as the default for `aws_get_region` but decided against it. It should still be useful in use cases such as:

```julia
global_aws_config(; region=something(AWS.ec2_instance_region(), "us-east-2"))
```

The one caveat with this change is that it requires a much newer version of HTTP.jl. Note the `connect_timeout` keyword was available in earlier revisions but had issues. Additionally, the special exception we're using was only available starting in HTTP.jl 0.9.6.

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -291,29 +291,29 @@ end
 
 
 """
-    _ec2_metadata(metadata_endpoint::String) -> Union{String, Nothing}
+    ec2_instance_metadata(path::AbstractString) -> Union{String, Nothing}
 
-Retrieve the EC2 meta data from the local AWS endpoint. Return the EC2 metadata request
-body, or `nothing` if not running on an EC2 instance.
+Retrieve the AWS EC2 instance metadata as a string using the provided `path`. If no instance
+metadata is available (typically due to not running within an EC2 instance) then `nothing`
+will be returned. See the AWS documentation for details on what metadata is available:
+https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
 
 # Arguments
-- `metadata_endpoint::String`: AWS internal meta data endpoint to hit
-    https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instance-metadata-ex-1
-
-# Throws
-- `StatusError`: If the response status is >= 300, except for 404
-- `ParsingError`: Invalid HTTP request target
+- `path`: The URI path to used to specify that metadata to return
 """
-function _ec2_metadata(metadata_endpoint::String)
-    try
-        request = @mock HTTP.request("GET", "http://169.254.169.254/latest/meta-data/$metadata_endpoint")
-
-        return request === nothing ? nothing : String(request.body)
+function ec2_instance_metadata(path::AbstractString)
+    uri = HTTP.URI(scheme="http", host="169.254.169.254", path=path)
+    request = try
+        @mock HTTP.request("GET", uri; connect_timeout=1)
     catch e
-        e isa HTTP.IOError || e isa HTTP.StatusError && e.status == 404 || rethrow(e)
+        if e isa HTTP.ConnectionPool.ConnectTimeout
+            nothing
+        else
+            rethrow()
+        end
     end
 
-    return nothing
+    return request !== nothing ? String(request.body) : nothing
 end
 
 
@@ -323,7 +323,7 @@ end
 Parse the EC2 metadata to retrieve AWSCredentials.
 """
 function ec2_instance_credentials()
-    info = _ec2_metadata("iam/info")
+    info = ec2_instance_metadata("/latest/meta-data/iam/info")
 
     if info === nothing
         return nothing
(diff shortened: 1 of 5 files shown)


omus:
bors try

omus:
Seems like I missed a patch. I took this opportunity to make the testset nicer too

omus:
bors try

omus:
bors try

src/AWSCredentials.jl
```diff
@@ -291,29 +291,41 @@ end
...
-- `StatusError`: If the response status is >= 300, except for 404
-- `ParsingError`: Invalid HTTP request target
+- `path`: The URI path to used to specify that metadata to return
 """
-function _ec2_metadata(metadata_endpoint::String)
-    try
-        request = @mock HTTP.request("GET", "http://169.254.169.254/latest/meta-data/$metadata_endpoint")
-
-        return request === nothing ? nothing : String(request.body)
+function ec2_instance_metadata(path::AbstractString)
```
mattBrzezinski, line 304: Can we keep prefacing this with an `_` as it's non-exported?

src/AWSCredentials.jl
```diff
@@ -291,29 +291,41 @@ end
...
+    uri = HTTP.URI(scheme="http", host="169.254.169.254", path=path)
+    request = try
+        @mock HTTP.request("GET", uri; connect_timeout=1)
```
mattBrzezinski, line 307: Is `1` potentially not too short?

src/AWSCredentials.jl
```diff
@@ -291,29 +291,41 @@ end
...
+
+"""
+    ec2_instance_region() -> Union{String, Nothing}
+
+Determine the AWS region of the machine executing this code if running inside of an EC2
+instance, otherwise `nothing` is returned.
+"""
+function ec2_instance_region()
+    az = ec2_instance_metadata("/latest/meta-data/placement/availability-zone")
+    return az !== nothing ? chop(az) : nothing
```
mattBrzezinski, line 328: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

```suggestion
    return ec2_instance_metadata("/latest/meta-data/placement/region")
```

ericphanson:
Does this resolve https://github.com/JuliaCloud/AWS.jl/issues/287? Or maybe it needs

> using this function as the default for `aws_get_region`

src/AWSCredentials.jl
omus: Alternatively, can we export this now?

mattBrzezinski: Sure, I suppose we're already exporting the other functions!

src/AWSCredentials.jl
omus: As the instance metadata uses a link-local address the HTTP server shouldn't actually require any network traversal and should be near instant. There is a chance that a system under heavy load could reach this timeout but that leaves the question: what should the timeout be? I'm inclined to give this a shot and change it if we find out it's too short.

Aside: as `connect_timeout` only supports integers this is the lowest timeout value we can use.

src/AWSCredentials.jl
omus: Oh nice, I was working off of some old knowledge here so possibly that's a new addition. In any case way better!

omus: I will say this makes this function way less of a necessity

omus:
> Does this resolve #287? Or maybe it needs
> 
> > using this function as the default for `aws_get_region`

That issue in particular will need support for `credential_source = Ec2InstanceMetadata`. Ultimately support for that does require access to the instance metadata but there is more work to be done there.

omus:
bors try

src/AWSCredentials.jl
omus: I did a little more research on this and discovered that the [instance metadata can be turned on/off via `aws ec2 modify-instance-metadata-options`](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) which may imply the metadata service runs externally from the instance itself.

I took a look at the route table and also attempted a trace route to attempt and determine a little more about how the service is setup but I don't have any clear answers.

omus: I used `lsof -i :80 | grep LISTEN` on an EC2 instance which has instance metadata available which should have listed something if the instance metadata service was running directly on the instance. Since every instance uses the same address and using `curl -v GET http://169.254.169.254/` doesn't reveal any identifying information of the source instance then the only way to identify the machine would be to use the IP address of the connection.

Pure speculation on my part but possibly the service runs on the instance host itself alongside the hypervisor?

mattBrzezinski: At first glance, I would have said to set it to `3`. If you are actively using this and don't mind maintaining it I have no problems with it!

mattBrzezinski (approved):
bors r+

ericphanson:
> I originally was planning on using this function as the default for `aws_get_region` but decided against it.

Why is that? I think it would be useful. Right now we set the config in `__init__` in some packages to ensure users get a working version, but that can be annoying (maybe they've already setup their config as they want it and don't want loading a package to mess with it). But I think if getting the region from the ec2 region was automatic then that would make it less necessary (and ensure that missing an env variable with the region in say a Dockerfile doesn't mean you get runtime errors from your job).

PR #351, opened 2021-04-29
Use a constant for the default region

omus:
Mainly just some code refactoring. Could be useful in conjunction with #349 such that you can do:

```julia
global_aws_config(; region=something(AWS.ec2_instance_region(), AWS.DEFAULT_REGION))
```

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -19,6 +19,8 @@ export ec2_instance_metadata, ec2_instance_region
 export global_aws_config, generate_service_url, set_user_agent, sign!, sign_aws2!, sign_aws4!
 export JSONService, RestJSONService, RestXMLService, QueryService
 
+const DEFAULT_REGION = "us-east-1"
+
 include("utilities.jl")
 include("AWSExceptions.jl")
 include("AWSCredentials.jl")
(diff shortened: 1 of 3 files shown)


omus:
We could also make use of this in the tests but I left them alone at the moment.

omus:
bors try

mattBrzezinski:
When the issues here are resolved I'll tag a new version with these changes alongside #349

src/AWS.jl
```diff
@@ -29,6 +29,7 @@ include("AWSMetadata.jl")
 using ..AWSExceptions
 using ..AWSExceptions: AWSException
 
+const DEFAULT_REGION = "us-east-1"
```
mattBrzezinski, line 32: I think you need to define this above `include("AWSCredentials.jl")`?

omus: Definitely. I'll move all the constants

mattBrzezinski: Yeah I'm starting to come across this here,

- https://github.com/JuliaCloud/AWS.jl/pull/352/files#diff-ec75e827e1f925b07f72f66dafb18428c8fae9572e458611da0b8421f6df6ca0R83-R84

I wonder if there's a better way of doing these

omus:
bors try

mattBrzezinski (approved):
bors r+

PR #352, opened 2021-04-29
NFC: Reorganized source tree

mattBrzezinski:
## Overview
The `src/AWS.jl` file and `src/AWSCredentials.jl` files were quite long and difficult to read and follow. I've created a new sub-directory `src/utilities/` and moved internal functions from various places into their respective files.


mattBrzezinski:
bors try

iamed2:
Feels weird to me that the API Generation utilities isn't in the same directory tree as the API Generation code. I feel like it should be `src/api_generation/utilities.jl` or `src/api_generation/utilities/utilities.jl` or something.

mattBrzezinski:
> Feels weird to me that the API Generation utilities isn't in the same directory tree as the API Generation code. I feel like it should be `src/api_generation/utilities.jl` or `src/api_generation/utilities/utilities.jl` or something.

I originally thought of this, and went this route instead. I'll swap it over, I'm 50/50 on it.

iamed2 (approved):

mattBrzezinski:
bors cancel

PR #357, opened 2021-04-30
Basic interactive MFA support

omus:
Experimenting with supporting MFA. Currently Julia will prompt you for an MFA token if you have defined `mfa_serial` in your config file. I'll note the AWS CLI seem to cache credentials under `~/.aws/cli/cache` so we can possibly use that information to avoid having to ask users for one-time passwords when switching between using the AWS CLI and AWS.jl. In an ideal world we would also be able to have AWS.jl write back to the cache.

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -9,7 +9,7 @@ using ..AWSExceptions
 export AWSCredentials,
     aws_account_number,
     aws_get_region,
-    aws_get_role_details,
+    aws_get_profile_settings,
     aws_user_arn,
     check_credentials,
     dot_aws_config,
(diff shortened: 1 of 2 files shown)


omus:
There's more that can be done with MFA support but the code should cover the basics for interactive usage. The main thing that was blocking this was the change to `aws_get_role_details` which was breaking. I've since made a new function `aws_get_profile_settings` and added an appropriate deprecation

ericphanson (approved):
Looks good to me :). Is there a way to test the MFA support?

omus:
> Looks good to me :). Is there a way to test the MFA support?

We can add tests for this but it would involve having a AWS user which has an MFA device attached. The two options for this are:
- Pre-create the MFA (and possibly a new user)
- Generate a new virtual MFA (probably via [`create-virtual-mfa-device`](https://docs.aws.amazon.com/cli/latest/reference/iam/create-virtual-mfa-device.html))

We'll then need some code to interpret the `otpauth` URI into the token. All of this is do-able but my personal preference is to punt this to an issue and have the basic functionality available as is.

PR #358, opened 2021-05-03
Use region as specifed by the profile

omus:
Adds support for using the [region specified by a config profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-settings).

While developing this I realized that a short-circuiting `something` function would be useful so I added it here. Ideally, this would be added into Base in the future.

### DIFF
--- a/src/AWSConfig.jl
+++ b/src/AWSConfig.jl
@@ -12,9 +12,16 @@ region(aws::AWSConfig) = aws.region
 function AWSConfig(;
     profile=nothing,
     creds=AWSCredentials(profile=profile),
-    region=get(ENV, "AWS_DEFAULT_REGION", DEFAULT_REGION),
+    region=nothing,
     output="json",
 )
+    region = @something(
+        region,
+        get(ENV, "AWS_DEFAULT_REGION", nothing),
+        dot_aws_region(profile),
+        DEFAULT_REGION,
+    )
+
     return AWSConfig(creds, region, output)
 end
 
--- a/src/AWSConfig.jl
+++ b/src/AWSConfig.jl
@@ -15,13 +15,7 @@ function AWSConfig(;
     region=nothing,
     output="json",
 )
-    region = @something(
-        region,
-        get(ENV, "AWS_DEFAULT_REGION", nothing),
-        dot_aws_region(profile),
-        DEFAULT_REGION,
-    )
-
+    region = @something region aws_get_region(profile=profile)
     return AWSConfig(creds, region, output)
 end
 
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -14,6 +14,7 @@ export AWSCredentials,
        check_credentials,
        dot_aws_config,
        dot_aws_credentials,
+       dot_aws_region,
        dot_aws_credentials_file,
        dot_aws_config_file,
        ec2_instance_credentials,
@@ -389,6 +390,29 @@ function dot_aws_config_file()
     get(ENV, "AWS_CONFIG_FILE") do
         joinpath(homedir(), ".aws", "config")
     end
+end
+
+
+"""
+    dot_aws_region(profile=nothing) -> Union{String, Nothing}
+
+Retrieve the region for the default or specified profile from the `~/.aws/config` file.
+If no region is specified then `nothing` is returned.
+
+# Arguments
+- `profile`: Specific profile used to get the region, default is `nothing`
+"""
+function dot_aws_region(profile=nothing)
+    region = nothing
+    config_file = dot_aws_config_file()
+
+    if isfile(config_file)
+        ini = read(Inifile(), config_file)
+        p = _aws_get_profile(profile)
+        region = _get_ini_value(ini, p, "region")
+    end
+
+    return region
 end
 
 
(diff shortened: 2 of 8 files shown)


src/AWSConfig.jl
```diff
@@ -12,9 +12,16 @@ region(aws::AWSConfig) = aws.region
...
+    region=nothing,
     output="json",
 )
+    region = @something(
+        region,
+        get(ENV, "AWS_DEFAULT_REGION", nothing),
+        dot_aws_region(profile),
+        DEFAULT_REGION,
+    )
+
```
omus, line 24: I think I should put this as a separate function but I'm unsure what I should call it.

test/AWSCredentials.jl
```diff
@@ -598,3 +598,45 @@ end
...
+                """
+            )
+            close(config_io)
+
+            withenv("AWS_CONFIG" => config_io) do
+                @test dot_aws_region() == test_values["Region"]
+            end
+        end
+    end
+end
```
omus, line 642: In looking over this credential tests file I think some reorganization is needed.

src/AWSCredentials.jl
```diff
@@ -472,6 +473,29 @@ function dot_aws_config(profile=nothing)
...
+"""
+    dot_aws_region(profile=nothing) -> Union{String, Nothing}
+
+Retrieve the region for the default or specified profile from the `~/.aws/config` file.
+If no region is specified then `nothing` is returned.
+
+# Arguments
+- `profile`: Specific profile used to get the region, default is `nothing`
+"""
+function dot_aws_region(profile=nothing)
```
omus, line 485: I'm up for suggestions into alternate names. It seems that functions like `dot_aws_config` are referring to `.aws/config`

omus:
bors try

omus:
bors try

src/AWSConfig.jl
mattBrzezinski: `get_region()`? And outline the order of how it is retrieve, similar to the credentials docstr?

omus:
bors try

src/AWSCredentials.jl
mattBrzezinski: You're correct with what they refer to. I think this name is fine, maybe in the future re-thinking what they are called in general, I didn't want to make them super long such as `dot_aws_config_get_region()` or something like that.

src/AWSConfig.jl
omus: I also just noticed that an `aws_get_region` function also exists

omus:
bors try

omus:
bors try

omus:
Needed to rebase everything to deal with conflicts.

src/AWSConfig.jl
omus: I've called this function `aws_get_region` and have dropped `dot_aws_region`

omus:
bors try

src/utilities/credentials.jl
```diff
@@ -31,6 +31,26 @@ function _get_ini_value(
...
 
+function _aws_profile_config(ini::Inifile, profile::AbstractString)
+    if profile != "default"
+        profile = "profile $profile"
+    end
+
+    return get(sections(ini), profile, Dict())
+end
+
+function _aws_profile_config(ini::Inifile, profile::Nothing=nothing)
```
omus, line 42:
```suggestion
function _aws_profile_config(ini::Inifile, profile::Nothing)
```

src/utilities/credentials.jl
```diff
@@ -31,6 +31,26 @@ function _get_ini_value(
...
+    _aws_profile_config(ini, _aws_get_profile())
+end
+
+function _aws_profile_config(config_file::AbstractString=dot_aws_config_file(), args...)
```
omus, line 46:
```suggestion
function _aws_profile_config(config_file::AbstractString, profile)
```

src/utilities/credentials.jl
```diff
@@ -31,6 +31,26 @@ function _get_ini_value(
...
+    isfile(config_file) || return Dict()
+    return _aws_profile_config(read(Inifile(), config_file), args...)
```
omus, line 48:
```suggestion
    return _aws_profile_config(read(Inifile(), config_file), profile)
```

src/utilities/credentials.jl
```diff
@@ -31,6 +31,26 @@ function _get_ini_value(
...
+end
+
+_aws_profile_config(config_file::Nothing, args...) = _aws_profile_config(args...)
```
omus, line 51:
```suggestion
function _aws_profile_config(config_file::Nothing, profile)
    return _aws_profile_config(dot_aws_config_file(), profile)
end
```

omus:
bors try

omus:
After a few iterations I'm liking this design direction. Either the design starting to shape up or I've just looked at it for too long

mattBrzezinski (approved):

mattBrzezinski:
> After a few iterations I'm liking this design direction. Either the design starting to shape up or I've just looked at it for too long

I like it, the `something` functionality is a bit, odd? If you're not used to it, but after looking at it for a while it makes sense.

PR #365, opened 2021-05-12
Add `@` to special-cased S3 escapes

kleinschmidt:
The `@` character is listed in the docs as a character that "may require special
handling".  This PR adds it to the list of special-cased escapes for S3 keys.


mattBrzezinski (approved):
bors try

mattBrzezinski:
Can you add in a test to create an object w/ `@` in the filename and very that it works.

mattBrzezinski (approved):
bors r+

PR #368, opened 2021-05-17
fix http_options in requests

zot:
Was getting an error trying to set http_options; this change makes them work

### DIFF
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -10,7 +10,7 @@ Base.@kwdef mutable struct Request
 
     return_stream::Bool=false
     response_stream::Union{<:IO, Nothing}=nothing
-    http_options::Array{String}=[]
+    http_options::Dict{Symbol,<:Any} = LittleDict{Symbol,String}()
     return_raw::Bool=false
     response_dict_type::Type{<:AbstractDict}=LittleDict
 end
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -10,7 +10,7 @@ Base.@kwdef mutable struct Request
 
     return_stream::Bool=false
     response_stream::Union{<:IO, Nothing}=nothing
-    http_options::Dict{Symbol,<:Any} = LittleDict{Symbol,String}()
+    http_options::Dict{Symbol,<:Any} = Dict{Symbol,String}()
     return_raw::Bool=false
     response_dict_type::Type{<:AbstractDict}=LittleDict
 end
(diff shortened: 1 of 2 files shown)


mattBrzezinski (approved):
bors try

zot:
Checking the tests...


-- Bill

zot:
bors try

zot:
Guess I can't do that :)

mattBrzezinski:
bors try

src/utilities/request.jl
```diff
@@ -10,7 +10,7 @@ Base.@kwdef mutable struct Request
 
     return_stream::Bool=false
     response_stream::Union{<:IO, Nothing}=nothing
-    http_options::Array{String}=[]
+    http_options::Dict{Symbol,<:Any} = Dict{Symbol,String}()
```
mattBrzezinski, line 13:
```suggestion
    http_options::AbstractDict{Symbol,<:Any}=LittleDict{Symbol,String}()
```

src/utilities/utilities.jl
```diff
@@ -65,7 +65,7 @@ function _extract_common_kw_args(service, args)
         return_raw=_pop!(args, "return_raw", false),
         response_stream=_pop!(args, "response_stream", nothing),
         headers=LittleDict{String, String}(_pop!(args, "headers", [])),
-        http_options=_pop!(args, "http_options", []),
+        http_options=_pop!(args, "http_options", Dict{Symbol, String}()),
```
mattBrzezinski, line 68:
```suggestion
        http_options=_pop!(args, "http_options", LittleDict{Symbol, String}()),
```

mattBrzezinski (approved):
Just for consistencies as other places use LD everywhere.

kogbill:
I tried that but using LD produces errors during the conversion calls.

zot:
Sorry, "kogbill" up there is me.

I tried using LittleDict but it causes conversion errors with http_options.

mattBrzezinski:
> Sorry, "kogbill" up there is me.
> 
> I tried using LittleDict but it causes conversion errors with http_options.

I just pulled down the branch, implemented the suggested changes above and it work. I think you may have not loosened the restriction on the `Request.http_options` type from `Dict -> AbstractDict` which would have cause an error when trying to default to a `LittleDict`.

kogbill:
OK, whatever changes you want to make are just fine with me!

mattBrzezinski (approved):
bors r+

Issue #348, opened 2021-04-28
AWS.jl@2


mattBrzezinski:
There are a couple issues and feature requests which would be nice to introduce. I'm creating this issue to keep track of all breaking changes which the community would introduce for a v2 of this package.

- #224 : Always return `Vector{UInt8}` for `S3.get_object()`
- #341 : Support kwargs for optional parameters instead of passing in a Dictionary

After a list is created we can create a feature branch to merge changes into and then go to master.

omus:
Note that #384 supersedes #224 by providing a more robust return type for AWS API calls.

omus:
We may also want to consider switching to JSON3 for AWS@2 (https://github.com/JuliaCloud/AWS.jl/issues/438)

mattBrzezinski:
> We may also want to consider switching to JSON3 for AWS@2 (#438)

Agreed. Also this is a known problem and one internally that we are wanting to address as well here at Invenia,

https://github.com/JuliaCloud/AWS.jl/issues/519

omus:
An update on AWS.jl@v2. I've made some progress on this work and have been tracking the various issues for a v2 in the [2.0 milestone](https://github.com/JuliaCloud/AWS.jl/issues?q=is%3Aissue%20state%3Aopen%20milestone%3A2.0). There are a few outstanding tasks I'd love to tackle as part of a new major version since we need to make a breaking change but I'm definitely bandwidth constrained and may not have time to tackle this. If I do find that progress is stalled out I'll just release v2 as is.

PR #384, opened 2021-06-18
Use `AWS.Response` to handle streaming/raw/parsing

omus:
The idea is to use a struct in AWS.jl that can be used to handle the automatic parsing that is currently used to turn XML/JSON into a dict while also giving the option of accessing the raw output as a string or stream without all the keywords currently needed to be specified.

Depends on:
- #457

Related:
- #346
- #348
- #438

Closes:
- #224
- #433
- #468 (when using `use_response_type` the passed in I/O is not closed)
- #471
- #433

Update: The tests in this PR run using the deprecated behaviour. Mainly I did that here to prove this change is non-breaking. For the updated tests see #458

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -36,6 +36,8 @@ include(joinpath("utilities", "response.jl"))
 include(joinpath("utilities", "sign.jl"))
 include(joinpath("utilities", "downloads_backend.jl"))
 
+include("deprecated.jl")
+
 using ..AWSExceptions
 using ..AWSExceptions: AWSException
 
@@ -228,7 +230,7 @@ function (service::RestXMLService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
-    _delete_outdated_kw_args!(args)
+    return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
@@ -254,7 +256,7 @@ function (service::RestXMLService)(
         aws_config, service.endpoint_prefix, request.resource
     )
 
-    return submit_request(aws_config, request)
+    return submit_request(aws_config, request; return_headers=return_headers)
 end
 
 """
@@ -282,7 +284,7 @@ function (service::QueryService)(
     feature_set::FeatureSet=FeatureSet(),
 )
     POST_RESOURCE = "/"
-    _delete_outdated_kw_args!(args)
+    return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
@@ -297,7 +299,7 @@ function (service::QueryService)(
     args["Version"] = service.api_version
     request.content = HTTP.escapeuri(_flatten_query(service.signing_name, args))
 
-    return submit_request(aws_config, request)
+    return submit_request(aws_config, request; return_headers=return_headers)
 end
 
 """
@@ -325,7 +327,7 @@ function (service::JSONService)(
     feature_set::FeatureSet=FeatureSet(),
 )
     POST_RESOURCE = "/"
-    _delete_outdated_kw_args!(args)
+    return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
@@ -338,7 +340,7 @@ function (service::JSONService)(
     request.headers["Content-Type"] = "application/x-amz-json-$(service.json_version)"
     request.headers["X-Amz-Target"] = "$(service.target).$(operation)"
 
-    return submit_request(aws_config, request)
+    return submit_request(aws_config, request; return_headers=return_headers)
 end
 
 """
@@ -367,7 +369,7 @@ function (service::RestJSONService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
-    _delete_outdated_kw_args!(args)
+    return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
@@ -386,7 +388,7 @@ function (service::RestJSONService)(
     request.headers["Content-Type"] = "application/json"
     request.content = json(args)
 
-    return submit_request(aws_config, request)
+    return submit_request(aws_config, request; return_headers=return_headers)
 end
 
 function (service::ServiceWrapper)(args...; feature_set=nothing, kwargs...)
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -44,7 +44,9 @@ using ..AWSExceptions: AWSException
 const user_agent = Ref("AWS.jl/1.0.0")
 const aws_config = Ref{AbstractAWSConfig}()
 
-Base.@kwdef struct FeatureSet end
+Base.@kwdef struct FeatureSet
+    use_response_type::Bool = false
+end
 
 """
     global_aws_config()
@@ -230,10 +232,13 @@ function (service::RestXMLService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
+    feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
+
     return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
+        use_response_type=feature_set.use_response_type,
         request_method=request_method,
         content=_pop!(args, "body", ""),
     )
@@ -283,11 +288,14 @@ function (service::QueryService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
+    feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
+
     POST_RESOURCE = "/"
     return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
+        use_response_type=feature_set.use_response_type,
         resource=POST_RESOURCE,
         request_method="POST",
         url=generate_service_url(aws_config, service.endpoint_prefix, POST_RESOURCE),
@@ -326,11 +334,14 @@ function (service::JSONService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
+    feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
+
     POST_RESOURCE = "/"
     return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
+        use_response_type=feature_set.use_response_type,
         resource=POST_RESOURCE,
         request_method="POST",
         content=json(args),
@@ -369,10 +380,13 @@ function (service::RestJSONService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
+    feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
+
     return_headers = _pop!(args, "return_headers", nothing)
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
+        use_response_type=feature_set.use_response_type,
         request_method=request_method,
         resource=_generate_rest_resource(request_uri, args),
     )
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -32,6 +32,7 @@ include("AWSConfig.jl")
 include("AWSMetadata.jl")
 
 include(joinpath("utilities", "request.jl"))
+include(joinpath("utilities", "response.jl"))
 include(joinpath("utilities", "sign.jl"))
 include(joinpath("utilities", "downloads_backend.jl"))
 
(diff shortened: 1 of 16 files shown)


omus:
At the moment I've made the basic `AWS.Response` structure and I am using it as part of `submit_request`. I've also introduced a shim into the service request calls to re-introduce the legacy behaviour while I'm working through the details

omus:
bors try

src/utilities/response.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+function legacy(r::Response; response_dict_type::Type, return_headers=false, return_stream=false, return_raw=false, response_stream=nothing)
+    # For HEAD request, return headers...
+    if r.response.request.method == "HEAD"
+        return response_dict_type(r.response.headers)
+    end
+
+    # Return response stream if requested...
+    if return_stream
```
mattBrzezinski, line 63: Should we also be checking if you want headers returned back and a streamed response?

src/utilities/response.jl
```diff
@@ -0,0 +1,75 @@
...
+    # if isempty(mime) && read(r.body, 5) == b"<?xml"
+    #     "xml"
+
+    # e.g. "application/xml" or "text/xml"
+    if occursin(r"/xml", mime)
+        MIME(Symbol("application/xml"))
+    elseif occursin(r"/x-amz-json-1.[01]$", mime) || endswith(mime, "json")
+        MIME(Symbol("application/json"))
+    elseif startswith(mime, "text/")
+        MIME(Symbol("text/plain"))
```
mattBrzezinski, line 31: Could just drop this case

src/AWSMetadata.jl
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
omus: Oh weird

src/utilities/response.jl
omus: Nope, the idea is you'll get a `Request` object back and you can pull the headers and/or content out without having to specify this information. This `legacy` function is just a shim to not break everything at the moment

src/utilities/response.jl
omus: Correct. I'll probably end up changing the `else` to use another type but haven't worked out what yet

src/utilities/response.jl
mattBrzezinski: Ah right, I'm briefly trying to look through all these PRs here. Since that's the case and we can pull it from the `Request` object back can we drop it from the `return_raw` and default return blocks of code below?

omus:
I'll pick this up again next week. Lots of details to work though yet but I think overall this is looking like the right approach. There is still a question of how to make this very user friendly for newcomers

src/utilities/response.jl
```diff
@@ -0,0 +1,75 @@
+# TODO: Include only the fields we care about
+struct Response
+    response::HTTP.Response
+    body::BufferStream
```
christopher-dG, line 4: This should probably be `IO`, the Downloads backend at least uses `IOBuffer`.

omus:
bors try

omus:
Things feel like they are starting to align here. Overall the concept of returning a `AWS.Response` type seems to be the right approach: it allows us to use dispatch to process the response in several different ways and allows end users to process returned data with whatever tooling they want (#438).

The change is less breaking than I originally expected as the `AWS.Response` implements the dictionary interface which provides some backwards compatibility. I need to think through the various other ways things were previously being handled yet to make sure I'm not missing something though.

I'm sure I've got some rounds of debugging to do yet but the only major change I anticipate is I still may want to support passing in a user supplied `IO` type. The reason for this is that it allows users to write directly to disk or to some other source where as currently we're always using an `IOBuffer`. I also need to validate the alternate download backends are working correctly.

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
Once https://github.com/JuliaWeb/HTTP.jl/pull/752 is tagged I can clean up some of the I/O logic here

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
As I need to resolve these conflicts I'll also take the time to squash some of these commits

omus:
bors try

omus:
Alright! Finally got this working with the test suite. Not being to run the suite locally made this process more painful than it should have been.

The final work remaining is to see if I can make this change close to being non-breaking. Here are the current known incompatibilities:
- `isa` checks against the `AWS.Response` (not fixable as `isa` is a built-in)
- `==` checks against the `AWS.Response` (can extend `==` to address)
- `return_headers` keyword expects a tuple to be returned (I can re-add support for this with a deprecation)

omus:
I experimented with having an extra parameter or keyword to enable the new functionality on demand but it became very annoying to try and update all the various calls. I instead am experimenting with an idea of being able to enable a feature when calling `@service`. The advantage primarily being you can opt-in to a breaking change like this with minimal churn with the old behaviour is eventually dropped

omus:
bors try

omus:
bors try

omus:
I moved the service feature code into a separate PR: #457

omus:
In restoring the "issue 324" testset I've realized that `Base.BufferStream` is now an incompatible I/O type even when using the backwards compatible mode. I'll see what can be done here

omus:
bors try

omus:
bors try

omus:
Rebased one last time to use the latest commit from #457

omus:
bors try

src/AWSExceptions.jl
```diff
@@ -27,39 +29,65 @@ struct AWSException <: Exception
...
-    error_content_types = ["", "application/xml", "text/xml"]
-    if content_type(e) in error_content_types && is_valid_xml_string(http_message(e))
-        info = parse_xml(http_message(e))
+        # Extract API error code from XML error message...
+        error_content_types = ["", "application/xml", "text/xml"]
+        if content_type(e) in error_content_types && is_valid_xml_string(body)
+            info = parse_xml(body)
+        end
+    elseif parse(Int, HTTP.header(e.response, "Content-Length", "0")) > 0
+        @error "Internal Error: provided body is empty while the reported content-length is non-zero"
```
omus, line 90: This error exists mainly for developers who are changing the `_http_request` functions. If you make a mistake you'll get an `AWSException` without any error content due to a flaw in how the stream was handled.

src/utilities/response.jl
```diff
@@ -0,0 +1,129 @@
...
+
+function mime_type(r::Response)
+    # Parse response data according to mimetype...
+    mime = HTTP.header(r.response, "Content-Type", "application/octet-stream")
+
+    # https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+
+    # Reading consumes the buffer so this is unsafe currently
+    # if isempty(mime) && read(r.io, 5) == b"<?xml"
+    #     "xml"
```
omus, line 30: Given that `AWS.Response` requires the I/O to be seekable we can use `_rewind` here. I haven't seen a scenario in which this is needed but that doesn't mean it doesn't exist

src/utilities/response.jl
```diff
@@ -0,0 +1,129 @@
...
+
+function Base.iterate(r::Response, state)
+    iter, s = state
+    x = iterate(iter, s)
+    x === nothing && return nothing
+    el, s = x
+    return (el, (iter, s))
+end
+
+Base.String(r::Response) = read(r.io, String)
```
omus, line 106: Originally this code was added to provide backwards compatibility. I'd be fine with dropping this and telling the users if they want this kind of support they need to handle it themselves. What do you think?

Related: If we do drop this then I would update the `show` method to print a pretty printed version of the stream (XML/JSON)

test/AWS.jl
```diff
@@ -541,6 +540,28 @@ end
...
+
+    @testset "low-level" begin
+        response = AWSServices.sts("GetCallerIdentity")
+        d = response["GetCallerIdentityResult"]
+
+        @test Set(keys(d)) == Set(["Arn", "UserId", "Account"])
+        @test occursin(r"^arn:aws:(iam|sts):", d["Arn"])
+        @test all(isdigit, d["Account"])
+    end
+end
```
omus, line 563: Added these as STS is generally available for all AWS users

test/runtests.jl
```diff
@@ -19,6 +19,7 @@ using AWS.AWSMetadata:
 using Base64
 using Compat: mergewith
 using Dates
+using Downloads: Downloads
```
omus, line 22: Used in the "Not authorized" test

omus:
Something we may want to address. We can't use the `@service` trick when accessing the low-level API. This can be annoying for the same reasons as mentioned in https://github.com/JuliaCloud/AWS.jl/pull/384#issuecomment-926140879. I don't think it's a big deal but it does mean the deprecation warnings may be misleading for those using the low-level functions.

omus:
Rebased to address conflicts with updates to `cv/service-features`

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

src/AWS.jl
```diff
@@ -227,10 +258,13 @@ function (service::RestXMLService)(
     aws_config::AbstractAWSConfig=global_aws_config(),
     feature_set::FeatureSet=FeatureSet(),
 )
-    return_headers = _pop!(args, "return_headers", false)
+    feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
```
mattBrzezinski, line 261: Maybe we can generalize code like this in a `function (service::Service)` to remove code duplication?

src/utilities/request.jl
```diff
@@ -33,12 +32,17 @@ Base.@kwdef mutable struct Request
...
+    # Note: User provided `IO` should support seeking in order to support API error handling
+    response_stream::Union{IO,Nothing} = nothing
+
     http_options::AbstractDict{Symbol,<:Any} = LittleDict{Symbol,String}()
-    return_raw::Bool = false
-    response_dict_type::Type{<:AbstractDict} = LittleDict
     backend::AbstractBackend = DEFAULT_BACKEND[]
+
+    # Deprecated fields
+    use_response_type::Bool = false
```
mattBrzezinski, line 42: Should we add deprecations when users attempt to access or set these fields?

omus: The deprecation warning when using these fields occurs when using `submit_request`. I'll note the reason `use_response_type` is under this deprecated heading is because it should be removed for the 2.0 release.

src/AWS.jl
omus: I'll look into that. It'll involve using `invoke` so we'll see how clean it looks

omus: A couple of things came out of this:
1. It doesn't seem possible to `invoke` an abstract functor: https://github.com/JuliaLang/julia/issues/42414
2. Due to this function needing access to `args` it's rather difficult to write a nice looking abstract functor

An alternate option would just be to use a private function to reduce the amount of duplicate code. Really that function would just be for `return_headers` and `_delete_legacy_response_kw_args!`. I'm not sure it's worth it. I think with additional refactoring of `Request` something nicer can be done but I think that's out of the scope of this PR.

omus:
Rebased to deal with #467 conflicting changes. Added a new commit to ensure compatibility

omus:
bors try

src/AWS.jl
```diff
@@ -116,13 +135,24 @@ using AWS: @service
...
+
+### Features
+- `use_response_type::Bool=false`: When enabled, high-level service calls will return an
+  `AWS.Response` which provides streaming/raw/parsed access to the response. When disabled,
+  the service call response typically be parsed but will vary depending on the following
+  parameters: "return_headers", "return_stream", "return_raw", "response_dict_type".
 
 # Return
-- `Expression`: Base.include() call to introduce the high-level service API wrapper functions in your namespace
+- `Expression`: Module which embedsd the high-level service API wrapper functions in your namespace
```
laborg, line 155:
```suggestion
- `Expression`: Module which embeds the high-level service API wrapper functions in your namespace
```

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

omus:
Looks like 0.9.15 is going to get yanked from the registry as it's being deemed as too breaking (https://github.com/JuliaWeb/HTTP.jl/pull/773). I'll need to revert the changes in "Remove workarounds for HTTP 0.9.14 and below"

omus:
bors try

test/AWS.jl
```diff
@@ -116,250 +116,247 @@ end
...
-                result = AWS.submit_request(aws, request)
+        message = "User is not authorized to perform: action on resource with an explicit deny"
+        body = "{\"__type\":\"AccessDeniedException\",\"Message\":\"$message\"}"
+        headers = [
+            "Content-Type" => "application/x-amz-json-1.1",
+            "Content-Length" => string(sizeof(body)),
+        ]
+
+        # Simulate the HTTP.request behaviour with a HTTP 400 response
+        patches = [
```
mattBrzezinski, line 212: Maybe move these into the patches file?

mattBrzezinski (approved):
Just one minor comment about moving patches into the respective file.
LGTM, feel free to merge whenever!

test/AWS.jl
omus: I can do that. I do personally prefer having non-reused patches closer to where they are used but I can stick to the convention

omus:
bors try

omus:
The amazing @mattBrzezinski was willing to review this giant PR. Thank you oh so much ❤️

omus:
Once I verify the latest changes pass the CI gauntlet I'll rebase the commits to apply the various fixups and we should be good to merge this.

omus:
Time to rebase.

omus:
Validated that #458 is working this these rebased changes: https://github.com/JuliaCloud/AWS.jl/pull/458#issuecomment-930538119

PR #385, opened 2021-06-18
Refactor `_aws_http_request_patch`

omus:
Previously this patch worked by mutating the global `response`. This revises the code to make new `HTTP.Message.Response` instances and effectively treats them as immutable.


omus:
bors try

mattBrzezinski (approved):
LGTM

PR #396, opened 2021-07-10
Alternate Downloads.jl-based backend (alternative to HTTP.jl)

ericphanson:
I was running into issues with HTTP when trying to make many concurrent requests to s3, so I thought I'd try libcurl through the Downloads.jl stdlib. Turned out to be pretty easy to at least get something basic working, e.g. I can pull down a test file from s3 with no problems with this little bit of code-- and you can even pass a `Downloads.Downloader` along to have more control over the client. I haven't had a chance to do more thorough testing, but I thought I'd put the PR up so maybe we can see what CI says to test out the Downloads-based backend. (Right now Downloads.jl is set to the default in this PR for testing purposes, but I think we should keep HTTP.jl as the default unless it really is a big upgrade).


ericphanson:
I tried a simple `asyncmap` loop to pull down many small objects (380k evenly sized objects totalling ~7GB), just

```julia
using Serialization, AWS, AWSS3
@time results = @showprogress asyncmap(keys) do key
    value = deserialize(IOBuffer(s3_get(bucket, key; raw=true))) # these were serialized Julia objects
    return key => value
end
```

With the HTTP.jl backend I left it overnight and it was stuck at 0% in the morning-- I've seen this in other experiments too, and calling
```julia
empty!(HTTP.ConnectionPool.POOL.conns)
```
tends to un-stick it for a little bit, but it gets stuck again when using many concurrent connections. @ararslan has said he had similar issues awhile ago and isolated it to MbedTLS issues.

But with the Downloads.jl backend it completed in 12 minutes with no issues-- and probably can be optimized further, this was just the simplest approach to start with (using the global Downloader with 100 concurrent tasks in an `asyncmap` loop).

src/utilities/request.jl
```diff
@@ -13,6 +13,31 @@ Base.@kwdef mutable struct Request
     http_options::AbstractDict{Symbol,<:Any}=LittleDict{Symbol,String}()
     return_raw::Bool=false
     response_dict_type::Type{<:AbstractDict}=LittleDict
+    downloader::Union{Nothing, Downloads.Downloader}=nothing
```
ericphanson, line 16: allows the user to pass a `Downloads.Downloader` to use instead of the global one

src/utilities/request.jl
```diff
@@ -13,6 +13,31 @@ Base.@kwdef mutable struct Request
...
+end
+
+abstract type AbstractBackend end
+struct HTTPBackend <: AbstractBackend end
+struct DownloadsBackend <: AbstractBackend end
+
+default_backend() = DownloadsBackend()
+
+submit_request(aws::AbstractAWSConfig, request::Request; return_headers::Bool=false) = submit_request(default_backend(), aws, request; return_headers)
```
ericphanson, line 25: right now, the user changes the backend by defining
```julia
AWS.default_backend() = AWS.HTTPBackend()
```
for example.

Another option would be to use [Preferences.jl](https://github.com/JuliaPackaging/Preferences.jl) to set a preference, but that's 1.6+ only.

The word "default" here is kind of misleading because there's no other way to change it (e.g. on a per-request basis). I'm not sure if anyone would want to mix backends between different requests or not.

I think this maybe should start as an experimental feature that the user opts into, and if it works well then Downloads.jl could be promoted to the default.

src/utilities/request.jl
```diff
@@ -13,6 +13,31 @@ Base.@kwdef mutable struct Request
...
+
+function _http_request(::DownloadsBackend, request)
+    output = IOBuffer()
+    input = IOBuffer()
+    write(input, request.content)
+
+    response = Downloads.request(request.url; input, output, method = request.request_method, request.headers, verbose=false, throw=true, request.downloader)
```
ericphanson, line 32:
```suggestion
    response = Downloads.request(request.url; input, output, method = request.request_method,
                                 request.headers, verbose=false, throw=true, request.downloader)
```

src/utilities/request.jl
mattBrzezinski: Ugh the more I look at this code I wrote the more I hate it. Setting a default here, then defaults to a value in each of the Request functions to default to a value feels bad.

mattBrzezinski (commented):
LGTM I'll kick `bors` to give it a shot!

mattBrzezinski:
bors try

ericphanson:
Oh no: https://github.com/JuliaLang/Downloads.jl/blob/cd002c3c6936d144ae668d70e18337931706c63a/Project.toml#L15. That might be a big problem since we support 1.0 here. Would AWS.jl be up for going to 1.3+? Maybe at least once 1.6 is the LTS?

mattBrzezinski:
> Oh no: https://github.com/JuliaLang/Downloads.jl/blob/cd002c3c6936d144ae668d70e18337931706c63a/Project.toml#L15. That might be a big problem since we support 1.0 here. Would AWS.jl be up for going to 1.3+? Maybe at least once 1.6 is the LTS?

I'd *like* to keep support for all LTS builds. It seems that with 1.6 and your [Preferences.jl](https://github.com/JuliaCloud/AWS.jl/pull/396#discussion_r667780862) comment that this would be the best version going forward.

src/utilities/request.jl
```diff
@@ -31,7 +111,7 @@ Submit the request to AWS.
 # Returns
 - `Tuple or Dict`: Tuple if returning_headers, otherwise just return a Dict of the response body
 """
-function submit_request(aws::AbstractAWSConfig, request::Request; return_headers::Bool=false)
+function submit_request(backend::AbstractBackend, aws::AbstractAWSConfig, request::Request; return_headers::Bool=false)
```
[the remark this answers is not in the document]
ericphanson: Yeah, that makes sense

ericphanson:
Just as an update on this-- I'm finding various edge cases when trying to use it internally, leading to fixes like https://github.com/JuliaCloud/AWS.jl/pull/396/commits/a481a46b6210390bb7dd9be45812bd7238b78362 and https://github.com/JuliaCloud/AWS.jl/pull/396/commits/16419cdcb5c593fadeaa7d132b3a1cf1498626b6. I am still having issues using it with `credentials_from_webtoken()`; for some reason I get a redirect (which led to #397) instead of the credentials.

I'm now using https://github.com/JuliaCloud/AWS.jl/pull/396/commits/57141e49fb93a858ca0650867bb3ef07a92c14d2 on my internal project in order to use the HTTPBackend for configuration and the Downloads backend for `s3_get`, to at least make progress there.

I'm thinking this might be a bigger job than I anticipated in order to sort out the edge cases and match the current implementation's handling of various things, so with that and the lack of 1.0 support in Downloads.jl, I'm thinking this might be a slow work-in-progress for awhile. But I plan to actively use this branch at least for `s3_get` because in my use it takes 12 minutes with `asyncmap` with Downloads.jl instead of ~ 4hr for serial access with the HTTP.jl backend, and hopefully I'll be able to find and fix edge cases over time so that when 1.6 LTS lands this might be reliable enough.

ericphanson:
I've redone this on top of #401. It requires Julia 1.3+ so it'll need to wait until the new LTS, but we can at least start to get it ready for then.

bors try

ericphanson:
bors try

ericphanson:
bors try

c42f:
Thanks for this @ericphanson, we're in a similar position of having trouble with HTTP.jl and S3 in production. It will be great to have the option to use Downloads.jl. I've been making some progress on the HTTP.jl side over at https://github.com/JuliaWeb/HTTP.jl/issues/517 and I think I may have identified the problem and a potential solution.

Did you consider whether it's possible to add an alternative Downloads-based backend to HTTP.jl instead of here in AWS.jl? That would fix the problem for all libraries using HTTP rather than AWS.jl only. One idea would be to allow an alternative backend to replace the connection and stream layers (maybe also other layers) in the HTTP default stack (as returned by `HTTP.stack()`).

ericphanson:
> Thanks for this @ericphanson, we're in a similar position of having trouble with HTTP.jl and S3 in production. It will be great to have the option to use Downloads.jl. I've been making some progress on the HTTP.jl side over at JuliaWeb/HTTP.jl#517 and I think I may have identified the problem and a potential solution.

Woah! That's great. I think I saw that issue awhile ago but somehow forgot about it. That looks like great progress. Perhaps an alternate backend isn't really needed then, if HTTP's client issues instead can be fixed. I found it a bit too intimidating to try to fix myself which is why I switched to just swapping over to Downloads.jl.

> Did you consider whether it's possible to add an alternative Downloads-based backend to HTTP.jl instead of here in AWS.jl? That would fix the problem for all libraries using HTTP rather than AWS.jl only. One idea would be to allow an alternative backend to replace the connection and stream layers (maybe also other layers) in the HTTP default stack (as returned by HTTP.stack()).

I did-- I actually remember that was once floated as a plan for HTTP.jl back when there was a bit of talk about making it a stdlib to replace `Base.download`. I believe that effort led to Downloads.jl instead. I asked a week or two ago on #web on Slack about using libcurl for the clientside functionality of HTTP.jl and Avik suggested using Downloads.jl directly instead, and that's what I've done here. But I think doing it on the HTTP.jl side could make sense too. #401 could be used here to swap between HTTP.jl-based-backends here too, instead of between direct-Downloads.jl and HTTP.jl backends.

c42f:
> I found it a bit too intimidating to try to fix myself

Agreed, the code is kind of intimidating to dive into! Especially the connection pool which is where I believe the issues are.

> I believe that effort led to Downloads.jl instead. I asked a week or two ago on #web on Slack about using libcurl for the clientside functionality of HTTP.jl and Avik suggested using Downloads.jl directly instead, and that's what I've done here.

Yes, I think this is great. Even if we have a larger plan that shouldn't block making progress here in AWS.jl. What you've got here is heading in the right direction IMO and would be extremely useful for me (I just need to port all our code from AWSSDK to AWS.jl :grimacing: )

Jacob isn't very available right now but I managed to catch him on a DM on slack. He's of the opinion that HTTP and Downloads should come together in some way so there's definitely appetite for change if we can pull together some sensible PRs and get things moving on that front. Perhaps I should open an issue in HTTP.jl with thoughts about how to proceed.

ericphanson:
Cool! I unfortunately don't have a lot of time for this (/ HTTP.jl) right now-- this draft of a PR is enough that I can performantly access my cached data on s3 which unblocks me to do the actual thing I'm working on (working on an ML model which happens to train on k8s in AWS) but I hope to at least be able to push this PR through in the next few weeks.

mattBrzezinski:
> Cool! I unfortunately don't have a lot of time for this (/ HTTP.jl) right now-- this draft of a PR is enough that I can performantly access cached data on s3 which unblocks me to do the actual thing I'm working on (working on an ML model which happens to train on k8s in AWS) but I hope to at least be able to push this PR through in the next few weeks.

I think we can definitely get this done and unblock everyone here! I'll be gone post-JuliaCon until August 9th (feel free to push through if I'm gone). I'm hoping we announce a new LTS at JuliaCon this year, which should make this PR go through easily enough. If not, I'm sure we can create something that works as an alternative.

.github/workflows/CI.yml
```diff
@@ -25,9 +25,9 @@ jobs:
         arch:
           - x64
         include:
-          # Add a 1.0 job just to make sure we still support it
+          # Add a 1.3 job just to make sure we still support it
```
omus, line 28:
```suggestion
          # Add a job using the earliest version of Julia supported by this package
```

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,87 @@
...
+    else
+        output_arg = NamedTuple()
+        body_arg = () -> NamedTuple()
+    end
+
+    # We pass an `input` only when we have content we wish to send.
+    if !isempty(request.content)
+        input = IOBuffer()
+        write(input, request.content)
+        input_arg = (; input=input)
```
ericphanson, line 60: Hm, maybe this will fix things?
```suggestion
        seekstart(input)
        input_arg = (; input=input)
```

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
CI is timing out; something is hanging unfortunately. I'll need to introduce some logging or something to see where.

christopher-dG:
bors try

christopher-dG:
bors try

I think I've fixed some bugs but the hang is still there.

christopher-dG:
bors try

should be down to 0 errors except the hang 🤞

christopher-dG:
bors try

christopher-dG:
bors try

christopher-dG:
bors try

christopher-dG:
CI passed!

Project.toml
```diff
@@ -32,7 +33,7 @@ OrderedCollections = "1"
 Retry = "0.3, 0.4"
 URIs = "1"
 XMLDict = "0.3, 0.4"
-julia = "1"
+julia = "1.3"
```
mattBrzezinski, line 36: I've spent some time thinking about this and going forward I'm okay with bumping the minimum version. However, I would want a plan in place if any bug fixes need to be back ported to the version support `1`.

I've never had to do it before, but how would you go about doing so? For instance say `AWS@1.60` supports minimum version `Julia@1.3`. There's a bug fix that needs to be backported to the version supporting `Julia@1`.

Would you not need to tag two versions, `v1.60.1` for the path supporting `Julia@1.3`? And another `v1.60.2` for the path supporting `Julia@1`?

Maybe I'm missing something obvious here, but it seems a bit complicated?

christopher-dG: I think the 1.59.X series of releases would be the backport target.

ericphanson: I think the way to go would be to bump 1.61.0 here (there's a new feature anyway, so it shouldn't be a patch bump), and then the 1.60.x series has its patch versions available for backported bugfixes (without adding new features)

ericphanson: but yeah, if there was a bugfix right away, there'd be a 1.61.1 release, and then a 1.60.1 backport release. We could have a `1.60-release` long-lived branch that we'd merge backported bugfixes to. Definitely some overhead! But maybe it could just be for a month or two until there's a new LTS.

There's a nice blog post on Julia Base's release process btw: https://julialang.org/blog/2019/08/release-process/, might be useful for us

mattBrzezinski: Right! And I guess we won't be backporting features, so that won't matter. Makes sense!

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,106 @@
...
+            D.easy_hook = (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
+            AWS_DOWNLOADER[] = D
+        end
+    end
+    return downloader
+end
+
+# https://github.com/JuliaWeb/HTTP.jl/blob/2a03ca76376162ffc3423ba7f15bd6d966edff9b/src/MessageRequest.jl#L84-L85
+body_length(x::AbstractVector{UInt8}) = length(x)
+body_length(x::AbstractString) = sizeof(x)
```
mattBrzezinski, line 32: Can this not also just be `length("foobar")`?

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,106 @@
...
+    #
+    # When the method is `HEAD`, the response may have a Content-Length
+    # but not send any content back (which appears to be correct,
+    # <https://stackoverflow.com/a/18925736/12486544>).
+    #
+    # Thus, if we did not set `CURLOPT_NOBODY`, and it gets a Content-Length
+    # back, it will hang waiting for that body.
+    #
+    # Therefore, we do not pass an `output` when the `request_method` is `HEAD`.
+    if request.return_stream && request.response_stream === nothing
```
mattBrzezinski, line 51: Not quite understanding the point of this `if` statement. On the line following this block, we're either getting in the set stream, or an `IOBuffer()`?

mattBrzezinski (approved):
Couple questions, but LGTM!

src/utilities/downloads_backend.jl
christopher-dG: Unicode is tricky:

```jl
julia> length("π")
1

julia> sizeof("π")
2
```

I just took the implementation from HTTP.jl.

mattBrzezinski: I knew there was probably a reason for it, lmao.

src/utilities/downloads_backend.jl
christopher-dG: I tried to simplify it a bit now, I think I was trying to avoid mutating `request` when possible but there's not really any benefit and it just gets more complicated.

christopher-dG:
bors try

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,109 @@
...
+    # only when we do not pass the `output` argument.
+    #
+    # When the method is `HEAD`, the response may have a Content-Length
+    # but not send any content back (which appears to be correct,
+    # <https://stackoverflow.com/a/18925736/12486544>).
+    #
+    # Thus, if we did not set `CURLOPT_NOBODY`, and it gets a Content-Length
+    # back, it will hang waiting for that body.
+    #
+    # Therefore, we do not pass an `output` when the `request_method` is `HEAD`.
```
ericphanson, line 50:
```suggestion
    # If we pass `output`, older versions of Downloads.jl will
    # expect a message body in the response. Specifically, it sets
    # <https://curl.se/libcurl/c/CURLOPT_NOBODY.html>
    # only when we do not pass the `output` argument.
    # See <https://github.com/JuliaLang/Downloads.jl/issues/131>.
    #
    # When the method is `HEAD`, the response may have a Content-Length
    # but not send any content back (which appears to be correct,
    # <https://stackoverflow.com/a/18925736/12486544>).
    #
    # Thus, if we did not set `CURLOPT_NOBODY`, and it gets a Content-Length
    # back, it will hang waiting for that body.
    #
    # Therefore, we do not pass an `output` when the `request_method` is `HEAD`.
    # (Note: this is fixed on the latest Downloads.jl, but we include this workaround
    #  for compatability).
```

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    @repeat 4 try
+        downloader = @something(backend.downloader, get_downloader())
+        # set the hook so that we don't follow redirects. Only
+        # need to do this on per-request downloaders, because we
+        # set our global one with this hook already.
+        if backend.downloader !== nothing
+            downloader.easy_hook = (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
+        end
```
ericphanson, line 88: This could be a thread-safety issue if multiple threads are using the same downloader. Though I think multiple threads using the same downloader is already an issue upstream.

This maybe makes things slightly better but doesn't actually fix the race:
```suggestion
        if backend.downloader !== nothing && downloader.easy_hook === nothing
            downloader.easy_hook = (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
        end
```
Not totally sure what we should do here.

ericphanson (commented):
great stuff @christopher-dG! Thanks a ton for picking this up. Left a few comments that are mostly things I meant to think about/do but didn't get to.

test/runtests.jl
```diff
@@ -34,6 +34,8 @@ function _now_formatted()
     return lowercase(Dates.format(now(Dates.UTC), dateformat"yyyymmdd\THHMMSSsss\Z"))
 end
 
+AWS.DEFAULT_BACKEND[] = AWS.DownloadsBackend()
```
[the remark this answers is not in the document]
christopher-dG: Might as well do it via GHA matrix.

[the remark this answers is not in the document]
christopher-dG: Actually on second thought I kind of dislike the idea of only doing "complete" tests in CI, since it makes whatever isn't the default backend second-class, but also this package is rarely tested locally... but I think just double-including should work fine.

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+    return downloader
+end
+
+# https://github.com/JuliaWeb/HTTP.jl/blob/2a03ca76376162ffc3423ba7f15bd6d966edff9b/src/MessageRequest.jl#L84-L85
+body_length(x::AbstractVector{UInt8}) = length(x)
+body_length(x::AbstractString) = sizeof(x)
+
+read_body(x::IOBuffer) = take!(x)
+read_body(x::IO) = readavailable(x)
```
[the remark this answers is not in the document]
christopher-dG: When the `IO` is a `BufferStream`, `read` blocks forever if it's still open. I think we can perhaps do `close(x); read(x)`? Honestly I never know how to handle `IO` in a general way, it seems like every subtype has its own quirks and any mistake causes an infinite hang lol.

test/runtests.jl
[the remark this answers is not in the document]
ericphanson: in AWSS3 to add Minio tests, we refactored the tests to be functions of a config object and then ran those functions with two config objects (https://github.com/JuliaCloud/AWSS3.jl/pull/163). So we could do something like that here... either way seems OK to me

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,109 @@
...
+
+function AWS._http_request(backend::DownloadsBackend, request)
```
christopher-dG, line 37:
```suggestion
function _http_request(backend::DownloadsBackend, request)
```

christopher-dG:
bors try

christopher-dG:
bors try

christopher-dG:
bors try

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,106 @@
...
+        while true
+            downloader = AWS_DOWNLOADER[]
+            downloader isa Downloader && return
+            D = Downloader()
+            D.easy_hook = (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
+            AWS_DOWNLOADER[] = D
+        end
+    end
+    return downloader
+end
```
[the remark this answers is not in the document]
christopher-dG: Perhaps we can leave it until it gets merged upstream

christopher-dG:
Rebased, looking forward to seeing what the formatter has to say 😅

christopher-dG:
Ah, I'm guessing the org-wide token settings are too restrictive. It might also be best to do a separate formatting PR because there are going to be a **lot** of suggestions.

mattBrzezinski:
> Ah, I'm guessing the org-wide token settings are too restrictive. It might also be best to do a separate formatting PR because there are going to be a **lot** of suggestions.

I don't know if it will run on existing PRs. I think only on creation and re-open?

christopher-dG:
It ran, it just failed.

mattBrzezinski:
> It ran, it just failed.
> 
> https://github.com/JuliaCloud/AWS.jl/pull/396/checks?check_run_id=3324750246#step:6:11

Ah! Just missed this, I'll make a new PR to address these.

christopher-dG:
If there's nothing left to do on this, can we merge it before the giant formatting PR goes in? I'd prefer not to deal with the merge conflicts.

mattBrzezinski:
> If there's nothing left to do on this, can we merge it before the giant formatting PR goes in? I'd prefer not to deal with the merge conflicts.

Yep! Makes sense to me!

test/AWS.jl
```diff
@@ -34,6 +34,7 @@ end
     @test AWS.user_agent[] == "AWS.jl/1.0.0"
     set_user_agent(new_user_agent)
     @test AWS.user_agent[] == new_user_agent
+    set_user_agent("AWS.jl/1.0.0")
```
ericphanson, line 37: This should be in a `finally` block so even if the test fails we properly reset the state

ericphanson (commented):
Lgtm! (Modulo one small test fix)

christopher-dG:
bors r+

:)

christopher-dG:
Isn't this supposed to be merged now?

ericphanson:
Bors runs CI and then merges when CI passes

ericphanson:
Maybe it’s waiting for another check and timing out while waiting? (The format check?). Can this just be manually merged?

christopher-dG:
I think the bors config just needed updating to look for the new 1.3 status. Not sure if it'll apply to this PR though, I bet it gets its value from the main branch.

@mattBrzezinski do you have the power to manually merge?

ericphanson:
> I think the bors config just needed updating to look for the new 1.3 status. Not sure if it'll apply to this PR though, I bet it gets its value from the main branch.

It actually sounds like the other way:

> Note: that bors reads bors.toml from the pull requests it’s merging, not the one in master, so changes to the file get checked before they land. You can therefore verify whether bors is working by using bors r+ to try and land the PR that enables it!

https://bors.tech/documentation/getting-started/

bors merge

mattBrzezinski:
You need to modify the bors.toml file to remove 1.0 from it with 1.3

mattBrzezinski:
Sorry I’m running around doing some errands currently.

mattBrzezinski (commented):
bors r+

mattBrzezinski:
Just needed the magic touch! ✨

Project.toml
c42f: It's not so bad.  We did some maintenance branches for StaticArrays quickfixes in the past. It's usually easy to fix the bug in master as usual, then just cherry pick only the fix you need onto the maintenance branch and create a new patch release.

src/utilities/downloads_backend.jl
```diff
@@ -0,0 +1,115 @@
...
+
+# https://github.com/JuliaWeb/HTTP.jl/blob/2a03ca76376162ffc3423ba7f15bd6d966edff9b/src/MessageRequest.jl#L84-L85
+body_length(x::AbstractVector{UInt8}) = length(x)
+body_length(x::AbstractString) = sizeof(x)
+
+read_body(x::IOBuffer) = take!(x)
+function read_body(x::IO)
+    close(x)
+    return read(x)
```
c42f, line 37: Was just reading through this, I don't understand `close(x); read(x)` — for normal files `read(x)` will return zero bytes after it's closed. (For IOBuffer it throws an exception).

src/utilities/downloads_backend.jl
c42f: @ericphanson I think we need per-request hooks to fix this. See https://github.com/JuliaLang/Downloads.jl/issues/141

src/utilities/downloads_backend.jl
christopher-dG: This was a workaround specifically for `Base.BufferStream`, without `close`ing it beforehand, `read` hangs. `readavailable` works as well and that was my first choice, but @ericphanson pointed out this warning:

```
  │ Warning
  │
  │  The amount of data returned is implementation-dependent; for example it can depend on the internal choice of buffer size. Other
  │  functions such as read should generally be used instead.
```

I'm no expert when it comes to the various `IO` types and how to properly handle them all, so any feedback is good feedback!

c42f: I think perhaps we just need to use `read()` here and hope for the best?  It turns out there's a lot of inconsistency in the blocking behavior of streams in Julia — see the issue here https://github.com/JuliaLang/julia/issues/24526

Perhaps for this case, the newly available but strangely named `Base.shutdown()` is what you want https://github.com/JuliaLang/julia/pull/40783

But even so, I'm not sure that AWS should be unilaterally closing or shutting down the write side of the the stream at all — to me that seems to be the business of the code which passes in the stream. For example, if reading from a pipe, presumably you want to keep reading until the writing side decides it's done?

vtjnash: Agreed, seems like whoever was filling the buffer needs to call `shutdown` (or close) to mark it as being done

ericphanson: > I'm not sure that AWS should be unilaterally closing or shutting down the write side of the the stream at all — to me that seems to be the business of the code which passes in the stream.

Well, we only do it if `return_stream` is false, in which case you are supposed to get back some bytes as far as I understand it, and it seems like we need to close it to get those bytes.

> Agreed, seems like whoever was filling the buffer needs to call shutdown (or close) to mark it as being done

I’m not sure who that is- Downloads.jl fills it but I don’t think they should be closing it (because of exactly what @c42f just said.

vtjnash: Well, somebody was wrong before, and now AWS.jl is now possibly also very bad at stream EOF handling here too. Whoever created the stream is generally responsible for marking when it is done, or passing it to another function which will take care of that. I suspect Downloads.jl absolutely should be copying the EOF marker from the underlying stream also. It is out-of-band data, but likely just as relevant as the in-band bytes.

c42f: > I’m not sure who that is- Downloads.jl fills it but I don’t think they should be closing it

Perhaps I've misunderstood what's going on here, but is the problem in Downloads.jl then? Shouldn't they close the write side when writing is done?

ericphanson: I 100% don't understand what's going on. But let me try to summarize the current behavior (since this PR has been merged).

## Summary

The HTTPBackend only sets a `response_stream` if both the user has not passed one and they ask for `return_stream=true`, whereas with the DownloadsBackend we create one whenever the user has not passed one. This is because Downloads.jl wants you to pass an `IO` object to get the response always, unlike HTTP.jl which will either populate a stream you pass or return a vector of bytes directly.

* Case 1: user passes `response_stream`, sets `return_stream=true`
  * neither backend creates a stream, nor closes it ✔️ 
* Case 2: user passes `response_stream`, sets `return_stream=false`
  * HTTP backend passes the stream along; HTTP.jl does not return a body, so the caller does not get back a body but does have their stream populated. No one closes the stream. Seems correct ✔️
  * Downloads backend passes the stream along, but closes the stream and collects the bytes to return. Seems wrong ❌ 
* Case 3: user does not pass `response_stream`, sets `return_stream=true`
  * Both backends create a stream and return it; neither closes it. Not sure if this is correct or not ❓ 
* Case 4: user does not pass `response_stream`, and sets `return_stream=false`
  * HTTPBackend does not create a stream, just gets the bytes from the `body` HTTP.jl returns ✔️ 
  * DownloadsBackend internally creates a stream, then closes it and collects the bytes to return ✔️ 

If I'm understanding then, I think Case 2 is probematic, and we should just not try to return bytes in this case, matching the behavior of the HTTP backend. I.e. we should not be closing the stream in this case.

Case 3: I don't really know if we should close the stream or not. I think ideally this case wouldn't exist, because I don't think there really *should* be a `return_stream` option-- I think a simpler API is "if you want a stream you should pass one and it will get populated (and not closed); if you don't want a stream, don't pass one and you will get a vector of bytes back".

So: if that analysis is right, Downloads.jl is fine, and we have at least one problematic case here. But I'll admit to definitely not fully understanding the semantics of how streams should work (and [the manual](https://docs.julialang.org/en/v1/manual/networking-and-streams/) isn't very helpful in this regard), though the discussion here helps me a bit.

### Appendix: Following the path of `get_object`

To see these cases, play yout, we let's say we are doing an s3 `get_object`.

* We start https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/services/s3.jl#L2239-L2246 and possibly submit `params` which is a dictionary with additional parameters
* we use those `params` to build ` Request`, changing the name from `params` to `args`: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/AWS.jl#L213-L217
* we populate with default values: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/utilities.jl#L65-L77
* we end up with our `Request` object: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/request.jl#L26-L42
* We pass it to `submit_request`: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/request.jl#L59-L61

Here, what happens depends on which backend we are using. This PR is about the new DownloadsBackend.

### Using the `DownloadsBackend`

* we send the `Request` to `_http_request` and if we are using a `DownloadsBackend`, we end up at the method in question, https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/downloads_backend.jl#L41
* Here, if user has not passed a `response_stream`, we create a default one: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/downloads_backend.jl#L58-L60
* And if they have not specified `return_stream==true`, we arrange to collect the bytes from the response stream, including possibly closing the `response_stream`: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/downloads_backend.jl#L70-L74

### Using the `HTTPBackend`

If the user is using the existant HTTP.jl backend, what happens?

* When we get to `_http_request`, we get to this method: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/request.jl#L186-L217
* There we create a `response_stream` if both the user has not passed one *and* they ask for `return_stream=true`: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/request.jl#L192-L194
* We pass the possibly-`nothing` `request.response_stream` to `HTTP.request`: https://github.com/JuliaCloud/AWS.jl/blob/181c340fa06809a578b6c1c0aa30cc8267029eed/src/utilities/request.jl#L203
* HTTP.jl will give us back a `body` if they were *not* given a `response_stream`, otherwise they put the bytes in the stream: https://github.com/JuliaWeb/HTTP.jl/blob/a2b467e24c9bbd45691f9d0f57b57ee7463bd15a/src/HTTP.jl#L77-L78

PR #401, opened 2021-07-13
provide hooks for custom backends

ericphanson:
Instead of adding a whole new backend as in #396, I thought it would be better to just start with an `AbstractBackend` to provide hooks to use alternate backends. This way we don't need to bump the Julia requirements to start trying out alternate backends downstream. It also provides a smaller incremental step than having AbstractBackends + a specific alternate backend in the same PR.

This also provides what is hopefully a useful feature: you could do
```julia
using AWS
AWS.DEFAULT_BACKEND[] = HTTPBackend(custom_http_options)
```
to specify custom HTTP options globally in your application (which can still be overwritten per-request); I would have found this useful when trying experiments like tweaking the pipelining and connection pooling settings.

In the same way, a Downloads.jl-based backend could pass a `Downloads.Downloader` along inside the backend object instead of requiring adding more fields to Request for the `downloader` like in #396.

### DIFF
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -8,9 +8,11 @@ An `HTTPBackend` can hold default `http_options::AbstractDict{Symbol,<:Any}`
 to pass to HTTP.jl, which can be overwritten per-request by any `http_options`
 supplied there.
 """
-Base.@kwdef struct HTTPBackend <: AbstractBackend
-    http_options::AbstractDict{Symbol,<:Any}=LittleDict{Symbol,String}()
+struct HTTPBackend{T<:AbstractDict{Symbol,<:Any}} <: AbstractBackend
+    http_options::T
 end
+
+HTTPBackend() = HTTPBackend(LittleDict{Symbol,String}())
 
 const DEFAULT_BACKEND = Ref{Union{Nothing, AbstractBackend}}(HTTPBackend())
 
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -14,7 +14,7 @@ end
 
 HTTPBackend() = HTTPBackend(LittleDict{Symbol,String}())
 
-const DEFAULT_BACKEND = Ref{Union{Nothing, AbstractBackend}}(HTTPBackend())
+const DEFAULT_BACKEND = Ref{AbstractBackend}(HTTPBackend())
 
 Base.@kwdef mutable struct Request
     service::String
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -16,7 +16,8 @@ function HTTPBackend(; kwargs...)
     isempty(kwargs) ? HTTPBackend(LittleDict{Symbol, Any}()) : HTTPBackend(LittleDict(kwargs))
 end
 
-const DEFAULT_BACKEND = Ref{AbstractBackend}(HTTPBackend())
+# populated in `__init__`
+const DEFAULT_BACKEND = Ref{AbstractBackend}()
 
 Base.@kwdef mutable struct Request
     service::String
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -12,7 +12,9 @@ struct HTTPBackend <: AbstractBackend
     http_options::AbstractDict{Symbol,<:Any}
 end
 
-HTTPBackend(; kwargs...) = HTTPBackend(LittleDict(kwargs))
+function HTTPBackend(; kwargs...)
+    isempty(kwargs) ? HTTPBackend(LittleDict{Symbol, Any}()) : HTTPBackend(LittleDict(kwargs))
+end
 
 const DEFAULT_BACKEND = Ref{AbstractBackend}(HTTPBackend())
 
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -12,7 +12,7 @@ struct HTTPBackend <: AbstractBackend
     http_options::AbstractDict{Symbol,<:Any}
 end
 
-HTTPBackend() = HTTPBackend(LittleDict{Symbol,String}())
+HTTPBackend(; kwargs...) = HTTPBackend(LittleDict(kwargs))
 
 const DEFAULT_BACKEND = Ref{AbstractBackend}(HTTPBackend())
 
@@ -170,7 +170,7 @@ function _http_request(http_backend::HTTPBackend, request::Request)
             request.response_stream = Base.BufferStream()
         end
 
-        return HTTP.request(
+        return @mock HTTP.request(
             http_stack,
             request.request_method,
             HTTP.URI(request.url),
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -1,4 +1,4 @@
-# this is used to allow custom dispatches to `_http_request`
+# Used to allow custom dispatches to `_http_request`
 abstract type AbstractBackend end
 
 """
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -1,3 +1,19 @@
+# this is used to allow custom dispatches to `_http_request`
+abstract type AbstractBackend end
+
+"""
+    AWS.HTTPBackend <: AWS.AbstractBackend
+
+An `HTTPBackend` can hold default `http_options::AbstractDict{Symbol,<:Any}`
+to pass to HTTP.jl, which can be overwritten per-request by any `http_options`
+supplied there.
+"""
+Base.@kwdef struct HTTPBackend <: AbstractBackend
+    http_options::AbstractDict{Symbol,<:Any}=LittleDict{Symbol,String}()
+end
+
+const DEFAULT_BACKEND = Ref{Union{Nothing, AbstractBackend}}(HTTPBackend())
+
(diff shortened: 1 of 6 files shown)


ericphanson:
This still needs tests for the new functionality but I'll kick off bors to see if I correctly updated the tests.

bors try

ericphanson:
bors try

src/utilities/request.jl
```diff
@@ -1,3 +1,21 @@
...
+"""
+    AWS.HTTPBackend <: AWS.AbstractBackend
+
+An `HTTPBackend` can hold default `http_options::AbstractDict{Symbol,<:Any}`
+to pass to HTTP.jl, which can be overwritten per-request by any `http_options`
+supplied there.
+"""
+struct HTTPBackend{T<:AbstractDict{Symbol,<:Any}} <: AbstractBackend
+    http_options::T
+end
```
ericphanson, line 13: wasn't sure if in this case it's better to do this or just
```suggestion
struct HTTPBackend <: AbstractBackend
    http_options::AbstractDict{Symbol,<:Any}
end
```
which I think matches the choices made elsewhere more. Leaving the type abstract means we pay for dynamic dispatch when we merge these options with the Request's options-- but the Request has abstractly typed fields anyway so maybe we need to do that regardless. Probably in the context of network operations, the microseconds of dynamic dispatch do not really matter at all... And it should need less compile time to make it abstract.

src/utilities/request.jl
```diff
@@ -1,3 +1,21 @@
...
+to pass to HTTP.jl, which can be overwritten per-request by any `http_options`
+supplied there.
+"""
+struct HTTPBackend <: AbstractBackend
+    http_options::AbstractDict{Symbol,<:Any}
+end
+
+HTTPBackend() = HTTPBackend(LittleDict{Symbol,String}())
+
+const DEFAULT_BACKEND = Ref{Union{Nothing, AbstractBackend}}(HTTPBackend())
```
omus, line 17: What scenario would this need to be set to `nothing`?

test/patch.jl
```diff
@@ -56,7 +56,7 @@ end
 
 
 function _aws_http_request_patch(response::HTTP.Messages.Response=_response())
-    return @patch AWS._http_request(request::Request) = response
+    return @patch AWS._http_request(::Any, request::Request) = response
```
omus, line 59: Not necessary but clearer
```suggestion
    return @patch AWS._http_request(::AWS.HTTPBackend, request::Request) = response
```

src/utilities/request.jl
ericphanson: Good point, probably never

ericphanson:
bors try

ericphanson:
bors try

src/utilities/request.jl
```diff
@@ -1,3 +1,23 @@
+# this is used to allow custom dispatches to `_http_request`
```
omus, line 1:
```suggestion
# Used to allow custom dispatches to `_http_request`
```

src/utilities/request.jl
```diff
@@ -1,3 +1,23 @@
...
+"""
+struct HTTPBackend <: AbstractBackend
+    http_options::AbstractDict{Symbol,<:Any}
+end
+
+function HTTPBackend(; kwargs...)
+    isempty(kwargs) ? HTTPBackend(LittleDict{Symbol, Any}()) : HTTPBackend(LittleDict(kwargs))
+end
+
+const DEFAULT_BACKEND = Ref{AbstractBackend}(HTTPBackend())
```
omus, line 19: I think we're fine in this case but sometimes to work best with PackageCompiler you want to initialize inside the `__init__`. Just calling this out if you discover some weird PackageCompiler issue with swapping out the backend

src/utilities/utilities.jl
```diff
@@ -69,6 +69,7 @@ function _extract_common_kw_args(service, args)
         headers=LittleDict{String, String}(_pop!(args, "headers", [])),
         http_options=_pop!(args, "http_options", LittleDict{Symbol, String}()),
         response_dict_type=_pop!(args, "response_dict_type", LittleDict),
+        backend=_pop!(args, "backend", DEFAULT_BACKEND[]),
```
omus, line 72: Hold on, are you planning on passing in an alternate backend via a keyword? In that case should we instead have
```julia
const DEFAULT_BACKEND = HTTPBackend()
```
?

omus: I see now from the tests you're planning on mutating the `Ref`. At least one test not using the default value here should be added.

test/AWS.jl
```diff
@@ -351,6 +351,50 @@ end
...
+            api_version="api_version",
+            request_method="GET",
+            url="https://s3.us-east-1.amazonaws.com/sample-bucket",
+        )
+        @test request.backend.http_options == Dict(:connection_limit => 3)
+        apply(Patches._http_options_patch) do
+            @test AWS._http_request(request.backend, request) == Dict(:connection_limit => 3)
+        end
+    finally
+        AWS.DEFAULT_BACKEND[] = AWS.HTTPBackend()
```
omus, line 394: Would be better to save the original backend and restore that one

test/AWS.jl
```diff
@@ -351,6 +351,50 @@ end
...
+        @test AWS._http_request(custom_backend, request) == Dict(:pipeline_limit => 20, :connection_limit => 5)
+
+        # per-request options override backend options:
+        custom_backend = AWS.HTTPBackend(Dict(:pipeline_limit => 5))
+        @test AWS._http_request(custom_backend, request) == Dict(:pipeline_limit => 20)
+    end
+
+    # Let's test setting the default backend
+    try
+        AWS.DEFAULT_BACKEND[] = AWS.HTTPBackend(; connection_limit = 3)
```
omus, line 382: Best if you define a new backend type and use that as part of the tests.

test/patch.jl
```diff
@@ -56,7 +56,7 @@ end
 
 
 function _aws_http_request_patch(response::HTTP.Messages.Response=_response())
-    return @patch AWS._http_request(request::Request) = response
+    return @patch AWS._http_request(::AWS.HTTPBackend, request::Request) = response
```
omus, line 59: Patch should work on any backend
```suggestion
    return @patch AWS._http_request(::AWS.AbstractBackend, request::Request) = response
```

test/patch.jl
```diff
@@ -73,7 +73,7 @@ _web_identity_patch = function (;
     session_token="web_session_token",
     role_arn="arn:aws:sts:::assumed-role/role-name",
 )
-    @patch function AWS._http_request(request)
+    @patch function AWS._http_request(::AWS.HTTPBackend, request)
```
omus, line 76:
```suggestion
    @patch function AWS._http_request(::AWS.AbstractBackend, request)
```

src/utilities/request.jl
ericphanson: Ah ok, like re-initialize inside the `__init__`, or add back a `nothing` option and only initialize inside `__init__`?

src/utilities/utilities.jl
ericphanson: My thought was to allow both, i.e. the backend can be overloaded per-request or set globally. E.g. the Downloads backend from #396 currently has edge cases around some of the requests made by `STS`, but works very well for simple `GET`s like for s3 access. So I'd like to set the backend per-request in my s3 requests, but I'd also like to set the backend globally to polish/test the Downloads backend until it works everywhere and could be the new default.

But even if there was only 1 backend (say the HTTPBackend we have now), I'd still like it to be overridable globally because you could change some global options (like a higher connection or pipelining limit).

src/utilities/request.jl
omus: Initializing inside of `__init__` and just leave this undefined:
```suggestion
const DEFAULT_BACKEND = Ref{AbstractBackend}()
```

test/AWS.jl
ericphanson: ok, done

test/AWS.jl
ericphanson: good point, done

ericphanson:
bors try

src/utilities/request.jl
ericphanson: ahh, didn't know one could do that, thanks!

src/utilities/utilities.jl
omus: I think you forgot to add the test for this.

ericphanson: Ah maybe I misunderstood. In https://github.com/JuliaCloud/AWS.jl/pull/401/commits/22974c952c1dd51796ce4da55c41b3ee9659fe48 I added a TestBackend. Did you mean a different test than that?

omus: I was thinking of a test that passed in a custom backend using the `backend` keyword

ericphanson: Ahh I see, added!

ericphanson:
bors try

omus (approved):

ericphanson:
Awesome, thanks! Can we get a release with it when it goes through? Excited to use this downstream

PR #402, opened 2021-07-14
add ':' to list of special S3 escapes

### LINKED ISSUE #387: S3 URIs not getting escaped?
Hi,
Recently I encountered an S3 bug where it doesn't escape certain characters in the Key which leads to AWS reporting the following error -
> Encountered exception : AWSException: SignatureDoesNotMatch -- The request signature we calculated does not match the signature you provided. Check your key and signing method.  
  
The Key contained `:` (eg `18T18:16:54.json`) which I had to manually escape before asking for the file to be downloaded; I'm worried if this was forgotten or is it intended to be this way?

wytbella:
close #387

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.49.0"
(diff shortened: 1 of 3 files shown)


mattBrzezinski (approved):
bors r+

Issue #408, opened 2021-07-24
SESv2 using the wrong endpoint


jsahil730:
SESv2 tries to access `ses.<region>.amazonaws.com` while instead the endpoints for this api are supposed to be under `email.<region>.amazonaws.com`; therefore a DNSError is observed with the request.

```
ERROR: MethodError: no method matching send_email(::Dict{Any, Any}, ::Dict{Any, Any})
Closest candidates are:
  send_email(::Any; aws_config) at C:\Users\DELL\.julia\packages\AWS\MbHfM\src\services\sesv2.jl:1465
  send_email(::Any, ::AbstractDict{String, V} where V; aws_config) at C:\Users\DELL\.julia\packages\AWS\MbHfM\src\services\sesv2.jl:1466
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1

julia> SESv2.send_email(Dict())
ERROR: DNSError: ses.ap-south-1.amazonaws.com, unknown node or service (EAI_NONAME)
Stacktrace:
  [1] getalladdrinfo(host::String)
    @ Sockets C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Sockets\src\addrinfo.jl:113
  [2] getalladdrinfo
    @ C:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.6\Sockets\src\addrinfo.jl:122 [inlined]
  [3] getconnection(::Type{Sockets.TCPSocket}, host::SubString{String}, port::String; keepalive::Bool, connect_timeout::Int64, readtimeout::Int64,
 kw::Base.Iterators.Pairs{Symbol, Union{Nothing, Bool}, Tuple{Symbol, Symbol, Symbol}, NamedTuple{(:require_ssl_verification, :iofunction, :response_stream), Tuple{Bool, Nothing, Nothing}}})
    @ HTTP.ConnectionPool C:\Users\DELL\.julia\packages\HTTP\4AvE2\src\ConnectionPool.jl:659
  [4] #getconnection#29
    @ C:\Users\DELL\.julia\packages\HTTP\4AvE2\src\ConnectionPool.jl:728 [inlined]
[...]

PR #415, opened 2021-08-09
Use proper service names

mattBrzezinski:
### Resolves: #335, #408

This is an addendum to #336, as it didn't really fix the issue. Currently we are trying to use one piece of information to both create the URL which to hit and sign the request. This is not enough. We need to have both the `endpointPrefix` and the `signingName`.

This PR creates functionality to store both of these pieces of information and to use them appropriate. I should note that not all services have a `signingName`. In these cases we will use the `endpointPrefix` in lieu.


mattBrzezinski:
bors try

mattBrzezinski:
bors try

src/api_generation/utilities.jl
```diff
@@ -239,3 +239,11 @@ function _get_function_parameters(input::String, shapes::AbstractDict{String})
...
     return (sort(required_parameters), sort(optional_parameters))
 end
+
+function _get_service_name(service::AbstractDict)
+    if haskey(service, "signingName") && service["signingName"] == service["endpointPrefix"]
+        return service["signingName"]
+    end
+
+    return service["endpointPrefix"]
+end
```
glennmoy, line 249: why do we need this function?
why not just return `service["endpointPrefix"]` in all cases?

mattBrzezinski: That makes sense... I actually just found an issue here. I'm revising it currently. Ignore this PR for the moment!

mattBrzezinski:
Okay. Revised this, and we're set.

bors try

src/AWS.jl
```diff
@@ -135,31 +135,35 @@ macro service(module_name::Symbol)
...
-    name::String
+    signing_name::String
+    endpoint_prefix::String
     api_version::String
 
     service_specific_headers::LittleDict{String, String}
 end
 
-RestJSONService(name::String, api_version::String) = RestJSONService(name, api_version, LittleDict{String, String}())
+RestJSONService(signing_name::String, endpoint_prefix::String, api_version::String) = RestJSONService(signing_name, endpoint_prefix, api_version, LittleDict{String, String}())
```
iamed2, line 166: Could this be a kwarg? Could make src/AWSServices.jl cleaner. 

```julia
RestJSONService(signing_name::String, api_version::String; endpoint_prefix::String=signing_name) = RestJSONService(signing_name, endpoint_prefix, api_version, LittleDict{String, String}())
```

glennmoy:
do you need to bump the patch version for this as well?

mattBrzezinski:
> do you need to bump the patch version for this as well?

Yes. I tend to do those in separate PRs as in the past I've stepped on toes and ended up accidentally trying to skip versions which JuliaRegistrator gets upset about.

iamed2 (approved):

PR #420, opened 2021-08-12
Make sure region is us-east-1 for signing tests

christopher-dG:
Currently those tests will fail if your default AWS config profile sets the region to anything but us-east-1.


mattBrzezinski (approved):
bors r+

PR #425, opened 2021-08-13
Added formatting for auto-generation

mattBrzezinski:
I added in auto-formatting for code in https://github.com/JuliaCloud/AWS.jl/pull/423, this will become problematic going forward when code is auto-generated each evening as you'll need to step through accepting all changes as it is breaking BlueStyle.

This PR address' that by adding in an additional step to the `update_apis.yml` action to automatically format the code before creating the merge request.

### DIFF
--- a/.github/workflows/update_apis.yml
+++ b/.github/workflows/update_apis.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        julia-version: [1] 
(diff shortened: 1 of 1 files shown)


fchorney:
does https://github.com/JuliaCloud/AWS.jl/pull/423 check for code style, or try and format the code itself in an action?

fchorney (approved):

mjram0s (approved):

PR #427, opened 2021-08-13
Bump version to 1.57.0

christopher-dG:
This will mark the end of 1.0 support; any backports should be branched off of 1.56.0 and released as 1.56.x.


christopher-dG:
@mattBrzezinski I'll let you pull the trigger on this if you agree with the plan above!

PR #429, opened 2021-08-16
Follow redirects

### LINKED ISSUE #397: Redirect handling not quite right?
In https://github.com/JuliaCloud/AWS.jl/pull/389 I added a throw if we could not set the URL from the returned location, but did not change the behavior otherwise.

But I'm not quite sure that original behavior was right: it seems like we set the URL in the request, but if nothing throws, we just continue on our way. I assumed that we would automatically retry with the new URL, but that was a misunderstanding of [`@retry`](https://github.com/JuliaWeb/Retry.jl#repeat-n-try). If there is no error, it does not retry. So I don't think setting the URL does much, since we don't use it to make another request.

I'm not really sure though when we will get a redirect from AWS that we want to follow; when you have the wrong region, we get 302s (without a `location` parameter sent back), but I'm not sure in what other circumstances it happens. On my #396 branch I have seen this code path triggered by
```julia
AWSServices.sts(
                             "AssumeRoleWithWebIdentity",
                             Dict(
                                 "RoleArn" => role_arn, # an actual role arn
                                 "RoleSessionName" => "test",
                                 "WebIdentityToken" => "123", # made up
                             );
                             aws_config=AWSConfig(creds=nothing)
```
but I don't know why-- AWS should be returning an error code about the WebIdentityToken being bad, not a redirect, like it does on master.

Also, I'm not sure why we don't use `redirect=true` in our HTTP stack, if we do want to follow redirects.

cc @omus

### LINKED ISSUE #413: Redirect errors should say the region it is redirecting *from*
E.g. `BUCKET` is defined in `us-east-2` and if I try to read something from it while my config is set to `us-east-1`, I get the error:
```julia
ERROR: AWSException: PermanentRedirect -- The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.
HTTP.ExceptionRequest.StatusError(301, "GET", "/BUCKET/", HTTP.Messages.Response:
"""
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: us-east-2
x-amz-request-id: XXX
x-amz-id-2: XXX
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Mon, 09 Aug 2021 12:24:58 GMT
Server: AmazonS3

<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint.</Message><Endpoint>BUCKET.s3.us-east-2.amazonaws.com</Endpoint><Bucket>BUCKET</Bucket><RequestId>XXX</RequestId><HostId>XXX</HostId></Error>""")
```
This helpfully says to use `us-east-2`, but if I think I'm already using `us-east-2` then I don't see what the problem is. It should also say somewhere that I'm actually on `us-east-1` so I can see the discrepency.

This also would be helpful to *rule out* region issues if one is getting a redirect for some other reason. The error message alone doesn't give enough info to know it it's a region issue or not.

from convo with @haberdashpi

christopher-dG:
Closes #397 and closes #413

I'm opening this more to start a conversation than to actually get it merged right away: Why don't we currently follow redirects? I would imagine there is some underlying reason to not just use the default HTTP behaviour. As mentioned in the issues above, it's caused some annoyances.

### DIFF
--- a/src/utilities/downloads_backend.jl
+++ b/src/utilities/downloads_backend.jl
@@ -4,30 +4,6 @@ end
 
 DownloadsBackend() = DownloadsBackend(nothing)
 
-const AWS_DOWNLOADER = Ref{Union{Nothing,Downloader}}(nothing)
-const AWS_DOWNLOAD_LOCK = ReentrantLock()
-
-# Here we mimic Download.jl's own setup for using a global downloader.
-# We do this to have our own downloader (separate from Downloads.jl's global downloader)
-# because we add a hook to avoid redirects in order to try to match the HTTPBackend's
-# implementation, and we don't want to mutate the global downloader from Downloads.jl.
-# https://github.com/JuliaLang/Downloads.jl/blob/84e948c02b8a0625552a764bf90f7d2ee97c949c/src/Downloads.jl#L293-L301
-function get_downloader(downloader=nothing)
-    lock(AWS_DOWNLOAD_LOCK) do
-        yield() # let other downloads finish
-        downloader isa Downloader && return nothing
-        while true
-            downloader = AWS_DOWNLOADER[]
-            downloader isa Downloader && return nothing
-            D = Downloader()
-            D.easy_hook =
-                (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
-            AWS_DOWNLOADER[] = D
-        end
-    end
-    return downloader
-end
-
 # https://github.com/JuliaWeb/HTTP.jl/blob/2a03ca76376162ffc3423ba7f15bd6d966edff9b/src/MessageRequest.jl#L84-L85
 body_length(x::AbstractVector{UInt8}) = length(x)
 body_length(x::AbstractString) = sizeof(x)
@@ -86,15 +62,6 @@ function _http_request(backend::DownloadsBackend, request)
     end
 
     @repeat 4 try
-        downloader = @something(backend.downloader, get_downloader())
-        # set the hook so that we don't follow redirects. Only
-        # need to do this on per-request downloaders, because we
-        # set our global one with this hook already.
-        if backend.downloader !== nothing && downloader.easy_hook === nothing
-            downloader.easy_hook =
-                (easy, info) -> Curl.setopt(easy, Curl.CURLOPT_FOLLOWLOCATION, false)
-        end
-
         # We seekstart on every attempt, otherwise every attempt
         # but the first will send an empty payload.
         seekstart(input)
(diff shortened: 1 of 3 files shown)


christopher-dG:
bors try

christopher-dG:
Unrelated to this PR, but the failure is interesting, the body that ended up being returned was this error message from AWS:

```xml
<Error>
  <Code>InternalError</Code>
  <Message>We encountered an internal error. Please try again.</Message>
  <RequestId>S25XVXJ5MWF71587</RequestId>
  <HostId>2vDzv6+vhBseugv0qFDNQkcx6rKgq2lbFzRVzWiQ2Dx/yQ9R8xQ6m0S00eJ8l5bpharuU9pTjEE=</HostId>
</Error>
```

This seems like it should have caused an error to be thrown.

omus:
I took a look through the AWS.jl history and it's predecessor AWSCore.jl and I believe we don't follow redirects just because that was the way it always was. I believe this is the first commit of AWSCore.jl that disabled redirects: https://github.com/JuliaCloud/AWSCore.jl/blob/1ab6d6dcefd6a66a62ec16221f25cb5946cb05d0/src/http.jl#L31

Before this it seems the HTTP.jl package wasn't used.

omus:
I took a quick look at botocore and based upon this test we may need to do some special handling around redirects, specifically for S3 bucket regions: https://github.com/boto/botocore/blob/23ee17f5446c78167ff442302471f9928c3b4b7c/tests/integration/test_s3.py#L1217

mattBrzezinski:
> I took a look through the AWS.jl history and it's predecessor AWSCore.jl and I believe we don't follow redirects just because that was the way it always was. I believe this is the first commit of AWSCore.jl that disabled redirects: https://github.com/JuliaCloud/AWSCore.jl/blob/1ab6d6dcefd6a66a62ec16221f25cb5946cb05d0/src/http.jl#L31
> 
> Before this it seems the HTTP.jl package wasn't used.

Yes, I just kept this going for historical reasons.

> I took a quick look at botocore and based upon this test we may need to do some special handling around redirects, specifically for S3 bucket regions: https://github.com/boto/botocore/blob/23ee17f5446c78167ff442302471f9928c3b4b7c/tests/integration/test_s3.py#L1217

I briefly looked into this in the past, and S3 definitely will want special casings for redirects into different regions.

christopher-dG:
Welp, AWS doesn't set the `Location` header for 301 responses, they put it in the `Endpoint` key of the response body XML, which prevents HTTP's automatic redirect handling from working. Maybe that's why it was disabled way back.

It can be worked around but I think it'll be a bit ugly.

PR #430, opened 2021-08-19
mfa support

SimonDanisch:
Rebase code from: https://github.com/JuliaCloud/AWS.jl/pull/357 
cc: @omus


omus:
I've updated #357 which is now ready for review

PR #443, opened 2021-09-08
Avoid using `ecode`

omus:
Prefer accessing the `code` field from `AWSException` instead of relying on the `ecode` function defined in Retry.jl:

https://github.com/JuliaWeb/Retry.jl/blob/b18e14cce8a3e06948c232b9a54737b1dc0093b1/src/Retry.jl#L8

Seems like an unnecessary abstraction to me.


omus:
bors try

mattBrzezinski (approved):

PR #457, opened 2021-09-24
Support specifying features with `@service`

omus:
While working on #384 I encountered a situation where the change I was making was breaking so in order to have to make a major release I wanted to make the change opt-in. Unfortunately doing that would result in every single AWS.jl API call being modified to opt-in to the new behaviour making updating quite tedious. 

The `@service` feature support added in this PR allows end users to opt-in to new behaviour when using the `@service` macro (e.g. `@service S3 my_new_feature=true`). This interface is much more convenient and probably is a pattern we want to use for incorporating future behaviour. Currently this PR defines no features. Additionally for low-level API calls I've introduced the `set_features` function (e.g. `set_features(::Service; my_new_feature=true)`) which allows for setting the feature set to be used for a service instance.

The only issue I can think of with this is if an end user makes use of multiple `@service` calls to the same service in the same module then the last call will define the behaviour. As doing this will result in a module being redefined I don't think this is a situation anyone will notice.

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -41,6 +41,8 @@ using ..AWSExceptions: AWSException
 const user_agent = Ref("AWS.jl/1.0.0")
 const aws_config = Ref{AbstractAWSConfig}()
 
+Base.@kwdef struct FeatureSet end
+
 """
     global_aws_config()
 
@@ -122,12 +124,17 @@ using AWS: @service
 # Return
 - `Expression`: Base.include() call to introduce the high-level service API wrapper functions in your namespace
 """
-macro service(module_name::Symbol)
+macro service(module_name::Symbol, features...)
     service_name = joinpath(@__DIR__, "services", lowercase(string(module_name)) * ".jl")
+    map(_assignment_to_kw!, features)
 
-    return Expr(:toplevel, :(module ($(esc(module_name)))
-    Base.include($(esc(module_name)), $(esc(service_name)))
-    end))
+    module_block = quote
+        using AWS: FeatureSet
+        const SERVICE_FEATURE_SET = FeatureSet(; $(features...))
+        include($service_name)
+    end
+
+    return Expr(:toplevel, Expr(:module, true, esc(module_name), esc(module_block)))
 end
 
 struct RestXMLService
@@ -207,6 +214,7 @@ function (service::RestXMLService)(
     request_uri::String,
     args::AbstractDict{String,<:Any}=Dict{String,Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
+    feature_set::FeatureSet=FeatureSet(),
 )
     return_headers = _pop!(args, "return_headers", false)
 
@@ -259,6 +267,7 @@ function (service::QueryService)(
     operation::String,
     args::AbstractDict{String,<:Any}=Dict{String,Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
+    feature_set::FeatureSet=FeatureSet(),
 )
     POST_RESOURCE = "/"
     return_headers = _pop!(args, "return_headers", false)
@@ -301,6 +310,7 @@ function (service::JSONService)(
     operation::String,
     args::AbstractDict{String,<:Any}=Dict{String,Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
+    feature_set::FeatureSet=FeatureSet(),
 )
     POST_RESOURCE = "/"
     return_headers = _pop!(args, "return_headers", false)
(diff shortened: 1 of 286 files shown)


omus:
bors try

omus:
There's a expression corner case on Julia 1.4 and below which I didn't capture in my tests.

omus:
bors try

omus:
Last change just fixes formatting so I won't bother re-running bors

omus:
I can apply the fixups prior to merging once approved

src/AWS.jl
```diff
@@ -207,6 +212,7 @@ function (service::RestXMLService)(
     request_uri::String,
     args::AbstractDict{String,<:Any}=Dict{String,Any}();
     aws_config::AbstractAWSConfig=global_aws_config(),
+    features::NamedTuple=DEFAULT_SERVICE_FEATURES,
```
omus, line 215: The intent here is to always have a `NamedTuple` which always contains the fields for all of the known features. This allows code internally to do `features.my_feature` instead of `get(feature, :my_feature, DEFAULT_SERVICE_FEATURES.my_feature)` or call some convenience function.

The only downside with the current approach is that if an end user wants pass in their own set of features for a specific call they need to define entries for all of the supported features.

omus:
bors try

omus:
In experimenting with this in #458 it's probably sensible to have the low-level API also make use of features

omus:
> In experimenting with this in #458 it's probably sensible to have the low-level API also make use of features

As the low-level API is from `AWSServices.jl` which is included in `AWS.jl` I can't use the same trick here. Not a big deal but that does bring up an issue for #384: https://github.com/JuliaCloud/AWS.jl/pull/384#issuecomment-926845069.

omus:
> In experimenting with this in #458 it's probably sensible to have the low-level API also make use of features

I found a decent approach to this problem: https://github.com/JuliaCloud/AWS.jl/pull/458/commits/94f7f563aa7ffaa83f647d940295453fff8a62aa. I'll bring that change into this PR once I get it all working

omus:
Rebasing fixups

omus:
bors try

src/api_generation/high_level.jl
```diff
@@ -144,28 +144,28 @@ function _generate_high_level_definition(
 
         if required_keys && (idempotent || headers)
             return """
-                $formatted_function_name($(join(req_keys, ", ")); aws_config::AbstractAWSConfig=global_aws_config()) = $service_name(\"$method\", \"$request_uri\", $params_headers_str; aws_config=aws_config)
-                $formatted_function_name($(join(req_keys, ", ")), params::AbstractDict{String}; aws_config::AbstractAWSConfig=global_aws_config()) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, $params_headers_str, params)); aws_config=aws_config)
+                $formatted_function_name($(join(req_keys, ", ")); aws_config::AbstractAWSConfig=global_aws_config()) = $service_name(\"$method\", \"$request_uri\", $params_headers_str; aws_config=aws_config, features=SERVICE_FEATURES)
+                $formatted_function_name($(join(req_keys, ", ")), params::AbstractDict{String}; aws_config::AbstractAWSConfig=global_aws_config()) = $service_name(\"$method\", \"$request_uri\", Dict{String, Any}(mergewith(_merge, $params_headers_str, params)); aws_config=aws_config, features=SERVICE_FEATURES)
```
omus, line 148: An alternative approach would be to define the following right after the generated `using` statements:
```julia
const $service_name = set_features(AWS.AWSServices.$service_name; SERVICE_FEATURES...)
```
This would reduce the changes to the generated code at the cost of having to unpack the `ServiceWrapper`. Probably the current approach in the PR is the right one.

src/AWS.jl
omus: One solution to this problem would be to make a custom `Features` struct that defined all of the possible features. Then we could do `features::Features` here which would enforce all of the properties are defined.

omus:
bors try

omus:
I'll mention that #460 wouldn't play well with the high-level service features as this PR currently implements them. I think the way these two things could compose is that AWS.jl would have the high-level services as submodules. If you want to use the high-level services with non-default features you would need to use the `@service` macro (may be a macro with a different) to do that.

mattBrzezinski:
Taking this out of our discussion from Slack before it gets lost into the void. This seems like a good step forward with allowing for breaking changes to occur in this package without releasing a new major version. When we eventually get to N amount of breaking changes, or just choose to make a new release it should be easy enough to incorporate. Especially since work in this package is kind of sporadic.

The code itself LGTM! The trade off of more generated code vs. unpacking makes sense, no one will be looking at the generated code regardless.

mattBrzezinski (approved):

omus:
Rebased to resolve conflict and I squashed some commits

PR #459, opened 2021-09-26
Use different file extension for generated files

omus:
The GitHub PR review interface allows you to filter files based upon an extension. This PR is an attempt to use an alternate file extension for generated files to make them easier to filter out. In order to make this work I needed to regenerate the API service definitions.


omus:
Using a `.gen.jl` isn't recognized as a file extension by GitHub: it sees `.jl` (which is reasonable). This is unfortunate as using another file extension means we'll lose syntax highlighting

omus:
Was worth a shot

Issue #466, opened 2021-09-28
`AWS._http_request` broken when using HTTP.jl v0.9.15


nickrobinson251:
Change: https://github.com/JuliaWeb/HTTP.jl/pull/752
> This could be considered a breaking change for users who use the
internal IO type Base.BufferStream with the response_stream keyword.

AWS code that uses `Base.BufferStream` with the `response_stream` keyword:
https://github.com/JuliaCloud/AWS.jl/blob/9a37112f63cc38a59f6b255bb1aff74fc697eec5/src/utilities/request.jl#L186-L196

Example now hangs: https://github.com/JuliaCloud/AWSS3.jl/issues/215 (i'll see about recreating this with AWS.jl directly)

I thought maybe this was something to fix at HTTP.jl (https://github.com/JuliaWeb/HTTP.jl/issues/772), but having seen the HTTP PR above, i wonder if it might really be here in AWS.jl that we should make a change

nickrobinson251:
Fixed in AWS.jl v1.6.2 via #467

Potential follow-up: #468

Issue #433, opened 2021-08-26
[Downloads backend] are we `close`ing streams correctly?


ericphanson:
Ref https://github.com/JuliaCloud/AWS.jl/pull/396#discussion_r694661438

c42f:
> I think maybe one thing to do is to just replace all usage of `BufferStream` with `IOBuffer`. I dunno if there is a specific reason that we are currently using `BufferStream`

I think I may have found the reason for this.  It turns out that `HTTP.StreamLayer` actually closes the stream:

https://github.com/JuliaWeb/HTTP.jl/blob/cfcc595449b9de8869b28f04dcb25ad1f14438e3/src/StreamRequest.jl#L149

But a closed `IOBuffer` may not have its buffer read:

```julia
julia> io = IOBuffer()
       close(io)
       read(io)
ERROR: ArgumentError: read failed, IOBuffer is not readable
...
```

The underlying problem here may be that `Base` had no distinction between `closeread()` and `closewrite()`.

* HTTP.jl calls `close` and assumes it means `closewrite` (I guess). But this makes it unusable with `IOBuffer`
* Downloads.jl doesn't close the stream at all which means the caller needs to close it if it's a `BufferStream`.

See also https://github.com/JuliaLang/julia/issues/24526

c42f:
Another cross ref: It's a known bug that `IOBuffer` doesn't work with HTTP.jl :cry:  https://github.com/JuliaWeb/HTTP.jl/issues/543

vtjnash:
`IOBuffer()` is a file-like object, so it is incompatible with streaming modification. Use `BufferStream` for a stream-like object.

omus:
> `IOBuffer()` is a file-like object, so it is incompatible with streaming modification. Use `BufferStream` for a stream-like object.

@vtjnash can you elaborate more? Unfortunately `Base.BufferStream` is undocumented and unexported and it would be great to know more about when it should be used and why.

vtjnash:
Using IOBuffer leaves you vulnerable to truncation and data loss. `Base.BufferStream` needs to be exported, since it does not have that vulnerability, but is otherwise pretty much a drop-in replacement.

vtjnash:
The exception is if you know that the data is being written synchronously (as HTTP does), as then you can treat the output as a file, instead of expecting a stream.

omus:
[...]

PR #467, opened 2021-09-28
Always close the `response_stream` after `HTTP.request`

nickrobinson251:
- We need to explicitly close the `response_stream`
  (`HTTP.request` no longer closes it for us)
  in HTTP.jl v0.9.15+ (see https://github.com/JuliaWeb/HTTP.jl/pull/752)
- Since `close` is safe to call multiple times,
  this should be compatible with old HTTP.jl versions. 
- Should fix issue #466 (and https://github.com/JuliaCloud/AWSS3.jl/issues/215)
- Also since we're fixing this here, we can close https://github.com/JuliaWeb/HTTP.jl/issues/772

~TODO: add test... i just need to recreate the issue using AWS.jl explicitly (rather than AWSS3)~ done.

### DIFF
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -214,7 +214,7 @@ function _http_request(http_backend::HTTPBackend, request::Request)
                         (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
         end
     finally
-        close(request.response_stream)
+        request.response_stream isa IO && close(request.response_stream)
     end
 end
 
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -213,6 +213,8 @@ function _http_request(http_backend::HTTPBackend, request::Request)
                         isa(e, Base.IOError) ||
                         (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
         end
+    finally
+        close(request.response_stream)
     end
 end
 
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -68,7 +68,9 @@ try
             #   => BUG: header `response_stream` is pushed into the query...
             io = Base.BufferStream()
             S3.get_object(
-                BUCKET_NAME, file_name, Dict("response_stream" => io, "return_stream" => true)
+                BUCKET_NAME,
+                file_name,
+                Dict("response_stream" => io, "return_stream" => true),
             )
             if bytesavailable(io) > 0
                 @test String(readavailable(io)) == body
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -1,6 +1,6 @@
 @service S3
 
-const BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
+BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
 
 try
     S3.create_bucket(BUCKET_NAME)
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -5,95 +5,95 @@ const BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
 try
     S3.create_bucket(BUCKET_NAME)
 
-@testset "issue 223" begin
-    # https://github.com/JuliaCloud/AWS.jl/issues/223
-    body = "Hello World!"
-    file_name = "contains spaces"
-
-    try
-        S3.put_object(BUCKET_NAME, file_name, Dict("body" => body))
-        resp = S3.get_object(BUCKET_NAME, file_name)
-
-        @test String(resp) == body
-    finally
-        S3.delete_object(BUCKET_NAME, file_name)
-    end
-end
-
-@testset "issue 227" begin
-    @testset "s3 public bucket" begin
-        # https://github.com/JuliaCloud/AWS.jl/issues/227
-        config = AWSConfig(; creds=nothing)
-        resp = S3.get_object("www.invenia.ca", "index.html"; aws_config=config)
-
-        @test !isempty(resp)
-    end
-
-    @testset "s3 private bucket" begin
-        bucket_name = "aws-jl-test-issues---" * _now_formatted()
-        file_name = "hello_world"
+    @testset "issue 223" begin
+        # https://github.com/JuliaCloud/AWS.jl/issues/223
+        body = "Hello World!"
+        file_name = "contains spaces"
 
         try
-            S3.create_bucket(bucket_name)
-            S3.put_object(bucket_name, file_name)
+            S3.put_object(BUCKET_NAME, file_name, Dict("body" => body))
+            resp = S3.get_object(BUCKET_NAME, file_name)
 
-            @test_throws AWSException S3.get_object(
-                bucket_name, file_name; aws_config=AWSConfig(; creds=nothing)
-            )
+            @test String(resp) == body
         finally
-            S3.delete_object(bucket_name, file_name)
-            S3.delete_bucket(bucket_name)
+            S3.delete_object(BUCKET_NAME, file_name)
         end
     end
 
-    @testset "lambda" begin
-        @service Lambda
+    @testset "issue 227" begin
+        @testset "s3 public bucket" begin
+            # https://github.com/JuliaCloud/AWS.jl/issues/227
+            config = AWSConfig(; creds=nothing)
+            resp = S3.get_object("www.invenia.ca", "index.html"; aws_config=config)
 
-        @test_throws NoCredentials Lambda.list_functions(;
-            aws_config=AWSConfig(; creds=nothing)
-        )
-    end
-end
-
-@testset "issue 324" begin
-    body = "Hello World!"
-    file_name = "streaming.bin"
-
-    try
-        S3.put_object(BUCKET_NAME, file_name, Dict("body" => body))
-        resp = S3.get_object(BUCKET_NAME, file_name)
-        @test String(resp) == body
-
-        # ERROR: MethodError: no method matching iterate(::Base.BufferStream)
-        #   => BUG: header `response_stream` is pushed into the query...
-        io = Base.BufferStream()
-        S3.get_object(
-            BUCKET_NAME, file_name, Dict("response_stream" => io, "return_stream" => true)
-        )
-        if bytesavailable(io) > 0
-            @test String(readavailable(io)) == body
-        else
-            @test "no body data was available" == body
+            @test !isempty(resp)
         end
 
-    finally
-        S3.delete_object(BUCKET_NAME, file_name)
-    end
-end
+        @testset "s3 private bucket" begin
+            bucket_name = "aws-jl-test-issues---" * _now_formatted()
+            file_name = "hello_world"
 
-@testset "issue 466" begin
-    file_name = "hang.txt"
+            try
+                S3.create_bucket(bucket_name)
+                S3.put_object(bucket_name, file_name)
 
-    try
-        S3.put_object(BUCKET_NAME, file_name)
-        stream = S3.get_object(BUCKET_NAME, file_name, Dict("return_stream" => true))
-        println("test #466")  # So we know if this is the reason for tests hanging.
-        @test eof(stream)  # This will hang if #466 not fixed and using HTTP.jl v0.9.15+
-        println("#466 fixed")
-    finally
-        S3.delete_object(BUCKET_NAME, file_name)
+                @test_throws AWSException S3.get_object(
+                    bucket_name, file_name; aws_config=AWSConfig(; creds=nothing)
+                )
+            finally
+                S3.delete_object(bucket_name, file_name)
+                S3.delete_bucket(bucket_name)
+            end
+        end
+
+        @testset "lambda" begin
+            @service Lambda
+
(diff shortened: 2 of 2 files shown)


src/utilities/request.jl
```diff
@@ -213,6 +213,8 @@ function _http_request(http_backend::HTTPBackend, request::Request)
                         isa(e, Base.IOError) ||
                         (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
         end
+    finally
+        close(request.response_stream)
```
fredrikekre, line 217: Probably should be conditional on if a BufferStream was actually initialized on L193? I don't know this code at all though, but seems like `.response_stream` is at least sometimes `nothing`?

nickrobinson251: good point!

```suggestion
        request.response_stream === nothing || close(request.response_stream)
```

ericphanson:
I'm a bit worried about the `finally`, since I see no other uses in this package where we use Retry.jl's macros with `finally`, and there are no tests for it: https://github.com/JuliaWeb/Retry.jl/blob/master/test/runtests.jl.

nickrobinson251:
> I'm a bit worried about the `finally`, since I see no other uses in this package where we use Retry.jl's macros with `finally`

i suppose we could just call `close` in both branches of the `try`, if we want to avoid using `finally`?

e.g.
```diff
-        return @mock HTTP.request(
+        resp = @mock HTTP.request(
             http_stack,
             request.request_method,
             HTTP.URI(request.url),
             HTTP.mkheaders(request.headers),
             request.content;
             require_ssl_verification=false,
             response_stream=request.response_stream,
             http_options...,
         )
+        request.response_stream === nothing || close(request.response_stream)
+        return resp
     catch e
+        request.response_stream === nothing || close(request.response_stream)
         # Base.IOError is needed because HTTP.jl can often have errors that aren't
```

src/utilities/request.jl
oxinabox: Maybe:
```suggestion
        request.response_stream isa IO && close(request.response_stream)
```

ericphanson:
The semantics of these macros are pretty confusing; both the `try` and `catch` branches can be executed multiple times if it retries, and we don't want to close the stream if we're going to retry. So I kind of think we need `finally`... just need to be sure it actually works 😅.

ericphanson:
Let’s see what CI says already— we can check at least if tests pass here with the new HTTP.jl release.

bors try

ericphanson:
Tests pass and we are using HTTP v0.9.15: https://github.com/JuliaCloud/AWS.jl/runs/3733253932#step:10:48

so that’s promising

nickrobinson251:
What's a sensible way to test something that causes code to hang?

For now i've just put some code (which hangs on `master`) into a test, but i'm not sure if that's the best way to go (since we rely on CI timing out or, if locally, realising something is wrong): https://github.com/JuliaCloud/AWS.jl/pull/467/commits/4dcecc6686537b563b953010fbee2b2e16efc07a

ericphanson:
We at least have timeouts now (https://github.com/JuliaCloud/AWS.jl/blob/master/.github/workflows/CI.yml#L16) due to hangs in the past, so I think the way you did it is OK.

Should we be creating buckets, or is there one we should use already @mattBrzezinski ?

nickrobinson251:
> Should we be creating buckets

ah yeah, i meant to ask -- the `testset`s in `test/issues.jl` all set up and then teardown buckets, which is kinda nice but also kinda slow and make for a lot of boilerplate. For now i've just followed the same pattern, but (perhaps as a follow-up) maybe all these `issues.jl` tests on S3 could just use different files in the same bucket?

src/utilities/request.jl
```diff
@@ -213,6 +213,8 @@ function _http_request(http_backend::HTTPBackend, request::Request)
                         isa(e, Base.IOError) ||
                         (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
         end
+    finally
+        request.response_stream isa IO && close(request.response_stream)
```
mattBrzezinski, line 217: If an end-user were to pass in their own `response_stream`, what would happen here if it closes? Is that data still available in it, or does it go away?

fredrikekre: I was gonna suggest something like
```diff
diff --git a/src/utilities/request.jl b/src/utilities/request.jl
index 12debd4..2ef8363 100644
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -189,11 +189,13 @@ function _http_request(http_backend::HTTPBackend, request::Request)
     @repeat 4 try
         http_stack = HTTP.stack(; redirect=false, retry=false, aws_authorization=false)
 
+        should_close = false
         if request.return_stream && request.response_stream === nothing
             request.response_stream = Base.BufferStream()
+            should_close = true
         end
 
-        return @mock HTTP.request(
+        r = @mock HTTP.request(
             http_stack,
             request.request_method,
             HTTP.URI(request.url),
@@ -203,6 +205,10 @@ function _http_request(http_backend::HTTPBackend, request::Request)
             response_stream=request.response_stream,
             http_options...,
         )
+        if should_close
+            close(request.response_stream)
+        end
+        return r
     catch e
         # Base.IOError is needed because HTTP.jl can often have errors that aren't
         # caught and wrapped in an HTTP.IOError
```
earlier. E.g. only close if it was opened internally.

nickrobinson251: the behaviour is exactly as before (i.e. as we had with older HTTP.jl versions). In older HTTP versions, `close` was already called inside the `HTTP.request` above ﻿https://github.com/JuliaWeb/HTTP.jl/pull/752/files#diff-2877f0a59db9c7dda95204097d3c2010021eb8f6c59c98d56d3f7d16e38b0acbL149

ericphanson: > the behaviour is exactly as before (i.e. as we had with older HTTP.jl versions)

I think the idea is that this maybe wasn't correct from HTTP.jl which is why it got changed there. https://github.com/JuliaCloud/AWS.jl/pull/396#discussion_r694661438 and #433  have lots of discussion on this (which I forgot completely about until just now)

nickrobinson251: > only close if it was opened internally

wouldn't this be different behaviour to what we had when using older HTTP.jl versions? 
i.e. didn't `HTTP.request` just call `close` on the `response_stream` (if it wasn't `nothing`)?

(i don't know, since i'm not too familiar with this stuff)

test/issues.jl
```diff
@@ -81,3 +81,24 @@ end
         S3.delete_bucket(bucket_name)
     end
 end
+
+@testset "issue 466" begin
+    bucket_name = "aws-jl-test-issues---" * _now_formatted()
```
mattBrzezinski, line 86: Do we need to create a new bucket for this test? This file creates one at the beginning.

mattBrzezinski (commented):
LGTM, left a couple comments.

src/utilities/request.jl
fredrikekre: Yea probably. I guess the motivation for the change in HTTP.jl is that people would like to handle the closing themselves, and maybe the same argument can apply here too. I wasn't very involved in those discussions though.

mattBrzezinski: I have not been paying too much attention to the HTTP.jl changes. Are there use cases anyone can think of to keeping the streams open and having users manually close them?

Personally I think closing the HTTP requests automatically makes sense. It gives a loose contract of, this is the data which AWS has provided you from your request.

Maybe I'm missing something obvious though.

fredrikekre: Yea, maybe it should be reverted.

test/issues.jl
nickrobinson251: the testsets above need a little bit of re-jigging if we're to use that same bucket throughout -- i've made that change as i think it make most sense to just use that one bucket when possible

src/utilities/request.jl
nickrobinson251: FWIW, i have no opinion on how best to fix this. But i am keen to fix it as fast as possible, as not being able get files from S3 is quite debilitating for some workflows at Invenia 😊

test/issues.jl
```diff
@@ -1,21 +1,20 @@
 @service S3
 
-bucket_name = "aws-jl-test-issues---" * _now_formatted()
-S3.create_bucket(bucket_name)
+const BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
+S3.create_bucket(BUCKET_NAME)
```
mattBrzezinski, line 4: This might need a bit better of a solution, but we should wrap all these `@testset` in a `try/finally` and delete the bucket afterwards. Otherwise on any failure it'll continue to exist.

mattBrzezinski (approved):
One comment about the `try/finally` in `test/issues.jl` and it LGTM!

mattBrzezinski:
FWIW You should be able to just run JuliaFormatter w/ BlueStyle on the entire package and it'll make the CI step happy. I believe I already cleaned this repo out.

test/issues.jl
nickrobinson251: done https://github.com/JuliaCloud/AWS.jl/pull/467/commits/27dfaee697cd1d1f223cbe9f3c98657fdfc9d8c9

ericphanson:
I'm concerned about this change; I think https://github.com/JuliaCloud/AWS.jl/pull/467#discussion_r717666189 is probably the right way. 

In https://github.com/JuliaCloud/AWS.jl/pull/396#discussion_r695321404 @vtjnash said
> Whoever created the stream is generally responsible for marking when it is done, or passing it to another function which will take care of that.

which sounds like we should only be closing the ones we create here.

mattBrzezinski:
> I'm concerned about this change; I think [#467 (comment)](https://github.com/JuliaCloud/AWS.jl/pull/467#discussion_r717666189) is probably the right way.
> 
> In [#396 (comment)](https://github.com/JuliaCloud/AWS.jl/pull/396#discussion_r695321404) @vtjnash said
> 
> > Whoever created the stream is generally responsible for marking when it is done, or passing it to another function which will take care of that.
> 
> which sounds like we should only be closing the ones we create here.

I haven't had time to read through all these threads, but will soon(TM). I think we can push these changes through as they mimic the existing functionality, then open an issue to discuss?

ericphanson:
Ok, that sounds fine to me

src/utilities/request.jl
oxinabox: Yes, I would say we should merge this PR which basically brings back the behavior of HTTP 0.9.14 and then continue this discussion in an issue.

I would really like to see this merged today.

mattBrzezinski:
CI failed, but also threw

> WARNING: redefinition of constant BUCKET_NAME. This may fail, cause incorrect answers, or produce other errors.

test/issues.jl
```diff
@@ -1,83 +1,102 @@
...
-
-    try
-        S3.put_object(bucket_name, file_name, Dict("body" => body))
-        resp = S3.get_object(bucket_name, file_name)
-
-        @test String(resp) == body
-    finally
-        S3.delete_object(bucket_name, file_name)
-        S3.delete_bucket(bucket_name)
+const BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
```
nickrobinson251, line 3: this file get's `include`d multiple times

```suggestion
BUCKET_NAME = "aws-jl-test-issues---" * _now_formatted()
```

src/utilities/request.jl
nickrobinson251: https://github.com/JuliaCloud/AWS.jl/issues/468

nickrobinson251:
i don't understand bors -- is CI running, or...?

omus:
Personally I think the way this should have been address was to upper bound HTTP.jl. That would have been a trivial change in comparison and would have given us more time to work out the details with this PR

oxinabox:
A fair point conceptually.
But in practice an incredibly annoying one, do to how julia's resolver/registry works.
since would have needed to go and retroactively capped the version of HTTP.jl in all past releases of  AWS.jl.
Which would have required manual registry edits.
(I have suggested before that for post 1.0 packages, like AWS.jl the registry should refuse to install any but the lasted patch release which would have fixed it.)

Another good option might have been to revert the change to HTTP.jl  and tag a release.
(since it was supposedly a nonbreaking change but did infact break a pretty core package for a lot of people) 
While we worked out weather or not it was breaking after all, or wether AWS was violating AWS.jl's public API (and if so fixed AWS.jl)

nickrobinson251:
Yeah, reverting the change to HTTP.jl would have been another option, but may have required more discussion.
Pinning HTTP is quite painful, as Lyndon says. This PR was a 2 line change to the `src` (the big diffs to the tests were various requests made in code review, mostly addressing issues in the existing tests, arguably out of scope for this PR tbh). And since this just maintains the existing (pre-HTTP.jl v0.9.15) behaviour in an internal function, i don't really see the harm. As far as this package is concerned, it just protects against the relevant change in HTTP. Especially as a change like #468 can be done, and maybe should be done, (just like it could have been before).

PR #470, opened 2021-09-29
Test `isopen` on `response_stream`

omus:
Follow up to #467 and #466. The original test for issue #466 used `eof` which could cause the tests to hang if the code was configured incorrectly. The tests have been revised to use `isopen` which doesn't have this issue.

I further improved upon the tests by checking the behaviour of using `"response_stream"` and `"return_stream"`. As it turns out it has revealed a bug with the `DownloadsBackend` which I won't be fixing at this time (#471)

Finally, I also duplicated this test suite for MinIO as this allows this behaviour to be tested locally and not just on the CI. I can remove this if desired.

### DIFF
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -103,7 +103,7 @@ try
             if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
                 @test !isopen(stream)
             else
-                @test !isopen(stream)
+                @test_broken isopen(stream)
             end
 
             stream = Base.BufferStream()
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -107,7 +107,11 @@ try
             end
 
             stream = Base.BufferStream()
-            S3.get_object(BUCKET_NAME, file_name, Dict("response_stream" => stream, "return_stream" => true))
+            S3.get_object(
+                BUCKET_NAME,
+                file_name,
+                Dict("response_stream" => stream, "return_stream" => true),
+            )
             if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
                 @test !isopen(stream)
             else
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -91,7 +91,6 @@ try
 
             # The tests below validate the current behavior of how streams are handled.
             # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
-            
 
             stream = S3.get_object(BUCKET_NAME, file_name, Dict("return_stream" => true))
             if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -89,7 +89,9 @@ try
         try
             S3.put_object(BUCKET_NAME, file_name)
 
+            # The tests below validate the current behavior of how streams are handled.
             # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
+            
 
             stream = S3.get_object(BUCKET_NAME, file_name, Dict("return_stream" => true))
             if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
@@ -103,6 +105,7 @@ try
             if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
                 @test !isopen(stream)
             else
+                # See: https://github.com/JuliaCloud/AWS.jl/issues/471
                 @test_broken isopen(stream)
             end
 
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -88,10 +88,31 @@ try
 
         try
             S3.put_object(BUCKET_NAME, file_name)
+
+            # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
+
             stream = S3.get_object(BUCKET_NAME, file_name, Dict("return_stream" => true))
-            println("test #466")  # So we know if this is the reason for tests hanging.
-            @test eof(stream)  # This will hang if #466 not fixed and using HTTP.jl v0.9.15+
-            println("#466 fixed")
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
+                @test !isopen(stream)
+            else
+                @test isopen(stream)
+            end
+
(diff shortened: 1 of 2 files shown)


omus:
Note I've tested these changes locally using HTTP 0.9.14 and 0.9.15

omus:
cc: @nickrobinson251

omus:
bors try

test/issues.jl
```diff
@@ -88,10 +88,35 @@ try
...
+            else
+                @test isopen(stream)
+            end
+
+            stream = Base.BufferStream()
+            S3.get_object(BUCKET_NAME, file_name, Dict("response_stream" => stream))
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
+                @test !isopen(stream)
+            else
+                @test_broken isopen(stream)
```
nickrobinson251, line 106: can there be a comment (or link to an issue) about this?

test/issues.jl
```diff
@@ -88,10 +88,35 @@ try
...
-            println("#466 fixed")
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
+                @test !isopen(stream)
+            else
+                @test isopen(stream)
+            end
+
+            stream = Base.BufferStream()
+            S3.get_object(BUCKET_NAME, file_name, Dict("response_stream" => stream))
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
```
nickrobinson251, line 103: Why do different backends have different behaviour?

test/issues.jl
```diff
@@ -88,10 +88,35 @@ try
...
+            stream = Base.BufferStream()
+            S3.get_object(
+                BUCKET_NAME,
+                file_name,
+                Dict("response_stream" => stream, "return_stream" => true),
+            )
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
+                @test !isopen(stream)
+            else
+                @test isopen(stream)
```
nickrobinson251, line 118: it's expected to be `test_broken` above, but to work here -- why is that?

nickrobinson251 (approved):
Changing to `isopen` seems a good idea.

I don't understand the behaviour of different backends -- i know you said there's a bug in Downloads.jl-backend, but from the code itself i don't follow what the expected behaviour is and which bits are broken/buggy. It looks like we _expect_ different behaviour from the different backends, but that is surprising to me, so i think warrants a comment with a link to an issue explaining

test/issues.jl
omus: They really shouldn't. I think this all comes the fact that `HTTPBackend` returns a `Base.BufferStream` by default while `DownloadsBackend` returns an `IOBuffer` by default. For this PR I'm just recording the current behaviour and avoiding changing it because as we know closing/not closing a file can be a breaking behaviour

test/issues.jl
omus: I'll file an issue for this

test/issues.jl
omus: This is the reason for the difference in behaviour. Calling `read_body` may or may not close the `IO`:

https://github.com/JuliaCloud/AWS.jl/blob/1722aa5a34b9e1c5084dd8677ff4898514956b4d/src/utilities/downloads_backend.jl#L35-L39

Whether this function is called is dependent on a couple of factors: 

https://github.com/JuliaCloud/AWS.jl/blob/1722aa5a34b9e1c5084dd8677ff4898514956b4d/src/utilities/downloads_backend.jl#L70-L74

test/issues.jl
omus: I'm going to address these inconsistencies with #384

nickrobinson251: thanks, i think the changes here would benefit from a comment explaining that they are testing the current behaviour and also linking to the issue you intend to open about possibly changing this behaviour

The changes themselves all LGTM, i was just struggling to follow what it is that these tests are documenting

test/issues.jl
```diff
@@ -88,10 +88,35 @@ try
...
         try
             S3.put_object(BUCKET_NAME, file_name)
+
+            # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
+
             stream = S3.get_object(BUCKET_NAME, file_name, Dict("return_stream" => true))
-            println("test #466")  # So we know if this is the reason for tests hanging.
-            @test eof(stream)  # This will hang if #466 not fixed and using HTTP.jl v0.9.15+
-            println("#466 fixed")
+            if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
```
mattBrzezinski, line 95: Would we not want to have something like `@testset "issue 466 - $(typeof(backend))" for backend in [AWS.HTTPBackend, AWS.DownloadBackend]`

And just have test cases for both here?

mattBrzezinski (approved):

test/issues.jl
omus: #471

test/issues.jl
omus: I think ideally the behaviour between the backends would be the same so we wouldn't have to make that distinction. For testing the current behaviour we can either do what is done in this PR or do:
```julia
if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
    @testset "issue 446 - HTTPBackend" begin
    end
else
    @testset "issue 446 - DownloadsBackend" begin
    end
end
```

test/issues.jl
```diff
@@ -88,10 +88,35 @@ try
 
         try
             S3.put_object(BUCKET_NAME, file_name)
+
+            # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
```
omus, line 92:
```suggestion
            # The tests below validate the current behavior of how streams are handled.
            # Note: Avoid using `eof` for these tests can hang when using an unclosed `Base.BufferStream`
            
```

test/issues.jl
omus, line 106:
```suggestion
                # See: https://github.com/JuliaCloud/AWS.jl/issues/471
                @test_broken isopen(stream)
```

omus:
I'll squash and merge this PR as this doesn't need to be more than a single commit

PR #484, opened 2021-10-08
Allow "[profile default]" in .aws/config

c42f:
The awscli allows either a section `[default]` or `[profile default]` in
the aws config file.

I had `[profile default]` in my config and was surprised when this worked with awscli but not with AWS.jl.


omus:
Seems reasonable. Needs a test though

c42f:
Test added.

test/AWSConfig.jl
```diff
@@ -19,4 +19,18 @@
...
+            for default_section_str in allowed_default_sections
+                config = """
+                [$default_section_str]
+                region = xx-yy-1
+                """
+                write(config_path, config)
+                @test aws_get_region(; profile="default", config=config_path) == "xx-yy-1"
+            end
+        end
+    end
```
mattBrzezinski, line 35: I believe this is the proper syntax.

```suggestion
    @testset "$(profile) section names" for profile in ["default", "profile_default"]
        mktemp() do config_path, _
                config = """
                [$(profile)]
                region = xx-yy-1
                """
                write(config_path, config)
                @test aws_get_region(; profile="default", config=config_path) == "xx-yy-1"
            end
        end
    end
```

mattBrzezinski: It's nbd, I just find different cases in a `@testset` are easier to debug.

Issue #496, opened 2021-10-20
Add retry for `RequestTimeTooSkewed`


ericphanson:
```julia
ERROR: LoadError: AWSException: RequestTimeTooSkewed -- The difference between the request time and the current time is too large.
```
We saw this error making a 1.3 MB upload that succeeded many times without issue in the same job / from the same k8s pod, but failed once. @christopher-dG and I think it was likely a transient issue. If this kind of error really can be due to a transient issue (and not an actual clock problem or other issue) then it's probably worth adding a retry for it.

FWIW it looks like s5cmd retries for it: https://github.com/peak/s5cmd/pull/221/files

PR #497, opened 2021-10-20
Retry RequestTimeTooSkewed

mattBrzezinski:
Resolves: #496 

This will retry if a `RequestTimeTooSkewed` error is thrown. Seems like this error is caused by issues with the NTP, [article](https://medium.com/@vuongtran/how-to-resolve-error-amazon-s3-requesttimetooskewed-5d723dc75735).


ericphanson:
> I think I might be sending off way too many requests with asyncmap such that some of the tasks start their connection but then don't get touched for a long time (since the CPU is busy handling other tasks and doing other things), and that could yield this kind of error... Hm, in which case I think retrying might help alleviate the symptoms here but there really is an underlying issue in my code.

Just hit this again without doing any kind of concurrent downloading. So I'm definitely back onboard with adding a retry aha. Could still be some kind of weird cluster/network issue on our end, but I've got no idea what it would be.

ericphanson:
@christopher-dG had the theory that since I ran into this on the first call for a particular function in a Julia session, it could be compilation latency causing a delay in processing the HTTP request, leading to this issue. (I found this theory pretty fun aha 😄).

Issue #515, opened 2021-11-14
S3.get_object sometimes returning incomplete ranged data without error


IanButterworth:
On modest/slow internet connections we're seeing `S3.get_object` sometimes return incomplete range chunks without error or warning
The MWE below is on a public 14.4 GB object so it should be testable.

Our work-around for this is to check the length returned and retry if incorrect, which seems to work.

But this seems like something that could/should be caught by the HTTP backend error handling?
Further, the length check is only checking length, not content. Looking into whether ranged S3 GETs can be checksum-ed, it appears that the `ETag` of these kind of requests is only in special cases the MD5 of the chunk, namely when the file was uploaded in the same chunks, which is a little limiting. But my understanding may be wrong here.

Is there perhaps a HTTP exception not being caught?

julia 1.6.3
```julia
julia> import Pkg; Pkg.activate(temp=true); Pkg.add("AWS")
  Activating new environment at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_4sjiZA/Project.toml`
    Updating registry at `~/.julia/registries/General`
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_4sjiZA/Project.toml`
  [fbe9abb3] + AWS v1.71.0
...

julia> using AWS: AWS, @service

julia> @service S3
Main.S3

julia> aws_config = AWS.AWSConfig(region = "us-east-1")
AWS.AWSConfig((xxx, yyy, 146138512-12-31T23:59:59), "us-east-1", "json")

julia> data = S3.get_object("ryft-public-sample-data", "demo_searchlynx_example.tgz", Dict("headers" => Dict("Range"=>"bytes=0-16777215")); aws_config);

julia> typeof(data)
Vector{UInt8} (alias for Array{UInt8, 1})

julia> Base.format_bytes(length(data))
"16.000 MiB"

julia> @sync for i in 1:10
           @async begin
	          bytes_start = 0
	          for j in 1:10
	              bytes_end = bytes_start + 16777216 - 1
	              @info "getting $bytes_start-$bytes_end"
	              newdata = S3.get_object("ryft-public-sample-data", "demo_searchlynx_example.tgz", Dict("headers" => Dict("Range"=>"bytes=$(bytes_start)-$(bytes_end)")); aws_config);
	              if length(newdata) != ((bytes_end - bytes_start) + 1)
[...]

PR #516, opened 2021-11-17
Fix incomplete get handling via HTTP compat bump and don't write failed retry streams to parent buffer in _http_request

IanButterworth:
I believe https://github.com/JuliaWeb/HTTP.jl/pull/781 is the proper fix needed for #515.
That PR introduces an EOFError throw if the http response stream has promised content via `Content-length` but hits `eof`.

The issue is rare and only in heavily concurrent operations, and the retry handling in AWS.jl is well set up to handle that

However, there was also an AWS.jl bug that this PR fixes, where each failed retry was being passed to the `response_stream`.


ericphanson:
> heavily concurrent operations

This is totally orthogonal but if you run into https://github.com/JuliaWeb/HTTP.jl/issues/517 style issues (which I certainly did), you might want to try the Downloads.jl based backend to use libcurl instead of HTTP.jl. `AWS.DEFAULT_BACKEND[] = AWS.DownloadsBackend()` (or on a per-request basis with `backend=…`).

IanButterworth:
HTTP v0.9.17 is out https://github.com/JuliaRegistries/General/pull/48956

IanButterworth:
Can someone run the CI please 🙏🏻

mattBrzezinski:
bors try

IanButterworth:
I assume the bors failure is something else?

mattBrzezinski:
bors try

IanButterworth:
@omus Matt recommended I check if you know what's up with bors here?

omus:
> @omus Matt recommended I check if you know what's up with bors here?

Ah, your move of `write(response_stream, buffer)` is the source of the problem. We need to write out the streamed data to `response_stream` during a failure otherwise the error is lost. That's exactly what this error is trying to communicate:

```
┌ Error: Internal Error: provided body is empty while the reported content-length is non-zero
└ @ AWS.AWSExceptions ~/work/AWS.jl/AWS.jl/src/AWSExceptions.jl:92
```

omus:
> However, there was also an AWS.jl bug that this PR fixes, where each failed retry was being passed to the `response_stream`.

To fix this you may need to make `buffer` outside of the retry block and have an outer `try`/`finally` which writes the final response to the `response_stream` (a test should be added for this).

A possible better solution could be to not have `HTTP.request` throw exceptions. You'd need to throw an exception on the last attempt though.

omus:
bors try

IanButterworth:
Great. Thanks @omus

omus (changes requested):
I'd like to see a test for the `EOFError` behaviour. Otherwise these changes look good

IanButterworth:
Ok. I'm not familiar with the mocking system here. Could you point me to a similar test?

ericphanson:
bors try

ericphanson:
bors try

IanButterworth:
Sorry for the noise.. I'm assuming I can't run tests locally first

ericphanson:
I wonder if this will work to let you run CI yourself:

bors delegate+

ericphanson:
bors try

IanButterworth:
I manually commented out test code and got this to work in isolation locally

ericphanson:
bors try

test/issues.jl
```diff
@@ -155,6 +155,32 @@ try
...
+                if ATTEMPT_NUM == 1
+                    write(response_stream, rand(UInt8, 34)) # an incomplete stream that shouldn't be retained
+                    throw(HTTP.IOError(EOFError(), "msg"))
+                else
+                    write(response_stream, data)
+                end
+                headers = [
+                    "Content-Length" => string(length(data)),
+                ]
+                return HTTP.Response(200, "{\"Location\": \"us-east-1\"}")  # Made up region
```
IanButterworth, line 174: Dead code/old comment
```suggestion
                return HTTP.Response(200, "{\"Location\": \"us-east-1\"}")
```

IanButterworth:
I think this PR still has an inconsistency between the HTTP and Downloads backends.
I looked into the Downloads behavior and it does throw, but because the Downloads backend doesn't use a sacrificial buffer, AWS doesn't prevent the incomplete stream from being passed to the output buffer

I expanded the tests to cover the Downloads backend, which now passes locally for the HTTP backend, and fails for the Downloads backend with:

```julia
issue 515: Test Failed at /Users/ian/Documents/GitHub/AWS.jl/test/issues.jl:188
  Expression: length(resp) == length(data)
   Evaluated: 199 == 100
```

Here's a lower level example of the behavior that passes
```julia
using HTTP, Sockets, Test, Downloads

function do_incomplete_server()
    server = Sockets.listen(ip"0.0.0.0", 8080)
    task = @async HTTP.listen("0.0.0.0", 8080; server=server) do http
        HTTP.setstatus(http, 200)
        HTTP.setheader(http, "Content-Length" => "64") # Promise 64 bytes...
        HTTP.startwrite(http)
        HTTP.write(http, rand(UInt8, 63)) # ...but only send 63 bytes.
        # Close the stream so that eof(stream) is true and the client isn't
        # waiting forever for the last byte.
        HTTP.close(http.stream)
    end
    return server, task
end

@testset "HTTP backend" begin
    server, task = do_incomplete_server()
    err = nothing
    io = Base.BufferStream()
    try
        HTTP.get("http://localhost:8080"; response_stream = io, retry=false)
    catch err
        @show err
    finally
        try; close(server); wait(task); catch; end # Shutdown
    end
    @test bytesavailable(io) == 63
    @test err isa HTTP.IOError
    @test err.e isa EOFError
end


@testset "Downloads backend" begin
    server, task = do_incomplete_server()
    err = nothing
    io = Base.BufferStream()
    try
        Downloads.request("http://localhost:8080"; output = io, throw = true)
    catch err
        @show err
    finally
        try; close(server); wait(task); catch; end # Shutdown
    end
    @test bytesavailable(io) == 63
    @test err isa RequestError
end
```

What's the right thing to do here? It seems the use of a sacrificial buffer is a patch.. so i'm not sure whether you'd want to add that to the Downloads backend to fix this?

IanButterworth:
I just pushed a proposed fix. Tests should pass now

ericphanson:
bors try

ericphanson:
bors try

src/utilities/downloads_backend.jl
```diff
@@ -65,6 +65,19 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
+        try
+            pos = position(output)
+            seek(output, pos)
+            pos, true
+        catch
+            bytesavailable(output), false
+        end
+    else
+        nothing, false
+    end
```
omus, line 79: I think we have to use a sacrificial buffer if you require `seek` support as not all `IO` types support it.

IanButterworth: Ok. I reverted and added a sacrificial buffer. Although the previous tests passed, the approach was flawed as `read` was reading from the start, not the end of the `response_buffer` which happened to be fresh in the test case, but may not be in the real world

ericphanson:
bors try

IanButterworth:
Ok. I'm now happy with this. Sorry it took so many attempts.

As this PR found and fixed, both the HTTP and Downloads backends were broken for `S3.get_object` for fixed-length responses when the response was incomplete, but in different ways;
- HTTP gave the incomplete data without error
- Downloads included the output from failed attempts, so too much data

Both are quite insidious for multi-part large file handling.

It would be great to get this out. (We're currently pinned at 1.62)

src/utilities/downloads_backend.jl
```diff
@@ -79,16 +79,36 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
+                output=buffer,
+                method=request.request_method,
+                headers=request.headers,
+                verbose=false,
+                throw=true,
+                downloader=downloader,
+            )
+        catch e
+            if e isa Downloads.RequestError
+                # EOFErrors indicate a broken connection, so don't pass the buffer forward
```
omus, line 99: Comment isn't quite right as there's no interaction with an `EOFError` here. We should probably be more specific in the conditional here as `Downloads.RequestError` will be thrown for any non 2XX status

src/utilities/downloads_backend.jl
```diff
@@ -79,16 +79,36 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
+                should_write = false
+            end
+            rethrow()
+        finally
+            if !isnothing(output)
+                close(buffer)
+                # Transfer the contents of the `BufferStream` into `response_stream` variable.
+                # but only if no EOFError error because of a broken connection.
```
omus, line 107:
```suggestion
                # Transfer the contents of the `BufferStream` into `response_stream` variable
                # but only if no EOFError error because of a broken connection.
```

src/utilities/request.jl
```diff
@@ -168,14 +168,22 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
+                should_write = false
+            end
+            rethrow()
         finally
             # We're unable to read from the `Base.BufferStream` until it has been closed.
             # HTTP.jl will close passed in `response_stream` keyword. This ensures that it
             # is always closed (e.g. HTTP.jl 0.9.15)
             close(buffer)
 
             # Transfer the contents of the `BufferStream` into `response_stream` variable.
```
omus, line 183:
```suggestion
            # Transfer the contents of the `BufferStream` into `response_stream` variable
```

test/issues.jl
```diff
@@ -155,6 +155,41 @@ try
...
+                end
+            end
+        end
+        config = AWSConfig(; creds=nothing)
+        apply(patch) do
+            resp = S3.get_object("www.invenia.ca", "index.html"; aws_config=config) # use public bucket as dummy
+            @test length(resp) == length(data)
+            @test resp == data
+        end
+    end
```
omus, line 191: I'm not sure we have an existing test for this but it would be good to double check the behaviour when the final response is an error. In that case we would want the output to be included in the stream

omus: My concern here is that exceptions from AWS will not be captured appropriately resulting an AWS failure exceptions with no context provided to the user.

IanButterworth: The tests I added are at the `S3.get_object` endpoint, so I've just added a test that it throws when all attempts fail. There's no buffer to check at that level

IanButterworth: I can't figure out a tidy way to do something specifically only on the final `@repeat` retry. Unfortunately `finally` blocks are called each retry.

IanButterworth: I added manual retry tracking as `@repeat` didn't seem to expose a tidier way to do so

IanButterworth:
Could someone please bors try

If this PR requires further changes, I'd greatly appreciate code suggestions. I'm keen for this problem to be fixed and released.

ericphanson:
bors try

omus:
I'll be 

> If this PR requires further changes, I'd greatly appreciate code suggestions. I'm keen for this problem to be fixed and released.

I'll try to help push this along today

test/issues.jl
```diff
@@ -155,6 +155,79 @@ try
...
+                        write(response_stream, data)
+                        return HTTP.Response(200, "{\"Location\": \"us-east-1\"}")
+                    end
+                end
+            elseif AWS.DEFAULT_BACKEND[] isa AWS.DownloadsBackend
+                @patch function Downloads.request(args...; output, kwargs...)
+                    ATTEMPT_NUM += 1
+                    if ATTEMPT_NUM <= i
+                        write(output, data[1:n_incomplete]) # an incomplete stream that shouldn't be retained
+                        throw(Downloads.RequestError("", 18, "transfer closed with 1 bytes remaining to read", Downloads.Response("http", "", 200, "HTTP/1.1 200 OK", ["content-length" => "64"])))
```
omus, line 183: I feel like I'm missing something.
```suggestion
                        throw(Downloads.RequestError("", 18, "transfer closed with 1 bytes remaining to read", Downloads.Response("http", "", 200, "HTTP/1.1 200 OK", ["content-length" => string(n)])))
```

test/issues.jl
```diff
@@ -155,6 +155,79 @@ try
...
+                    else
+                        write(output, data)
+                        return Downloads.Response("http", "", 200, "HTTP/1.1 200 OK", ["content-length" => "64"])
+                    end
+                end
+            end
```
omus, line 222: Avoid duplicating these patches. Something like this would allow you to generate the patches you need:
```julia
function _incomplete_patch(; data, max_attempts=4)
    attempt_num = 0
    n = length(data)

    patch = if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
        @patch function HTTP.request(args...; response_stream, kwargs...)
            attempt_num += 1
            if attempt_num <= max_attempts
                write(response_stream, data[1:n - 1]) # an incomplete stream that shouldn't be retained
                throw(HTTP.IOError(EOFError(), "msg"))
            else
                write(response_stream, data)
                return HTTP.Response(200, "{\"Location\": \"us-east-1\"}")
            end
        end
    elseif AWS.DEFAULT_BACKEND[] isa AWS.DownloadsBackend
        @patch function Downloads.request(args...; output, kwargs...)
            attempt_num += 1
            if attempt_num <= max_attempts
                write(output, data[1:n - 1]) # an incomplete stream that shouldn't be retained
                throw(Downloads.RequestError("", 18, "transfer closed with 1 bytes remaining to read", Downloads.Response("http", "", 200, "HTTP/1.1 200 OK", ["content-length" => string(n)])))
            else
                write(output, data)
                return Downloads.Response("http", "", 200, "HTTP/1.1 200 OK", ["content-length" => string(n))
            end
        end
    end

    return patch
end
```
Note: I added in the `string(n)` for content length and make a couple other tweaks

src/utilities/request.jl
```diff
@@ -144,7 +144,10 @@ end
 function _http_request(http_backend::HTTPBackend, request::Request, response_stream::IO)
     http_options = merge(http_backend.http_options, request.http_options)
 
-    @repeat 4 try
+    attempts = 4
```
omus, line 147: More descriptive variable name:
```suggestion
    max_attempts = 4
```

src/utilities/request.jl
```diff
@@ -168,14 +171,24 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
 
-            # Transfer the contents of the `BufferStream` into `response_stream` variable.
-            write(response_stream, buffer)
+            # Transfer the contents of the `BufferStream` into `response_stream` variable
+            # but only if no EOFError error because of a broken connection OR it's the final attempt.
+            # i.e. Multiple EOFError retries shouldn't be passed to the `response_stream`
+            if should_write || attempt == attempts
+                write(response_stream, buffer)
+            end
         end
```
omus, line 192: Although it's a bit more work I think we should probably switch the `@repeat try` and this one. That should result in less logic to maintain (I'm fine with doing this in a follow up PR if you prefer). Here's a quick demo of how that would work:
```julia
julia> function _request(io::IO, num_fails)
           i = 0
           local buffer
           try
               @repeat 4 try
                   i += 1
                   println("try: $i")
                   buffer = Base.BufferStream()
                   write(buffer, "data")
                   i < num_fails && error("retry")
               catch e
                  println("catch: $i")
                  @retry if e.msg == "retry" end
               end
           finally
               println("finally: $i")
               close(buffer)
               write(io, buffer)
           end
       end
_request (generic function with 1 method)

julia> function demo(num_fails)
           io = IOBuffer()
           try
               _request(io, num_fails)
           finally
               @show String(take!(io))
           end
       end
demo (generic function with 3 methods)

julia> demo(2)
try: 1
catch: 1
try: 2
finally: 2
String(take!(io)) = "data"
false

julia> demo(5)
try: 1
catch: 1
try: 2
catch: 2
try: 3
catch: 3
try: 4
catch: 4
finally: 4
String(take!(io)) = "data"
ERROR: retry
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] macro expansion
   @ ./REPL[33]:10 [inlined]
 [3] macro expansion
   @ ~/.julia/packages/Retry/vS1bg/src/repeat_try.jl:192 [inlined]
 [4] _request(io::IOBuffer, num_fails::Int64)
   @ Main ./REPL[33]:5
 [5] demo(num_fails::Int64)
   @ Main ./REPL[34]:4
 [6] top-level scope
   @ REPL[37]:1
```

test/issues.jl
omus: You can then use this via:
```julia
apply(_incomplete_patch(; data=data, max_attempts=2) do
    ...
end

apply(_incomplete_patch(; data=data, max_attempts=4) do
    ...
end

**Comment by IanButterworth** [comment]
Thanks @omus I just pushed the easy suggestions. Feel free to push the bigger change here, or in a following PR, whichever you prefer

**Comment by ericphanson** [comment]
bors try

**Comment by ericphanson** [comment]
bors try

**Comment by IanButterworth** [comment]
AWS was having a bad day, I think the error above is unrelated and cleared now
```
Load Credentials: Test Failed at /home/runner/work/AWS.jl/AWS.jl/test/AWSCredentials.jl:8
  Expression: e.code in [["AccessDenied", "EntityAlreadyExists"];]
   Evaluated: "400" in ["AccessDenied", "EntityAlreadyExists"]
Stacktrace:
 [1] top-level scope at /home/runner/work/AWS.jl/AWS.jl/test/AWSCredentials.jl:8
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
 [3] top-level scope at /home/runner/work/AWS.jl/AWS.jl/test/AWSCredentials.jl:17
```

File: src/utilities/downloads_backend.jl [hunk 11]
```diff
@@ -65,7 +65,10 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
     # We pass an `input` only when we have content we wish to send.
     input = !isempty(request.content) ? IOBuffer(request.content) : nothing
 
-    @repeat 4 try
+    attempts = 4
```
omus, line 68:
```suggestion
    max_attempts = 4
```

src/utilities/downloads_backend.jl
```diff
@@ -65,7 +65,10 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
+    attempt = 0
+    @repeat attempts try
```
omus, line 70:
```suggestion
    @repeat max_attempts try
```

src/utilities/downloads_backend.jl
```diff
@@ -79,16 +82,38 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
+                should_write = false
+            end
+            rethrow()
+        finally
+            if !isnothing(output)
+                close(buffer)
+                # Transfer the contents of the `BufferStream` into `response_stream` variable.
+                # but only if no EOFError error because of a broken connection OR it's the final attempt.
+                # i.e. Multiple EOFError retries shouldn't be passed to the `response_stream`
+                if should_write || attempt == attempts
```
omus, line 112:
```suggestion
                if should_write || attempt == max_attempts
```

src/utilities/request.jl
```diff
@@ -144,7 +144,10 @@ end
 function _http_request(http_backend::HTTPBackend, request::Request, response_stream::IO)
     http_options = merge(http_backend.http_options, request.http_options)
 
-    @repeat 4 try
+    attempts = 4
+    attempt = 0
+    @repeat attempts try
```
omus, line 149:
```suggestion
    @repeat max_attempts try
```

src/utilities/request.jl
```diff
@@ -168,14 +171,24 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
             # HTTP.jl will close passed in `response_stream` keyword. This ensures that it
             # is always closed (e.g. HTTP.jl 0.9.15)
             close(buffer)
 
-            # Transfer the contents of the `BufferStream` into `response_stream` variable.
-            write(response_stream, buffer)
+            # Transfer the contents of the `BufferStream` into `response_stream` variable
+            # but only if no EOFError error because of a broken connection OR it's the final attempt.
+            # i.e. Multiple EOFError retries shouldn't be passed to the `response_stream`
+            if should_write || attempt == attempts
```
omus, line 189:
```suggestion
            if should_write || attempt == max_attempts
```

omus (approved):
I'm good with to proceed with these changes now. We can attempt https://github.com/JuliaCloud/AWS.jl/pull/516/files#r756212793 as a follow up

omus:
bors try

src/utilities/request.jl
omus: Working in this in https://github.com/JuliaCloud/AWS.jl/pull/522

Issue #434, opened 2021-08-27
ec2_instance_credentials requires role arn


yalwan-sage:
I was looking here: https://github.com/JuliaCloud/AWS.jl/blob/master/src/AWSCredentials.jl#L259
because I was troubleshooting an issue why the credentials were not automatically obtained.
Our systems block this call: `ec2_instance_metadata("/latest/meta-data/iam/info")`

Further to that, even if it didn't throw (because the block 404s), but instead returned nothing, I wouldn't be able to get credentials because at line 256 we bail early if there isn't any info. As best I can tell it is used to fill in the ARN field, but I can see other constructors where the ARN field is permitted to be empty.

Is it possible to relax this?

yalwan-sage:
I wanted to start on this but also wanted to get a feedback:

I'm thinking of modifying the `ec2_instance_metadata` function like so:

```julia
function ec2_instance_metadata(path::AbstractString, ignored_codes::Tuple{Vararg{<:Integer}}=())
    uri = HTTP.URI(; scheme="http", host="169.254.169.254", path=path)
    request = try
        Mocking.@mock HTTP.request("GET", uri; connect_timeout=1)
    catch e
        if e isa HTTP.ConnectionPool.ConnectTimeout
            nothing
        elseif e isa HTTP.StatusError
            if e.status in ignored_codes
                nothing
            else
                rethrow()
            end
        end
    end

    return request !== nothing ? String(request.body) : nothing
end
```

and then for the info retrieval request we can ask to ignore 404 specifically, and then do the remainder of what was needed (i.e. still return credentials without RoleArn if we don't get this)

Is this acceptable?

PR #517, opened 2021-11-17
adding the basic version of RoleARN skipping under KIAM filtering

yalwan-sage:
This is my attempt to deal with #434 (worth noting it works on the setup I've tested) 

Appreciate there's no tests and this might not be the solution you'd like. Appreciate any guidance on what you'd need to see to move this forward.


yalwan-sage:
@mattBrzezinski any remarks?

src/AWSCredentials.jl
```diff
@@ -208,18 +208,27 @@ metadata is available (typically due to not running within an EC2 instance) then
...
 - `path`: The URI path to used to specify that metadata to return
+- `ignored_codes`: a tuple of HTTP status codes to ignore (default empty tuple)
 """
-function ec2_instance_metadata(path::AbstractString)
+
+function ec2_instance_metadata(path::AbstractString, ignored_codes::Tuple{Vararg{<:Integer}}=())
     uri = HTTP.URI(; scheme="http", host="169.254.169.254", path=path)
     request = try
-        @mock HTTP.request("GET", uri; connect_timeout=1)
+        Mocking.@mock HTTP.request("GET", uri; connect_timeout=1)
```
mattBrzezinski, line 222: The `@mock` macro is exported by `Mocking.jl`, no need to preface here.

```suggestion
        @mock HTTP.request("GET", uri; connect_timeout=1)
```

src/AWSCredentials.jl
```diff
@@ -208,18 +208,27 @@ metadata is available (typically due to not running within an EC2 instance) then
...
+If using KIAM or something similar which filters the calls to instance metadata,
+set the `ignored_codes` parameter to prevent exceptions being raised for certain statuses.
+
 # Arguments
 - `path`: The URI path to used to specify that metadata to return
+- `ignored_codes`: a tuple of HTTP status codes to ignore (default empty tuple)
 """
-function ec2_instance_metadata(path::AbstractString)
+
+function ec2_instance_metadata(path::AbstractString, ignored_codes::Tuple{Vararg{<:Integer}}=())
```
mattBrzezinski, line 219: Is there a reason for using a `::Tuple{Vararg{<:Integer}}` here? Why not just a list of integers?

src/AWSCredentials.jl
```diff
@@ -252,8 +261,8 @@ function ec2_instance_credentials(profile::AbstractString)
         source == "Ec2InstanceMetadata" || return nothing
     end
 
-    info = ec2_instance_metadata("/latest/meta-data/iam/info")
-    info === nothing && return nothing
+    info = ec2_instance_metadata("/latest/meta-data/iam/info", (404,))
```
mattBrzezinski, line 264: We would want to give some thoughts to the entry point for passing along this information. Not too sure how you're calling to get credentials currently, but if you're using these functions directly having a function argument for codes is fine.

mattBrzezinski (commented):
As a first pass this seems reasonable. Some discussion needed around how the entry points here will work, and some tests added in.

But definitely a good starting point!

src/AWSCredentials.jl
yalwan-sage: Sorry, habit of mine

src/AWSCredentials.jl
yalwan-sage: Preference for non-mutable data objects. Can change if you desire.

src/AWSCredentials.jl
yalwan-sage: So with this as it is, I could simply do something like `@service S3` and then start calling methods. I appreciate you probably don't want this hard coded here, but I didn't have a better idea

yalwan-sage: @mattBrzezinski  I'd like to revisit this -- any ideas on how you think this would be best passed along to the function from a design/API perspective? I couldn't see any obvious entry points from the top level user perspective because for example the config object isn't passed along to the `AWSCredentials` constructor. 

I also noticed that there is `_ec2_metadata` in `utilities/credentials.jl` which seems unused and might do something similar?

Issue #341, opened 2021-04-13
support kwargs for params without having to create a Dict


oxinabox:
Every function in the API takes a `param::Dict{Any,String}`

It would be useful if we could also accept normal julia keyword arguments
and have that get wrapped up into a `param` Dict automatically.
via `params = Dict{String, Any}(string(k)=>v for (k,v) in pairs(kwargs))`

mattBrzezinski:
Yes, this would be nice to do! One annoying thing is the usage of hyphens in optional parameters for instance `S3.list_buckets()` has a `max-keys` optional parameter (among others).

We would need to introduce some check to convert `_` to `-`, but otherwise this is very doable.

mattBrzezinski:
I'm looking into this now, and this won't be easily possible to do. The issue stems from certain services having argument names with hyphens, or underscores, or both.

For example, `SSO` has an argument called `x-amz-sso_bearer_token`. If we want this to be a kwarg we could call it `x_amz_sso_bearer_token`, but we then need  to convert this back to `x-amz-sso_bearer_token`, but its' not as simple as just converting underscores to hyphens as we only want to do this conversion for some.

We would essentially need to create some master list of `expected_name => valid_julia_symbol`, which seems like a lot of work.

AWS' inconsistencies strike again!

oxinabox:
What if we translated `-` into `_`, and `_` into `__` (double underscore)?

mattBrzezinski:
> What if we translated `-` into `_`, and `_` into `__` (double underscore)?

That works! I don’t really like going down this gross path but I suppose it’s the only real option.

oxinabox:
I think `var"foo-bar_qux"` might work for kwarg names but I haven't checked

omus:
How about this as an alternative? Use varargs to make an interface that is keyword-like but avoids having to mess with translation:

```julia
julia> f(aws_kwargs::Pair{String,<:Any}...) = aws_kwargs
f (generic function with 1 method)

julia> f("x-amz-sso_bearer_token" => "1234")
("x-amz-sso_bearer_token" => "1234",)
```

I think this is pretty reasonable from a user interface perspective and avoids having to learn the special translation logic when reading AWS documentation. The only real downside here is a little bit of processing required to be done.

mattBrzezinski:
> How about this as an alternative? Use varargs to make an interface that is keyword-like but avoids having to mess with translation:
> 
> ```julia
[...]

PR #518, opened 2021-11-17
Use Kwargs instead of Dict for optional parameters

mattBrzezinski:
## Overview
Resolves: #341 

This *is* a breaking change and would require the tagging of `AWS@2.0.0`. The changes here modify how optional parameters are passed into requests. Currently users would need to do;

```julia
using AWS
@service S3

S3.put_object(bucket, name, Dict("body" => "Hello world!")
```

With these proposed changes any optional parameters would be passed in as kwargs. The kwarg syntax will be defined in the docstr for each function as how to name them. Same example from above,

```julia
using AWS
@service S3

S3.put_object(bucket, name; body="Hello World!")
```

## Technical Changes
There are certain optional parameters which must be passed in that contain invalid Julia syntax characters. For example `max-keys`, to handle these we need to have a mapping of `Julia Syntax -> AWS Request Syntax`. When generating each high level definition we will keep track of all optional parameters for that service in this mapping.

These mappings are specific to each service, for example if we define the Julia `max_keys` this MAY NOT translate to `max-keys` in every service. Certain services may want this defined as `maxKeys` for example. There is an assumption built in here that this specification stops at the service level and not the operation level. I am not able to find a single example that breaks this assumption.

This mapping of `Julia -> AWS` syntaxes is stored in the `src/service/{service_name.jl` as as Dict and we will `amazonify(MAPPING, kwargs)` to translate any passed in kwargs to their appropriate names.

We use [aws-sdk-js](https://github.com/aws/aws-sdk-js/) here to create our definitions. There is a large gotcha here, where these definitions contain either lies, or a misunderstanding on my part. When a `locationName` is specified for an optional parameter this name maybe incorrect. This is not limited to a certain service, but is wide-spread. I've logged an issue for it here;

- https://github.com/aws/aws-sdk-js/issues/3892

And unsurprisingly have not received a response. I'm keeping track of these incorrect values in an `AWS_MAP_OF_LIES::Dict` and when we are generating our `Julia -> AWS` mappings we adjust for these incorrect values. This map is not comprehensive and will need to be dealt with on a case-by-case basis.

There is also another caveat here, when request `return_raw`, `return_stream`, etc (our built-in parameters) for specifying how our requests are dealt with we are assuming that these are NOT optional parameters which AWS themselves have defined. This *should* not be a problem, but it is one worth being aware of.

These changes are also limited to the high-level Julia AWS functions and not the low-level ones.


mattBrzezinski:
bors try

src/api_generation/high_level.jl
```diff
@@ -40,6 +59,8 @@ function _generate_high_level_wrapper(
              using AWS.AWSServices: $service_name
              using AWS.Compat
              using AWS.UUIDs
+
+             MAPPING = $(aws_service_param_mapping)
```
iamed2, line 63:
```suggestion
             const MAPPING = $(aws_service_param_mapping)
```
You don't expect to reassign `MAPPING` right?

Also I think it could use a better name. MAPPING just indicates its type and not its contents.

src/api_generation/utilities.jl
```diff
@@ -254,20 +254,47 @@ function _get_function_parameters(input::String, shapes::AbstractDict{String})
...
 end
+
+function add_aws_param_names!(aws_service_param_mapping, service_name, optional_parameters)
+    # AWS SDK lies, their API definitions say that they want certain capitalization however
+    # when making requests as they define, they do not work. They also do not respond back.
+
+    # https://github.com/aws/aws-sdk-js/issues/3892
+
+    # Keep track of all their fucking lies in this dictionary so requests will succeed.
+    # This is a mapping of Julia snake syntax to what AWS wants
```
iamed2, line 285: Please tone down the aggression here.

src/api_generation/high_level.jl
mattBrzezinski: I've added a comment describing the relationship and kept the same name. I feel like that's better than something more lengthy like `const JULIA_TO_AWS_SYNTAX_MAPPING`?

src/api_generation/utilities.jl
```diff
@@ -254,20 +254,48 @@ function _get_function_parameters(input::String, shapes::AbstractDict{String})
...
+    # AWS SDK JS has inconsistencies with their API definitions. Certain parameters specify
+    # various capitalizations however the requests themselves actually want different ones.
+
+    # https://github.com/aws/aws-sdk-js/issues/3892
+
+    # Keep track of the proper values in this mapping, using Julia syntax for the keys which
+    # users will specify in their request to the correct values which AWS wants in the
+    # requests.
+    AWS_INCONSISTENCIES = Dict(
+        "s3" => Dict("body" => "body"), "ec2" => Dict("client_token" => "ClientToken")
```
iamed2, line 288: IIUC, the S3 thing is different than EC2 and it's just another mechanism that we haven't yet considered. The fact that `"body"` is a parameter we pass through is entirely due to us [hard-coding](https://github.com/JuliaCloud/AWS.jl/blob/57696ba9a34dfff75f4c0dea3e42cba4c92e7001/src/AWS.jl#L263) special behaviour for the RestXMLService. 

They include the `"payload"` parameter in the service description which describes what gets put in the HTTP request body. So a conforming solution for S3 would be something like (pseudocode):

```julia
if "payload" in keys(endpoint)
    http_body = pop!(params, endpoint["payload"])
end
```

This appears to be the case for a number of services. 

*However*, the above is only correct for simple `"shape"`s. And in the majority of cases, the `"shape"` is of `"type": "structure"`. In that case we're actually supposed to pass a dictionary-like structure, and we don't want to strip off the key, we just want to do a normal `locationName` translation. This applies to endpoints like `CreateCloudFrontOriginAccessIdentityRequest` in `cloudfront-2018-11-05`. It would be interesting to see if that endpoint just breaks with our current system due to us hardcoding `body` as the source for `Request`'s `content` argument. 

---
All that to say I think it's valid to special case S3 here for now, but we should not consider it an inconsistency since it's really just something we're not handling properly yet. And we should also change the input argument name to `"Body"` to match both the spec and Boto3.

iamed2 (commented):
I think we should use the PascalCase form from the specs for the kwargs, like boto3 does. If we do that, we can be reasonably sure we won't get collisions between our special kwargs and AWS parameters.

src/api_generation/high_level.jl
iamed2: `const PARAMETER_MAPPING` would be better I think. Just any word describing what is being mapped. Having a comment in addition is also good though.

oxinabox:
> This is a breaking change and would require the tagging of AWS@2.0.0. The changes here modify how optional parameters are passed into requests. Currently users would need to do;


Can't we keep the old way as well and just deprecate it?

src/api_generation/high_level.jl
```diff
@@ -1,3 +1,21 @@
...
+
+function _parse_protocol(function_name::AbstractString, protocol::AbstractString)
+    if protocol in ("json", "query", "ec2")
+        return JSONBased()
+    elseif protocol in ("rest-xml", "rest-json")
+        return RestBased()
+    else
+        throw(
+            ProtocolNotDefined(
+                "$function_name is using a new protocol; $protocol which is not supported."
```
oxinabox, line 13: I am not sure the word "new" should be here.
since someone might for example be using this code many years after the protocol was added.
(say they were pinned to an old version for some reason)

idk maybe
```suggestion
                "$function_name is using an unexpected protocol; $protocol which is not supported."
```

src/AWS.jl
```diff
@@ -44,6 +44,23 @@ using ..AWSExceptions: AWSException
...
+    for kwarg in kwargs
+        str = string(kwarg.first)
+        _key = if haskey(mapping, str)
+            mapping[str]
+        else
+            str
+        end
+
+        new_kwargs[string(_key)] = kwarg.second
+    end
```
ericphanson, line 59: I think `.first` and `.second` is accessing the internal fields, so I'd do
```suggestion
    for (k, v) in kwargs
        str = string(k)
        _key = if haskey(mapping, str)
            mapping[str]
        else
            str
        end

        new_kwargs[string(_key)] = v
    end
```

src/api_generation/utilities.jl
```diff
@@ -254,20 +254,48 @@ function _get_function_parameters(input::String, shapes::AbstractDict{String})
...
     end
 
     return (sort(required_parameters), sort(optional_parameters))
 end
+
+function add_aws_param_names!(aws_service_param_mapping, service_name, optional_parameters)
+    # AWS SDK JS has inconsistencies with their API definitions. Certain parameters specify
+    # various capitalizations however the requests themselves actually want different ones.
+
+    # https://github.com/aws/aws-sdk-js/issues/3892
```
oxinabox, line 282: This link should be next to the particular mapping it is about.

src/api_generation/high_level.jl
ericphanson, line 13:
```suggestion
                "$function_name is using a new protocol $protocol which is not supported."
```

src/api_generation/high_level.jl
```diff
@@ -258,7 +303,7 @@ function _generate_high_level_definition(
         if !isempty(optional_parameters)
             operation_definition *= """
                 # Optional Parameters
-                Optional parameters can be passed as a `params::Dict{String,<:Any}`. Valid keys are:
+                Optional parameters can be passed as a keyword argument. Valid keys are:
```
ericphanson, line 306:
```suggestion
                Optional parameters can be passed as a keyword argument. Valid keyword arguments are:
```

ericphanson (commented):
Cool! I like the explicit `MAPPING` approach and that the keyword arguments show up in the docstrings, and `snake_case` looks pretty reasonable for the few that I spot-checked. I haven't read enough of the discussion about inconsistencies/locationNames/shapes etc to understand/comment on that stuff (although I think it's pretty important).

ericphanson:
Re- breakingness, we can probably use @omus's `FeatureSet` construction to make it opt-in: https://github.com/JuliaCloud/AWS.jl/blob/57696ba9a34dfff75f4c0dea3e42cba4c92e7001/src/AWS.jl#L47-L62

src/AWS.jl
```diff
@@ -44,6 +44,23 @@ using ..AWSExceptions: AWSException
...
+
+    for kwarg in kwargs
+        str = string(kwarg.first)
+        _key = if haskey(mapping, str)
+            mapping[str]
+        else
+            str
+        end
+
+        new_kwargs[string(_key)] = kwarg.second
```
oxinabox, line 58: I don't think  `(::Pair).first` and `(::Pair).second` is part of the public API -- fields generally are not.
Would it be clearer if we used iterator destructuring to put them in variables?
idk if `k` and `v` are the best names -- all the names here are unclear to me.
I don't really now what this function is doing.

Also perhaps clearer tpo use `get` with default
```suggestion
    for (k, v) in kwargs
        str = string(k)
        _key = get(mapping, str, str)
        new_kwargs[string(_key)] = v
```

oxinabox:
(My comments are pretty throw away, just making them in passing. I haven't reviewed this properly, and will leave that to people who understand the package)

src/services/athena.jl
```diff
@@ -4,9 +4,25 @@ using AWS.AWSServices: athena
 using AWS.Compat
 using AWS.UUIDs
 
+MAPPING = Dict(
```
oxinabox, line 7: These should be generated as `const`

We also might want co consider generating them as 
`LittleDict` that are backed by tuples, for maximum speed.
But this is not going to be the performance critical part anyway.

src/api_generation/high_level.jl
```diff
@@ -76,6 +99,8 @@ function _generate_high_level_definitions(
             )
```
oxinabox, line 99: I think we might want to think about pulling this section out into it's own function.
Such that above rather than having
```julia
aws_service_param_mapping, operations = _generate_high_level_definitions(
    service_name, protocol, operations, shapes
)
```
we have
```julia
service_details = high_level_service_details(service_name, protocol, operations, shapes)
operations = _generate_operations(service_details)
aws_service_param_mapping = _aws_param_names!(service_name, service_details)
```

I am not sure though.
Multiple return values feels like a little bit of a codesmell of a function that is doing too much.
but maybe it is doing one thing, it is just that that one thing is best expresed as two things.

mattBrzezinski:
> > This is a breaking change and would require the tagging of AWS@2.0.0. The changes here modify how optional parameters are passed into requests.
> 
> Can't we keep the old way as well and just deprecate it?

Yes we definitely can just add this in as another function that is callable for each of the services operations! Probably the path of least resistance.

src/services/athena.jl
mattBrzezinski: Oops @iamed2 suggested this, I just have not regenerated the code for the service files.

src/api_generation/utilities.jl
mattBrzezinski: I left it above as I'm not sure if this is currently limited to just EC2, I have a suspect that it is possibly more wide-spread but need to confirm.

iamed2: @mattBrzezinski My hunch is that the particular issue you filed is unique to EC2. I think this because it's the only service with a unique protocol. Using grep to grab protocols:

```
ericdavies@rutter ~/r/aws-sdk-js (master)> grep --no-filename -r \"protocol\"\:\ \" apis | cut -d\" -f4 | sort | uniq
ec2
json
query
rest-json
rest-xml
```

I suspect this case is "technically correct" because they've internally defined the EC2 protocol as "like query, but with weird capitalization".

src/services/athena.jl
iamed2: I think LittleDict is also a good idea

src/api_generation/utilities.jl
mattBrzezinski: Yes that's true that EC2 has it's own protocol. I'll dig around more later today to see if I can find more examples of these inconsistencies.

I'm sure botocore would also have these hard-coded workarounds as well.

iamed2:
> > > This is a breaking change and would require the tagging of AWS@2.0.0. The changes here modify how optional parameters are passed into requests.
> > 
> > 
> > Can't we keep the old way as well and just deprecate it?
> 
> Yes we definitely can just add this in as another function that is callable for each of the services operations! Probably the path of least resistance.

Maybe a kwarg to `@service` or something? Another benefit to having `@service` is people can upgrade independently.

src/api_generation/high_level.jl
```diff
@@ -1,3 +1,21 @@
+abstract type OperationType end
+struct RestBased <: OperationType end
```
omus, line 2:
```suggestion
struct RESTBased <: OperationType end
```

src/api_generation/high_level.jl
```diff
@@ -109,7 +134,8 @@ function _generate_high_level_definition(
     """
     Generate function definition for a service request given required, header and idempotent parameters.
     """
-    function _generate_rest_operation_defintion(
+    function _generate_operation_definition(
+        protocol::RestBased,
```
omus, line 138: Seems like we don't really need an instance at all can just do 
```suggestion
        protocol::Type{RestBased},
```

src/api_generation/high_level.jl
```diff
@@ -40,6 +59,9 @@ function _generate_high_level_wrapper(
              using AWS.AWSServices: $service_name
              using AWS.Compat
              using AWS.UUIDs
+
+             # Julia syntax for service-level optional parameters to the AWS request syntax
+             const MAPPING = $(aws_service_param_mapping)
```
omus, line 64:
```suggestion
             const SERVICE_PARAMETER_MAP = $(aws_service_param_mapping)
```

src/services/accessanalyzer.jl
```diff
@@ -49,7 +50,7 @@ function apply_archive_rule(
                 Dict{String,Any}(
                     "analyzerArn" => analyzerArn,
                     "ruleName" => ruleName,
-                    "clientToken" => string(uuid4()),
+                    "client_token" => string(uuid4()),
```
omus, line 53: ❗ As `params` has already been "amazonified" this should be "clientToken" and not "client_token". Looks like something is wrong in the code generation

src/services/accessanalyzer.jl
```diff
@@ -163,31 +140,16 @@ Creates an analyzer for your account.
...
 
 # Optional Parameters
-Optional parameters can be passed as a `params::Dict{String,<:Any}`. Valid keys are:
-- `"archiveRules"`: Specifies the archive rules to add for the analyzer. Archive rules
+Optional parameters can be passed as a keyword argument. Valid keys are:
+- `"archive_rules"`: Specifies the archive rules to add for the analyzer. Archive rules
   automatically archive findings that meet the criteria you define for the rule.
-- `"clientToken"`: A client token.
+- `"client_token"`: A client token.
 - `"tags"`: The tags to apply to the analyzer.
```
omus, line 147: Should be rendered as:
```
# Keyword Parameters
- `archiveRules`: Specifies the archive rules to add for the analyzer. Archive rules
Optional parameters can be passed as a keyword argument. Valid keys are:
- `archive_rules`: Specifies the archive rules to add for the analyzer. Archive rules
  automatically archive findings that meet the criteria you define for the rule.
- `clientToken`: A client token.
- `tags`: The tags to apply to the analyzer.
```
We definitely don't need quotes around the keywords and the preface is no longer needed

src/api_generation/high_level.jl
```diff
@@ -258,7 +303,7 @@ function _generate_high_level_definition(
         if !isempty(optional_parameters)
             operation_definition *= """
                 # Optional Parameters
```
omus, line 305:
```suggestion
                # Keyword Parameters
```

src/api_generation/high_level.jl
omus, line 306: Preface is no longer needed as these are just standard keywords and don't really need to be explained this much:
```suggestion
```

test/AWS.jl
```diff
@@ -566,18 +566,13 @@ end
...
+            secret_name; secret_string=secret_string, client_request_token=string(uuid4())
         )
 
         try
             @test _get_secret_string(secret_name) == secret_string
         finally
-            Secrets_Manager.delete_secret(
-                secret_name, LittleDict("ForceDeleteWithoutRecovery" => "true")
-            )
+            Secrets_Manager.delete_secret(secret_name; force_delete_without_recovery="true")
```
omus, line 575: What happens if we pass in a `Bool` here?

test/AWS.jl
```diff
@@ -837,7 +832,7 @@ end
         try
             # PUT with parameters operation
             body = "sample-file-body"
-            S3.put_object(bucket_name, file_name, Dict("body" => body))
+            S3.put_object(bucket_name, file_name; body=body)
```
omus, line 835: Consider doing:
```suggestion
            @compat S3.put_object(bucket_name, file_name; body)
```

test/AWS.jl
```diff
@@ -944,23 +939,22 @@ end
 
         try
             # POST
-            tags = Dict("Tags" => LittleDict("Tag-01" => "Tag-01", "Tag-02" => "Tag-02"))
+            tags = Dict("Tag-01" => "Tag-01", "Tag-02" => "Tag-02")
 
             for vault in vault_names
-                Glacier.add_tags_to_vault("-", vault, tags)
+                Glacier.add_tags_to_vault("-", vault; Tags=tags)
```
omus, line 945: Appears this is set incorrectly according to the glacier mapping:
```suggestion
                Glacier.add_tags_to_vault("-", vault; tags=tags)
```

omus: If this isn't failing we should probably investigate why... Thinking about it more it's because `amazonify` just translates names and doesn't cause a failure if you use the actual names used in the `params` dict. I don't think we should support both and should fail if a key is provided that isn't in the mapping

omus (changes requested):
Very promising. A few more iterations are needed yet.
- Be

src/api_generation/high_level.jl
omus: I like `SERVICE_PARAMETER_MAP`. It doesn't matter much if it's wordy

src/api_generation/high_level.jl
iamed2: Instances are better for the compiler and easier to read.

iamed2:
I want to make sure this gets discussed and not buried under suggestions so I'm just going to copy and paste:

I think we should use the PascalCase form from the specs for the kwargs, like boto3 does. If we do that, we can be reasonably sure we won't get collisions between our special kwargs and AWS parameters.

src/api_generation/high_level.jl
mattBrzezinski: This is used for generating high-level APIs. End users shouldn't care for this, as it only matters for maintainers of this package.

test/AWS.jl
mattBrzezinski: In the couple of lines above we were passing in,

> tags = Dict("Tags" => LittleDict("Tag-01" => "Tag-01", "Tag-02" => "Tag-02"))

This worked well because Glacier was expecting `Tags` at the top-level value in the structure. Now however since we're using snake_case kwargs we have a translation inside of `src/services/glacier.jl` that specifies,

> tags => Tags

So we need to pass in this structure as `tags=tags`, this will then convert the high-level object to `Tags` and continue along as it previously worked when making the request.

mattBrzezinski: @iamed2 This is another example of inconsistencies in the operation definitions. Every other Glacier optional parameter is specified in camel case EXCEPT for `Tags`.

```julia
"part_size" => "x-amz-part-size",
"job_parameters" => "jobParameters",
"archive_size" => "x-amz-archive-size",
"checksum" => "x-amz-sha256-tree-hash",
"range" => "Range",
"completed" => "completed",
"archive_description" => "x-amz-archive-description",
"body" => "body",
"policy" => "policy",
"vault_notification_config" => "vaultNotificationConfig",
"tag_keys" => "TagKeys",
"statuscode" => "statuscode",
"marker" => "marker",
"tags" => "Tags",
"limit" => "limit",
```

test/AWS.jl
mattBrzezinski: Works both ways! I don't have an example off hand but I did remember times where requests would fail for certain services.

I remember passing in a `string(1)` or `1` and it failed and it needed to be the opposite type. I can set this as a `Bool` going forward, but things *can* be finicky.

test/AWS.jl
iamed2: Sure but in this case the difference is clearly communicated in the service spec, and there's nothing special necessary to handle this case as it follows the standard `rest-json` protocol.

mattBrzezinski: Correct, but in the current implementation in this package you as a user need to know that this specific parameter is the only one with a capitalization. If you make a request now you'll get some vague error back from AWS saying that the structure is malformed.

Going forward with these changes or like-wise ones, it's not a problem. But it's more-so to just highlight that there are inconsistencies all throughout the SDK, and it wouldn't surprise me if there's another service in a situation like EC2.

iamed2: > Correct, but in the current implementation in this package you as a user need to know that this specific parameter is the only one with a capitalization.

This is clear in the SDK docs though. I don't know how a user would expect to know what each argument is called without checking the docs. And it's consistent between boto3 and the PHP and JS SDKs, which both match how we pass arguments (as kwargs or a dict). The SDKs which are different (Go, C++, Ruby, Java) standardize casing across every type of service. The Ruby SDK is notably the only official SDK which passes a generic map of arguments AND standardizes casing.

mattBrzezinski: I don't think it's a stretch to assume that if the parameter is `tags` for one service it would be that for another and not `Tags`.

My issue here is that all of these inconsistencies don't inspire confidence that this package will just work out of the box for every service.

test/AWS.jl
mattBrzezinski: Ah here's an example,

https://github.com/JuliaCloud/AWS.jl/blob/57696ba9a34dfff75f4c0dea3e42cba4c92e7001/test/AWS.jl#L961

test/AWS.jl
iamed2: > My issue here is that all of these inconsistencies don't inspire confidence that this package will just work out of the box for every service.

But as we progress to implementing the specs more correctly, we make progress. So far there has been exactly one case (EC2) where the spec did not appear to match required behaviour. This glacier situation was an example of the package working perfectly. This PR changed what the API expected, and so we needed to change what the tests passed.

mattBrzezinski:
Coming back around to this, more though needs to be put in here... I was under the assumption that at a service level each optional parameter has the same mapping. 

In this example with S3. The `GetObjectRequest` has the optional parameter `VersionId` with a location name `verisonId`. However the `WriteGetObjectResponseRequest` has this as `x-amz-fwd-header-x-amz-version-id`.

I guess this can now become a function level? Quite sad.

mattBrzezinski:
Closing as this is lingering and not being addressed.

PR #522, opened 2021-11-25
Refactor requests write to `response_stream` once

omus:
Follow up to: https://github.com/JuliaCloud/AWS.jl/pull/516#discussion_r756212793

By making the repeat try/catch the innermost one we can eliminate having to keep track of the number of attempts.

### DIFF
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -209,28 +209,28 @@ try
         end
 
         n = 100
-        n_incomplete = n - 1
         data = rand(UInt8, n)
 
         config = AWSConfig(; creds=nothing)
 
         @testset "Fail 2 attempts then succeed" begin
-            apply(_incomplete_patch(; data=data, num_attempts_to_fail=2)) do
-                resp = S3.get_object("www.invenia.ca", "index.html"; aws_config=config) # use public bucket as dummy
-                @test length(resp) == n
-                @test resp == data
+            apply(_incomplete_patch(; data = data, num_attempts_to_fail = 2)) do
+                retrieved = S3.get_object("www.invenia.ca", "index.html"; aws_config=config) # use public bucket as dummy
+
+                @test length(retrieved) == n
+                @test retrieved == data
             end
         end
         @testset "Fail all 4 attempts then throw" begin
-            apply(_incomplete_patch(; data=data, num_attempts_to_fail=4)) do
-                err_t = if AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend
-                    HTTP.IOError
-                else
-                    Downloads.RequestError
-                end
-                @test_throws err_t S3.get_object(
-                    "www.invenia.ca", "index.html"; aws_config=config
-                ) # use public bucket as dummy
+            io = IOBuffer()
+            apply(_incomplete_patch(; data = data, num_attempts_to_fail = 4)) do
+                err_t = AWS.DEFAULT_BACKEND[] isa AWS.HTTPBackend ? HTTP.IOError : Downloads.RequestError
+                @test_throws err_t S3.get_object("www.invenia.ca", "index.html", Dict("response_stream" => io); aws_config=config) # use public bucket as dummy
+
+                seekstart(io)
+                retrieved = read(io)
+                @test length(retrieved) == n - 1
+                @test retrieved == data[1:n - 1]
             end
         end
     end
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -155,8 +155,8 @@ try
         end
     end
 
+    # https://github.com/JuliaCloud/AWS.jl/issues/515
     @testset "issue 515" begin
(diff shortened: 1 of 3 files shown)


omus:
I'll call out I've removed the special EOF logic from #516 just to validate it is still needed (I think it is)

omus:
bors try

omus:
I'm concerned the tests from #516 don't actually test the EOF behaviour noticed in #515

omus:
bors try

omus:
@IanButterworth can you review this? I removed the EOF logic which stops the `response_stream` from being written to on the last failure but the tests you added didn't fail. Do you actually need that behaviour?

omus:
bors try

omus:
bors try

omus:
bors try

omus:
bors try

test/issues.jl
```diff
@@ -190,22 +196,36 @@ try
...
+            io = IOBuffer()
+
+            apply(_incomplete_patch(; data=data, num_attempts_to_fail=4)) do
+                params = Dict("response_stream" => io)
+                @test_throws err_t S3.get_object(bucket, key, params; aws_config=config)
+
+                seekstart(io)
+                retrieved = read(io)
+                @test length(retrieved) == n - 1
+                @test retrieved == data[1:(n - 1)]
```
omus, line 228: Note this is exactly the behaviour produced from #516 and is probably not what was intended. The reason this is the case is that for the check `should_write || attempt == max_attempts` the variable `should_write = false` but `attempt = 4` which is equivalent to `max_attempts`.

I personally think we should probably keep the behaviour of having the final attempt always written to `response_stream` even if the result is partial. As an exception is raised already the user has been signalled there's a problem with the data and keeping the partial data could be useful in the case the result is an AWS exception.

IanButterworth:
> I removed the EOF logic which stops the response_stream from being written to on the last failure but the tests you added didn't fail. Do you actually need that behaviour?

I don't see any issue with how this has been re-written. AFAICT, it's a different way of achieving the same behavior: Retry failures aren't passed to the `response_stream` except for the last failed attempt.

The only functional difference I can tell is that on master `response_stream` is written to in all but the type of errors thrown my premature EOF in fixed content length responses. In this PR any error on an attempt stops the buffer from being passed to `response_stream`. 

The reason #516 specifically only prevented the buffer forwarding for #515-related errors was the suggestion somewhere in the reviews (or a slack discussion, I can't remember) that the existing behavior might be needed for some other response type?

If that's not the case, this PR looks good to me

test/issues.jl
IanButterworth: That sounds good to me. 

In my experience testing #516 the network intermittent issues causing #515 rarely (perhaps never) happened multiple times in a row, and if that happened such that a attempts are exhausted and an error throws, then passing the incomplete content forward seems fair.

omus:
> The only functional difference I can tell is that on master `response_stream` is written-to in all but the type of errors thrown my premature EOF in fixed content length responses. In this PR _any_ error on an attempt stops the buffer from being passed to `response_stream`.

A quick correction: this PR always writes the buffer from the final attempt to the `response_stream` even if the last attempt was a failure or an incomplete transfer. 

> The reason #516 specifically only prevented the buffer forwarding for #515-related errors was the suggestion somewhere in the reviews (or a slack discussion, I can't remember) that the existing behavior might be needed for some other response type?
> 
> If that's not the case, this PR looks good to me

I'll call out that we probably do want to forward the buffer when errors occur, even EOF errors, as there may relevant information included in that content. The specific example I'm thinking of is an AWS exception where the information is all included in the content so even if we have a incomplete transfer we should pass that information along. Also worth mentioning we already throw an exception when an EOF event occurs so the caller should be adequately notified of the situation.

IanButterworth:
>  this PR always writes the buffer from the final attempt to the response_stream even if the last attempt was a failure or an incomplete transfer.

Yeah, that's what master does too.

IanButterworth:
Ok, I was mistaken.. I thought the `should_write = true` was in the try, so it's state wasn't being reset.

test/issues.jl
IanButterworth: The error is propagated, is just whether the incomplete data is also written to the buffer.

IanButterworth:
friendly bump

omus:
I was waiting on a review from @mattBrzezinski but I think we're safe to merge this.

omus:
@mattBrzezinski can you review this? It needs approval. I'll use my admin privileges to merge this at EOD tomorrow otherwise

omus:
It's important to get this merged soon as otherwise in [some cases AWS exceptions will be suppressed](https://github.com/JuliaCloud/AWS.jl/pull/522#issuecomment-980044166). I was hoping this was going to be merged before another release occurred ([v1.73.0 was recently released](https://github.com/JuliaCloud/AWS.jl/pull/523))

Issue #529, opened 2022-01-04
when region is missing


jondan99:
Hi All

I spent a lot of time, unsuccessfully, trying to make AWS.jl work. I could the command line aws just fine, but not AWS.jl. Finally I discovered why, I did not have region set in .aws/config  and that was never a problem on the command line.

So, I would suggest adding a check for region and inform the user, and perhaps allow one to add region to requests.

all the best, Jack

mattBrzezinski:
We actually set the default region if one is not available to `us-east-1`. Could you provide some code example of what you were trying to do here?

jondan99:
Hi 

thanks for the reply.

My region = eu-west-1 and I was accessing sns and s3

When I set region  in .aws/config the below works fine 

SNS.publish("test",Dict("Subject" => "Julia", "TargetArn"  =>  "my arn..."))
s3("GET", "/my-bucket...")

when region is not set both sns and s3 fail with these errors (and the "Invalid parameter: TargetArn" is particularly confusing) 

SNS.publish("test",Dict("Subject" => "Julia", "TargetArn"  =>  "my arn..."))
ERROR: AWS.AWSExceptions.AWSException: InvalidParameter -- Invalid parameter: TargetArn
HTTP.ExceptionRequest.StatusError(400, "POST", "/", HTTP.Messages.Response:
"""
HTTP/1.1 400 Bad Request
x-amzn-RequestId: 111b2190-0c8f-5bb6-8434-7bde9cf97c1e
Content-Type: text/xml
Content-Length: 275
Date: Tue, 04 Jan 2022 18:13:10 GMT

[Message Body was streamed]""")

<ErrorResponse xmlns="http://sns.amazonaws.com/doc/2010-03-31/">
  <Error>
    <Type>Sender</Type>
    <Code>InvalidParameter</Code>
    <Message>Invalid parameter: TargetArn</Message>
  </Error>
  Stacktrace:
  [1] macro expansion
  .....



s3("GET", "/my-bucket")ERROR: AWS.AWSExceptions.AWSException: 301 -- AWSException
HTTP.ExceptionRequest.StatusError(301, "GET", "/my-bucket", HTTP.Messages.Response:
"""
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: eu-west-1
x-amz-request-id: GZQEQ1K0MZBG3XSZ
[...]

PR #530, opened 2022-01-04
Added docs for high-level kwarg passing, fixed docs for AWSConfig kwargs

mattBrzezinski:
Stemming from discussions in #529, I've added some documentation to the README about how high-level functions are wrapped around low-level ones and you can pass along those `kwargs`.

There were also typos in the docs for the kwarg `aws_config` which I have updated to reflect it's proper name.


morris25 (approved):
LGTM!

PR #543, opened 2022-04-18
Retry certain errors more times

ericphanson:
Closes #542
Closes #541

Issue #542, opened 2022-04-17
Increase number of retries for 500 errors?


ericphanson:
I had a long running job go down with
```julia
HTTP.ExceptionRequest.StatusError(500, "GET", "/bucket/key", HTTP.Messages.Response:
    """
    HTTP/1.1 500 Internal Server Error
    x-amz-request-id: 9E4ZYFR318DFC20R
    x-amz-id-2: ftAYyOnFknM17wNC+01olBbN8Gj7O1vY7FdIQ6Ap/tF2ge2ndN5waZzu12+QyLtaqM1/2YQ70GU=
    content-type: application/xml
    transfer-encoding: chunked
    date: Sun, 17 Apr 2022 04:51:19 GMT
    server: AmazonS3
    connection: close
    
    [Message Body was streamed]""")
    
    <?xml version="1.0" encoding="UTF-8"?>
    <Error><Code>InternalError</Code><Message>We encountered an internal error. Please try again.</Message><RequestId>9E4ZYFR318DFC20R</RequestId><HostId>ftAYyOnFknM17wNC+01olBbN8Gj7O1vY7FdIQ6Ap/tF2ge2ndN5waZzu12+QyLtaqM1/2YQ70GU=</HostId></Error>
    
    Stacktrace:
      [1] _http_response(req::Request, res::Downloads.Response; throw::Bool)
        @ AWS ~/.julia/packages/AWS/5goc1/src/utilities/downloads_backend.jl:111
      [2] macro expansion
        @ ~/.julia/packages/AWS/5goc1/src/utilities/downloads_backend.jl:85 [inlined]
      [3] macro expansion
        @ ~/.julia/packages/Retry/vS1bg/src/repeat_try.jl:192 [inlined]
      [4] _http_request(backend::AWS.DownloadsBackend, request::Request, response_stream::IOBuffer)
        @ AWS ~/.julia/packages/AWS/5goc1/src/utilities/downloads_backend.jl:63
      [5] macro expansion
        @ ~/.julia/packages/Mocking/MsKoy/src/mock.jl:29 [inlined]
      [6] macro expansion
        @ ~/.julia/packages/AWS/5goc1/src/utilities/request.jl:88 [inlined]
      [7] macro expansion
        @ ~/.julia/packages/Retry/vS1bg/src/repeat_try.jl:192 [inlined]
      [8] submit_request(aws::AWSConfig, request::Request; return_headers::Nothing)
[...]

PR #548, opened 2022-05-11
translate Retry to `Base.retry`

ericphanson:
Suggested in https://github.com/JuliaCloud/AWS.jl/issues/542#issuecomment-1100890974

I'm not sure we should merge this as-is, because before we could do immediate retries OR delay retries, and now we can only do delay retries. ~~I'm also not 100% sure the delay amounts are the same.~~

The functional approach is nicer than the macro approach though in my opinion, ~~since the retries don't need to be literals~~. So we could let folks request more retries on a per-request basis, or even put it in the Backend struct.

edit 1: Retry.jl expands to a for-loop, so non-literals are OK. I thought it unrolled the loop.

edit 2: the delay amounts are definitely not the same, but that's a good thing- now we copy the upstream recommendation from https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html. In particular, putting a cap of 20s means we can increase the retries further without danger, xref #550 (we can't do that with Retry.jl since the retries are hard-coded with no cap and get too big- ref https://github.com/JuliaCloud/AWS.jl/pull/543#issuecomment-1125034984).

---
12 May update:

I'm now in favor of merging this, because I want to get in #550. The first few retries are pretty short, e.g.
```julia
julia> collect(AWS.AWSExponentialBackoff(; max_attempts=10))
9-element Vector{Any}:
  0.3365311959885935
  1.2370247946870188
  2.925043257220695
 12.165245886682294
 20.0
 20.0
  5.5969335490336505
 20.0
 20.0
```
So I think having a short delay instead of an immediate retry in a few cases is probably OK. Also, the AWS docs don't talk about non-delayed retries, they seem to only do delayed retries.


ericphanson:
Going to check on CI:

bors try

ericphanson:
bors try

ericphanson:
bors try

.JuliaFormatter.toml
```diff
@@ -0,0 +1 @@
+style="blue"
```
ericphanson, line 1: This lets me use "format" command in VSCode

ericphanson:
bors try

ericphanson:
bors try

src/utilities/downloads_backend.jl
```diff
@@ -76,8 +76,9 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
 
     local buffer
     local response
-    try
-        @repeat 4 try
+
+    get_response =
+        () -> begin
```
omus, line 81: May as well use the long-form anonymous function syntax
```suggestion
    get_response = function ()
```
Even better option:
```suggestion
    try
        retry(; check, delays) do
```
This eliminates the need to create `get_response_with_retry` and avoid the nasty double call

src/utilities/downloads_backend.jl
```diff
@@ -100,13 +101,22 @@ function _http_request(backend::DownloadsBackend, request::Request, response_str
...
-                isa(e, Downloads.RequestError)
-            )
-            end
+            # We'll rely on lexical scoping; `buffer` and `response`
+            # are bindings in the outer scope, so we don't need to return here.
+            return nothing
+        end
+
+    check =
+        (s, e) -> begin
```
omus, line 110:
```suggestion
    check = function (s, e)
```

src/utilities/request.jl
```diff
@@ -116,57 +116,84 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
         end
+    end
 
-        # Handle ExpiredToken...
-        # https://github.com/aws/aws-sdk-go/blob/v1.31.5/aws/request/retryer.go#L98
-        @retry if e isa AWSException && e.code in EXPIRED_ERROR_CODES
-            check_credentials(credentials(aws); force_refresh=true)
-        end
+    check =
+        (s, e) -> begin
```
omus, line 152:
```suggestion
    check = function (s, e)
```

src/utilities/request.jl
```diff
@@ -206,17 +234,30 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
+                   isa(e, Base.IOError) ||
+                   (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
         end
+
+    get_response_with_retry = retry(
+        get_response; check=check, delays=AWSExponentialBackoff(; max_attempts=4)
+    )
+
+    try
+        get_response_with_retry()
```
omus, line 260:
```suggestion
    delays=AWSExponentialBackoff(; max_attempts=4)
    
    try
        retry(; check, delays) do
            get_response()   # Inline this function
        end
```

src/utilities/request.jl
```diff
@@ -206,17 +234,30 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
             )
-        catch e
+
+            # We'll rely on lexical scoping; `buffer` and `response`
+            # are bindings in the outer scope, so we don't need to return here.
+            return nothing
+        end
+
+    check =
+        (s, e) -> begin
```
omus, line 244:
```suggestion
    check = function (s, e)
```

src/utilities/request.jl
```diff
@@ -116,57 +116,84 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
+            end
+            return false
         end
-    end
+
+    retry(
+        upgrade_error(get_response);
+        delays=AWSExponentialBackoff(; max_attempts=3),
+        check=check,
+    )()
```
omus, line 196:
```suggestion
    delays = AWSExponentialBackoff(; max_attempts=3)
    
    retry(; check, delays) do
        # Inline both of these functions
        upgrade_error() do
            get_response()
        end
    end
```

src/utilities/utilities.jl
```diff
@@ -139,3 +139,26 @@ end
...
+# https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html
+struct AWSExponentialBackoff
+    max_attempts::Int
+    max_backoff::Float64
+end
+
+# Defaults for AWS's `standard` retry mode. Note: these can be overridden elsewhere.
+function AWSExponentialBackoff(; max_attempts=3, max_backoff=20.0)
+    return AWSExponentialBackoff(max_attempts, max_backoff)
+end
```
omus, line 152: If you're comfortable with using `@kwdef`:
```suggestion
# Default values for AWS's `standard` retry mode. Note: these can be overridden elsewhere.
Base.@kwdef struct AWSExponentialBackoff
    max_attempts::Int = 3
    max_backoff::Float64 = 20.0
end
```

src/utilities/utilities.jl
```diff
@@ -139,3 +139,26 @@ end
...
+
+# We make one more attempt than the number of delays
+Base.length(exp::AWSExponentialBackoff) = exp.max_attempts - 1
+
+function Base.iterate(exp::AWSExponentialBackoff, i=1)
+    i >= exp.max_attempts && return nothing
+    # rand() has values in [0, 1), so we use 1.0 - rand() which has values in (0, 1] required.
+    b = 1.0 - rand()
```
omus, line 160: Should pass in an `rng` into this for testing purposes
```suggestion
    b = 1.0 - rand(exp.rng)
```

omus (commented):
Comments mainly around formatting. Otherwise seems reasonable as long as the tests pass

src/utilities/request.jl
```diff
@@ -206,16 +236,28 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
...
+                   isa(e, Base.IOError) ||
+                   (isa(e, HTTP.StatusError) && _http_status(e) >= 500)
+        end
+
+    delays=AWSExponentialBackoff(; max_attempts=4)
+    
+    try
+        retry(; check, delays) do
+            get_response()   # Inline this function
         end
```
ericphanson, line 261: @omus I think we would need
```suggestion
        retry(; check, delays) do
            get_response()   # Inline this function
        end()
```
or
```suggestion
        retry(get_response; check, delays)()
```
still, since the `do`-block is just constructing `retry(() -> get_response(); check, delays)` here. (Also, we need to use the old keyword arg syntax or `@compat`)

ericphanson:
bors try

ericphanson:
locally, I can load AWS on julia 1.3.1... not sure why CI has an issue

ericphanson:
maybe we can try again...

bors try

ericphanson:
bors try

src/utilities/request.jl
```diff
@@ -116,46 +116,64 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
+
+            if response.status in REDIRECT_ERROR_CODES
+                if HTTP.header(response, "Location") != ""
+                    request.url = HTTP.header(response, "Location")
+                else
+                    e = HTTP.StatusError(response.status, response)
+                    throw(AWSException(e, stream))
+                end
             end
         end
```
omus, line 137:
```suggestion
    get_response = function ()
        credentials(aws) === nothing || sign!(aws, request)

        aws_response = @mock _http_request(request.backend, request, stream)
        response = aws_response.response

        if response.status in REDIRECT_ERROR_CODES
            if HTTP.header(response, "Location") != ""
                request.url = HTTP.header(response, "Location")
            else
                e = HTTP.StatusError(response.status, response)
                throw(AWSException(e, stream))
            end
        end
    end
```

src/utilities/request.jl
```diff
@@ -116,46 +116,64 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
+        () -> try
+            return f()
+        catch e
+            if e isa HTTP.StatusError
+                e = AWSException(e, stream)
+                rethrow(e)
+            end
+            rethrow()
         end
+    end
```
omus, line 149:
```suggestion
    function upgrade_error(f)
        try
            return f()
        catch e
            if e isa HTTP.StatusError
                e = AWSException(e, stream)
                rethrow(e)
            end
            rethrow()
        end
    end
```

src/utilities/request.jl
```diff
@@ -166,8 +184,13 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
                 ),
             )
         end
+        return false
     end
 
+    delays = AWSExponentialBackoff(; max_attempts=3)
+
+    retry(upgrade_error(get_response); check=check, delays=delays)()
```
omus, line 192: Requires that `() ->` is removed from `upgrade_error`
```suggestion
    retry(upgrade_error ∘ get_response; check=check, delays=delays)()
```

test/utilities.jl
```diff
@@ -79,3 +79,27 @@ end
...
+    return c
+end
+
+@testset "AWSExponentialBackoff" begin
+    for (n, max_backoff) in [(3, 5.0), (10, 20.0)]
+        itr = AWS.AWSExponentialBackoff(;
+            max_attempts=n, max_backoff=max_backoff, rng=StableRNG(1)
+        )
+        @test count_len(itr) == n - 1
+        @test length(collect(itr)) == n - 1
```
omus, line 101: Not sure why we need `count_len` if we have the `length(collect(itr))` test

ericphanson: `collect` uses `Base.length`, so if you implement `Base.length` wrong (too short), you won't know it with a `collect` test

src/utilities/request.jl
ericphanson: This doesn't work, since `upgrade_error` now calls its input as a function. So we do `result = get_response()`, then `ugrade_error(result)`. I think we really need the function-wrapper approach we had before to do this.

ericphanson:
bors try

ericphanson:
Ok! I don't really want to do more formatting changes since it's kind of pain to get it through Bors again. Are there any other changes we should make here? Or is it good to go?

src/utilities/request.jl
omus: You are correct. I seem to be having an off day

test/utilities.jl
```diff
@@ -79,3 +79,27 @@ end
...
+function count_len(itr)
+    c = 0
+    result = iterate(itr)
+    while result !== nothing
+        c += 1
+        state = result[2]
+        result = iterate(itr, state)
+    end
+    return c
+end
```
omus, line 93: As the [`for` loop interface uses iteration under the hood](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration) we can simply this to:
```suggestion
# Count the elements in an iterator without using `length`
function count_len(itr)
    c = 0
    for _ in itr
        c += 1
    end
    return c
end
```
If you really want to use the iteration iterface directly let's stick to the more canonical variable names:
```suggestion
# Count the elements in an iterator without using `length`
function count_len(itr)
    c = 0
    next = iterate(itr)
    while next !== nothing
        c += 1
        (item, state) = next
        next = iterate(itr, state)
    end
    return c
end
```

omus (approved):
LGTM

test/utilities.jl
omus: That's exactly what I needed to know. Made one more suggestion: https://github.com/JuliaCloud/AWS.jl/pull/548#discussion_r877542741

Issue #552, opened 2022-05-12
Do not re-use HTTP connection when retrying a transient error


ericphanson:
Ref
> Most AWS SDKs use an HTTP library that uses connection pools so that you can reuse an existing connection when making an HTTP request. Generally, connections are reused when retrying requests due to throttling errors. Requests are not reused when retrying due to transient errors.

> Transient errors (HTTP status codes 400, 408, 500, 502, 503, and 504)

from https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html.

I believe this may be an alternate/better path to solving #542 and #541, which we are currently trying to address with #550. These are both transient errors that I hit, which we currently retry 4 times, BUT crucially our retries use the same connection pool. Perhaps more retries isn't the answer, but rather better retries- ones with new connections.

---

Each backend will probably need to support this differently. For Downloads.jl, I believe we can use a new Downloader object, ref https://github.com/JuliaLang/Downloads.jl/blob/20e099026481d3b893daf8ba5258eaa2c4d13831/src/Downloads.jl#L31 by swapping it out in the `catch` part of Retry.jl's block, or in the `check` function of `Base.retry` (ref #548).

It's easy to "just do" this, but it exposes somewhat of an API flaw: the `DownloadsBackend` object holds a single downloader, meaning the user may supplied us with a given Downloader they expect us to use (perhaps one that they have specially configured with an `easy_hook`). However, we now see that there are times from within AWS.jl that we know we should try a new Downloader object, e.g. if we hit a transient error.

I am not really sure what to do. I thought the downloader should be settable per-request  so one could try to do stuff like have different Downloader objects for different threads, and for maximum customizability. I haven't really been using that though; not sure if anyone has. Perhaps the user should not only supply a `Downloader` but also a function to create a new one if needed. Then we can use the one they supply until we get a transient error and then we create a new one. Perhaps the `DownloadsBackend` should be mutable so we can swap out the downloader there.

Would love to know folks thoughts on this!

PR #553, opened 2022-05-13
New Downloader for transient retries

ericphanson:
fixes https://github.com/JuliaCloud/AWS.jl/issues/552 for the Downloads backend.

Based on #548 

It's a little awkward, because we have retries at 2 levels- on the level of the HTTP request (via whichever backend), and in `submit_request`. We don't have AWSExceptions until the 2nd layer, where we actually read from the stream. In the first layer, we don't read from the stream yet. So we need a way to tell `_http_request` that we are retrying from a transient error, so I added a new keyword argument.

Would really appreciate any thoughts.


src/utilities/request.jl
```diff
@@ -100,16 +100,23 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
         "ThrottledException",
         "RequestThrottledException",
         "TooManyRequestsException",
         "ProvisionedThroughputExceededException",
+        "TransactionInProgressException",
+        "RequestLimitExceeded",
+        "BandwidthLimitExceeded",
         "LimitExceededException",
         "RequestThrottled",
-        "PriorRequestNotComplete",
```
ericphanson, line 112: This is a transient error, not a throttling one, according to https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes-standard.title

However, in this PR only Downloads backend is getting special transient error handling, so maybe I should add this back with a comment, so we don't mess up the HTTP behavior.

ericphanson: I added it later on as a separate check

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
bors try

ericphanson:
bors try

src/utilities/downloads_backend.jl
```diff
@@ -31,13 +32,14 @@ const AWS_DOWNLOAD_LOCK = ReentrantLock()
 # because we add a hook to avoid redirects in order to try to match the HTTPBackend's
 # implementation, and we don't want to mutate the global downloader from Downloads.jl.
 # https://github.com/JuliaLang/Downloads.jl/blob/84e948c02b8a0625552a764bf90f7d2ee97c949c/src/Downloads.jl#L293-L301
-function get_downloader(downloader=nothing)
+function get_downloader(; fresh=false)
+    downloader = nothing
     lock(AWS_DOWNLOAD_LOCK) do
         yield() # let other downloads finish
         downloader isa Downloader && return nothing
```
omus, line 39: Condition will never be true
```suggestion
```

src/utilities/downloads_backend.jl
```diff
@@ -6,22 +6,23 @@ using HTTP.MessageRequest: body_was_streamed
...
 struct DownloadsBackend <: AWS.AbstractBackend
     downloader::Union{Nothing,Downloads.Downloader}
+    create_new_downloader::Any
 end
 
-DownloadsBackend() = DownloadsBackend(nothing)
+DownloadsBackend() = DownloadsBackend(nothing, () -> get_downloader(; fresh=true))
+DownloadsBackend(D::Downloader) = DownloadsBackend(D, () -> get_downloader(; fresh=true))
 
 const AWS_DOWNLOADER = Ref{Union{Nothing,Downloader}}(nothing)
```
omus, line 27: Not to derail this PR but using undef when the downloader is not yet defined seems preferable
```suggestion
const AWS_DOWNLOADER = Ref{Downloader}()
```
You'd just need to change some code in `get_downloader` to use `isassigned(AWS_DOWNLOADER)` before dereferencing

src/utilities/downloads_backend.jl
```diff
@@ -31,13 +32,14 @@ const AWS_DOWNLOAD_LOCK = ReentrantLock()
 # because we add a hook to avoid redirects in order to try to match the HTTPBackend's
 # implementation, and we don't want to mutate the global downloader from Downloads.jl.
 # https://github.com/JuliaLang/Downloads.jl/blob/84e948c02b8a0625552a764bf90f7d2ee97c949c/src/Downloads.jl#L293-L301
-function get_downloader(downloader=nothing)
+function get_downloader(; fresh=false)
+    downloader = nothing
     lock(AWS_DOWNLOAD_LOCK) do
         yield() # let other downloads finish
         downloader isa Downloader && return nothing
         while true
```
omus, line 40: I'm possibly missing something but I believe the only way we break out of this loop is with `return nothing`. If that is true the later `return downloader` will never be reached

src/utilities/downloads_backend.jl
```diff
@@ -6,22 +6,23 @@ using HTTP.MessageRequest: body_was_streamed
...
 1.7.3 and 1.8 releases ([Downloads.jl#182](https://github.com/JuliaLang/Downloads.jl/issues/182])).
 """
 struct DownloadsBackend <: AWS.AbstractBackend
     downloader::Union{Nothing,Downloads.Downloader}
+    create_new_downloader::Any
 end
 
-DownloadsBackend() = DownloadsBackend(nothing)
+DownloadsBackend() = DownloadsBackend(nothing, () -> get_downloader(; fresh=true))
+DownloadsBackend(D::Downloader) = DownloadsBackend(D, () -> get_downloader(; fresh=true))
```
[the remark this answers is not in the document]
omus: I would suggest you have the following for backwards compatibility:
```julia
DownloadsBackend(D::Downloader) = DownloadsBackend(() -> D)
```
Using this doesn't work with your transient fixes but is effectively just uses the old behaviour. If you want the fix you need to pass in a function. I can't see another option for this as we can't `copy` instances of `Downloader`s.

> But... it's even more complicated, because if we make a new one because we think the old one might have a problem, we don't want to use the old one anymore. But `DownloadsBackend` is immutable

Why not make `DownloadsBackend` mutable or use a `Ref` for the `downloader` field?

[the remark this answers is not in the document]
ericphanson: I am worried about concurrent access to the field: we can have multiple readers and at least one writer to the field. I suppose we can add a lock though.

src/utilities/downloads_backend.jl
ericphanson: BTW: we can likely simplify this once we are on 1.7.3+ or 1.8, following https://github.com/JuliaLang/Downloads.jl/pull/136

src/utilities/downloads_backend.jl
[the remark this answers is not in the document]
ericphanson: I've added a lock and updated the constructors

src/utilities/downloads_backend.jl
```diff
@@ -6,22 +6,23 @@ using HTTP.MessageRequest: body_was_streamed
...
+- `create_new_downloader::Any`: a zero-argument function which returns a new Downloader object to use.
+  Defaults to creating a new global downloader. This is called when a transient error occurs.
 
 Downloads.jl tends to perform better under concurrent operation than HTTP.jl,
 particularly with `@async` / `asyncmap`. As of March 2022, threading (e.g. `@spawn` or `@threads`) with Downloads.jl is broken on all releases of Julia ([Downloads.jl#110](https://github.com/JuliaLang/Downloads.jl/issues/110)), and there are still reported issues on the upcoming
 1.7.3 and 1.8 releases ([Downloads.jl#182](https://github.com/JuliaLang/Downloads.jl/issues/182])).
 """
 struct DownloadsBackend <: AWS.AbstractBackend
     downloader::Union{Nothing,Downloads.Downloader}
+    create_new_downloader::Any
```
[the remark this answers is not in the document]
ericphanson: That isn't documented; is it really better than `Any`?

src/utilities/downloads_backend.jl
ericphanson: Let's do that separately if we are going to change it. I vaguely remember some problems with that approach when trying it in the original implementation. Maybe it was useful to be able to reset back to nothing, or something like that...

src/utilities/downloads_backend.jl
omus: > We are inside an anonymous function, so we are only returning out of the lock

That's what I was overlooking

Issue #537, opened 2022-03-21
JSON parser error when putting CloudWatch log events


Octogonapus:
I got this exception when submitting log data using CloudWatchLogs.jl, though it looks more like an AWS.jl issue given that some JSON parsing failed. Any idea how this exception can happen? Sorry, but I can't create an MWE for this one. If you don't have any ideas, feel free to just close this.

```
┌ Error: tid:1 Wiped logger event buffer due to exception.
│   exception =
│    Expected end of input
│    Line: 0
│    Around: ...1c4b0cd"}{"nextSequenceToken":...
│                        ^
│    
│    Stacktrace:
│      [1] error(s::String)
│        @ Base ./error.jl:33
│      [2] _error(message::String, ps::JSON.Parser.MemoryParserState)
│        @ JSON.Parser ~/.julia/packages/JSON/NeJ9k/src/Parser.jl:140
│      [3] parse(str::String; dicttype::Type, inttype::Type{Int64}, allownan::Bool, null::Nothing)
│        @ JSON.Parser ~/.julia/packages/JSON/NeJ9k/src/Parser.jl:453
│      [4] legacy_response(request::AWS.Request, response::AWS.Response{IOBuffer}; return_headers::Nothing)
│        @ AWS ~/.julia/packages/AWS/5goc1/src/deprecated.jl:160
│      [5] submit_request(aws::AWS.AWSConfig, request::AWS.Request; return_headers::Nothing)
│        @ AWS ~/.julia/packages/AWS/5goc1/src/utilities/request.jl:140
│      [6] (::AWS.JSONService)(operation::String, args::Dict{String, Any}; aws_config::AWS.AWSConfig, feature_set::AWS.FeatureSet)
│        @ AWS ~/.julia/packages/AWS/5goc1/src/AWS.jl:376
│      [7] put_log_events(logEvents::Vector{CloudWatchLogs.LogEvent}, logGroupName::String, logStreamName::String, params::Dict{String, String}; aws_config::AWS.AWSConfig)
│        @ CloudWatchLogs.CloudWatch_Logs ~/.julia/packages/AWS/5goc1/src/services/cloudwatch_logs.jl:1415
│      [8] _put_log_events(stream::CloudWatchLogs.CloudWatchLogStream, events::Vector{CloudWatchLogs.LogEvent})
│        @ CloudWatchLogs ~/.julia/packages/CloudWatchLogs/gXZUD/src/stream.jl:199
```

[...]

PR #554, opened 2022-05-30
Handle 200 responses w/ throttles in body

mattBrzezinski:
## Problem
Taken from the issue, but CloudWatchLogs will at times return a response code of `200` with a `ThrottlingException` inside of the response body. Should be retrying these, however since it's a `200` we're assuming everything is fine.

## Solution
Instead what we need to do is read the response string, check and see if the body contains any of the `THROTTLING_ERROR_CODES`, and throw an AWSException with a `429` response code which we will automatically retry.

This does not immediately solve the issue outlined in #537 as custom handling needs to be done for CloudWatchLogs, but it does enable a fix for [CloudWatchLogs.jl](https://github.com/invenia/CloudWatchLogs.jl) to be made (coming soon!)


src/AWSExceptions.jl
```diff
@@ -77,9 +78,11 @@ function AWSException(e::HTTP.StatusError, body::AbstractString)
 
             # Extract API error code from JSON error message...
             if occursin(r"^application/x-amz-json-1\.[01]$", content_type)
-                info = JSON.parse(body)
-                if haskey(info, "__type")
-                    code = rsplit(info["__type"], '#'; limit=2)[end]
+                # We need to use JSON3, as JSON cannot parse multiple objects in the same string
```
iamed2, line 81: Did you get some indication that this is normal behaviour of the CloudWatch Logs API?

AFAICT the concatenation of JSON in the response is not a valid response, and I can't find any examples online for this. 

If we can't find anything that states this is normal behaviour, or any documented case of this happening outside Julia, I think it could be a bug in HTTP.jl or elsewhere in Julia. Mixing two responses could explain both the `200` and the double-JSON.

mattBrzezinski: > Did you get some indication that this is normal behaviour of the CloudWatch Logs API?

No. However AWS is inconsistent.

> If we can't find anything that states this is normal behaviour, or any documented case of this happening outside Julia, I think it could be a bug in HTTP.jl or elsewhere in Julia. Mixing two responses could explain both the 200 and the double-JSON.

That is a possibility, and I'm going to leave #537 open, I do not have time to look into this further however.

oxinabox:
Doesn't this mean if i have a perfectly reasonable object like a log message (or S3 file?) that happens to contain  the string "Throttling" that it will now error?

mattBrzezinski:
> Doesn't this mean if i have a perfectly reasonable object like a log message (or S3 file?) that happens to contain the string "Throttling" that it will now error?

Good catch! Yep...

src/utilities/request.jl
```diff
@@ -125,6 +125,19 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
...
+        # CloudWatchLogs returns a 200 response code when they throttle us, we need to check the body for a ThrottlingException being returned
+        # Check if a THROTTLING_ERROR_CODE is present in the body, if so return an AWSException w/ a 429 code so we can retry the request
+        body = if request.response_stream !== nothing
+            aws_response.response.body
+        else
+            aws_response.body
+        end
+        body = String(copy(body))
+        if !iszero(occursin.(THROTTLING_ERROR_CODES, body))
+            e = HTTP.StatusError(429, response)
```
mattBrzezinski, line 137: @oxinabox I guess we could do a second check here that the `request` was targetting `CloudWatchLogs`? That would not break for S3.

src/AWSExceptions.jl
ericphanson: > If we can't find anything that states this is normal behaviour, or any documented case of this happening outside Julia, I think it could be a bug in HTTP.jl or elsewhere in Julia. Mixing two responses could explain both the 200 and the double-JSON.

You could try using the Downloads backend; if you can repro then at least you know it's not an HTTP.jl bug

mattBrzezinski: Yeah I tried the Downloads backend, same thing :(

ericphanson: I did some boto sleuthing and found https://github.com/boto/botocore/blob/36eaad651b8cedea30ca3ff32890bdda3d0b2193/botocore/handlers.py#L128-L174 which looks relevant although they don't mention cloudwatch. Can we do the same thing and look for an error tag? I think that would distinguish it from a usual response

src/utilities/request.jl
oxinabox: feels nasty and unsafe still

mattBrzezinski: Yeah thinking about it more, it *is* a solution, but gross. We would also need additional checks for the endpoint we're targeting as I assume reading logs from CW would break this as well.... ugh....

src/AWSExceptions.jl
iamed2: > Okay, actually this seems to be done on purpose! I've ran into a 200 response w/ multiple error messages such as

Going back, just want to confirm this was with Julia?

mattBrzezinski: > > Okay, actually this seems to be done on purpose! I've ran into a 200 response w/ multiple error messages such as
> 
> Going back, just want to confirm this was with Julia?

Correct.

PR #557, opened 2022-06-15
Consume buffer when creating AWSException

quinnj:
At least partially fixes #537. While investigating the depths of how AWS.jl
and HTTP.jl interact, and working on several PRs to improve HTTP.jl, I
noticed a pretty simple way to fix #537 in the most common case where
the user doesn't pass in their own response_stream. Because the default
response_stream is an `IOBuffer`, we can "consume" it's contents when
an AWSException is thrown and thus when the request is retried, the buffer
will be "clean" in that the exception response body has been extracted.

PR #559, opened 2022-06-16
Update minor version to 1.75.0

iamed2:
MR to update the version in case @ericphanson @omus don't think the retry changes are ready to be tagged as-is. Otherwise I will backport https://github.com/JuliaCloud/AWS.jl/pull/557 to 1.74


mattBrzezinski (approved):

ericphanson:
I think the retry changes should be OK; all tests passed and it was meant to be a 1-1 translation (except tweaking the timeouts to be more in line with upstream).

PR #560, opened 2022-06-20
Support HTTP.jl 1.0

quinnj:


src/utilities/downloads_backend.jl
```diff
@@ -126,7 +124,7 @@ end
 
 function _http_response(req::Request, res::Downloads.Response; throw::Bool=true)
     response = HTTP.Response(
-        res.status, res.headers; body=body_was_streamed, request=nothing
+        res.status, res.headers; body=UInt8[], request=nothing
     )
```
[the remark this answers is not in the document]
quinnj: applied and pushed the suggestion manually as the github UI couldn't apply it

quinnj:
Oh, I just barely tagged GitHub.jl release w/ HTTP.jl 1.0, so we may need to wait 10-15 minutes until that release is public for things to work here.

src/utilities/request.jl
```diff
@@ -240,7 +235,6 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
         # https://github.com/JuliaWeb/HTTP.jl/issues/382
         return isa(e, Sockets.DNSError) ||
                isa(e, HTTP.ParseError) ||
-               isa(e, HTTP.IOError) ||
```
iamed2, line 243: Shouldn't this stay in for HTTP 0.9?

src/utilities/request.jl
```diff
@@ -219,12 +214,12 @@ function _http_request(http_backend::HTTPBackend, request::Request, response_str
         buffer = Base.BufferStream()
 
         response = @mock HTTP.request(
-            http_stack,
             request.request_method,
             HTTP.URI(request.url),
             HTTP.mkheaders(request.headers),
             request.content;
-            require_ssl_verification=false,
```
iamed2, line 227: Does this change behaviour?

src/utilities/request.jl
```diff
@@ -200,11 +200,6 @@ end
 function _http_request(http_backend::HTTPBackend, request::Request, response_stream::IO)
     http_options = merge(http_backend.http_options, request.http_options)
 
-    # HTTP options such as `status_exception` need to be used when creating the stack
-    http_stack = HTTP.stack(;
-        redirect=false, retry=false, aws_authorization=false, http_options...
```
iamed2, line 205: Do we not need to keep this for HTTP 0.9? Especially for the `aws_authorization=false, http_options...`?

src/utilities/request.jl
quinnj: It ensures `https` requests verify the tls handshake cert chain, so should make things strictly more secure

src/utilities/request.jl
quinnj: We don't need the `stack` call, but it would be could to keep including the `aws_authorization=false` in our call to `request` later

quinnj:
Ok, looks like I'm going to need to go thru the tests so me more and fix things up.

quinnj:
bors try

quinnj:
bors try

quinnj:
bors try

quinnj:
bors try

quinnj:
Slowly making progress here, but a quick update:
  * Because of how intricately AWS.jl is integrated w/ HTTP.jl I don't see any reasonable way to support both 0.9 and 1.0. AWS.jl just has too much baked in that involved some of the breaking changes, IMO. I've updated the compat to only `HTTP = "1"`. This also involves bumping the Julia compat up to `julia = "1.6"`, since that's the minimum HTTP.jl 1.0 supports. Hopefully not too controversial as 1.6 has been the LTS for ~7 months now?
  * Found a bug we need to fix upstream in HTTP.jl, so waiting on that PR [here](https://github.com/JuliaWeb/HTTP.jl/pull/864)

quinnj:
bors try

quinnj:
bors try

quinnj:
Bump; anyone up for reviewing? Anyone mind if I merge next Monday and tag a new release?

mattBrzezinski:
> Bump; anyone up for reviewing? Anyone mind if I merge next Monday and tag a new release?

I can take a look and play around with it this weekend most likely!

iamed2:
> Woohoo! Tests pass! @iamed2, do you have any concerns on the Julia version bump? Or anyone else?

Maybe this got lost in an Internet blip, but I don't have any concerns there. I think we should check whether downstream packages like CloudWatchLogs.jl and AWSS3.jl continue to work under both backends, then I'm comfortable proceeding. @mattBrzezinski would that be something you could test?

mattBrzezinski:
> 

Sorry I've been busy last couple days with other stuff. Going to look into these changes now, I will test w/ `AWSS3.jl` and `CloudWatchLogs.jl` as well.

mattBrzezinski:
Code changes LGTM! Going to play around with this for a bit, see if anything breaks and will report back.

mattBrzezinski:
Just tested AWSS3.jl with this branch, everything went well. I forgot that CWL.jl requires a stack setup, just made a dummy PR to test this functionality, [here](https://github.com/invenia/CloudWatchLogs.jl/pull/43).

src/deprecated.jl
```diff
@@ -7,7 +7,11 @@ function legacy_response(
 
     # When a user defined I/O stream is passed in use the actual `HTTP.Response` body
     # instead of the `AWS.Response` body which requires the I/O stream to be seekable.
-    body = request.response_stream !== nothing ? response.response.body : response.body
+    body = if request.response_stream !== nothing
+        b"[Message Body was streamed]"
```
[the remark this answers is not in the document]
quinnj: Yeah, it just seemed simpler to leave this as-is for now, since it's deprecated.

mattBrzezinski (approved):
These changes LGTM, not sure if @iamed2 has any other thoughts?

test/runtests.jl
```diff
@@ -38,6 +38,11 @@ using StableRNGs
 
 Mocking.activate()
 
+# backwards compat
+function HTTP.StatusError(status, resp)
+    return HTTP.StatusError(status, resp.request.method, resp.request.target, resp)
+end
+
```
iamed2, line 45: This is bad piracy; this actually gets called by `submit_request` as far as I can tell.

quinnj: Ah yes, forgot to clean this up.

mattBrzezinski:
bors try

iamed2:
Alright I'll tag https://github.com/JuliaCloud/AWS.jl/pull/562 as a patch and then merge and tag this as a minor

iamed2 (approved):

src/utilities/request.jl
```diff
@@ -129,7 +133,7 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
             if HTTP.header(response, "Location") != ""
                 request.url = HTTP.header(response, "Location")
             else
-                e = HTTP.StatusError(response.status, response)
+                e = statuserror(response.status, response)
```
ericphanson, line 136: I think this broke some things for the Downloads backend, because it assumes the response is coming from HTTP.jl and has `method` and `target` fields, whereas before it just wrapped the response, which could be coming from Downloads.jl and may not have those fields. I guess the AWS tests aren't enough to check this case, but I saw when adding Downloads.jl tests to AWSS3: https://github.com/JuliaCloud/AWSS3.jl/pull/262.

PR #563, opened 2022-06-29
Fix error type for HTTP 1.0

### LINKED ISSUE #558: Uncaught error when not running on EC2
I am seeing a sporadic error in `ec2_instance_metadata` when not running on EC2. Sometimes it correctly returns `nothing` and sometimes it throws an IOError instead:

```
julia> AWS.ec2_instance_metadata("")

julia> AWS.ec2_instance_metadata("")

julia> AWS.ec2_instance_metadata("")
ERROR: IOError(Base.IOError("connect: host is down (EHOSTDOWN)", -64) during request(http://169.254.169.254))
...
```

It seems to happen when I run the calls in quick succession, the first few succeed and then an error is thrown.

It seems like the [error being caught](https://github.com/JuliaCloud/AWS.jl/blob/d52c6391dd676e659d3b58f90ab81292fedc7db0/src/AWSCredentials.jl#L219) might also need to include IOError to cover this case, but I'm not sure of the implications so didn't want to assume that's the case.

FWIW, I am running on my local laptop:

```
julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
```

JackDunnNZ:
Fixes #558, ref https://github.com/JuliaCloud/AWS.jl/issues/558#issuecomment-1157957784

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -216,7 +216,7 @@ function ec2_instance_metadata(path::AbstractString)
     request = try
         @mock HTTP.request("GET", uri; connect_timeout=1)
     catch e
-        if e isa HTTP.ConnectionPool.ConnectTimeout
+        if e isa HTTP.ConnectError
             nothing
         else
             rethrow()
--- a/test/patch.jl
+++ b/test/patch.jl
@@ -129,7 +129,10 @@ end
 _instance_metadata_timeout_patch = @patch function HTTP.request(
     method::String, url; kwargs...
 )
-    return throw(HTTP.ConnectionPool.ConnectTimeout("169.254.169.254", "80"))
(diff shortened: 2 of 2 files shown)


quinnj (approved):
Thanks @JackDunnNZ!

quinnj:
bors try

mattBrzezinski:
@quinnj gave you perms so you should be able to merge now

PR #578, opened 2022-09-01
RFC: Support S3 Acceleration, and move to Virtualhost Requests

mattBrzezinski:
## Problem

AWS apparently has plans to [deprecate path-styled requests](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) in the future, these style is currently what we are doing and look like:

> https://s3.us-east-1.amazonaws.com/mattbr-open-world/foobar

The new style of requests are virtualhost, which look like:

> https://mattbr-open-world.s3.amazonaws.com/foobar

We also would like to make use of [S3 Transfer Acceleration](https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration.html#transfer-acceleration-getting-started). Which uses virtualhost however instead of just `s3` it's `s3-accelerated`.

The only caveat for virtualhost which I've come across from the documentation is that the bucket cannot have a `.` character in its name (which is a valid character currently).

## Solution

This PR has a proposed solution for it with our existing code. It does get a bit messy as there are a few use cases we want to handle here. Specifically for S3 Acceleration, this option is disabled by default and is on a per-bucket and request basis. I've added a `use_s3_acceleration = Ref{Bool}(false)` which users can set much like the default `AWSConfig` to allow all requests to use acceleration. If you only wish to accelerate on certain requests the argument `s3_acceleration` can be passed along.

The other quirky part when making requests we cannot pass along an HTTP `resource`, we need to directly specify this in the `URL` of the requests for what we're targeting. This gets a bit messy as a request such as:

```julia
S3.get_object("mattbr-open-world, "foobar")
```

Will go to the [underlying code](https://github.com/JuliaCloud/AWS.jl/blob/de0d4c30b7e735a0303af6e4aa92b60c4962dad9/src/services/s3.jl#L2626) and pass the resource along as `/mattbr-open-world/foobar`. We need the ability to split out the `bucket` from the `key` as well as maintain the `query` arguments as well. This PR handles these cases, and I've commented what it's doing, but it's not that nice.

Thoughts?

### DIFF
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -227,7 +227,7 @@ function generate_service_url(
 
     # S3 virtualhosts cannot have periods in their bucket names, default to path style in these cases
     if service == "s3" && !contains(resource, ".")
-        return s3_virtualhost(service, resource, region; s3_acceleration=s3_acceleration)
+        return s3_virtualhost(service, resource, reg; s3_acceleration=s3_acceleration)
     end
 
     regionless_services = ("iam", "route53")
--- a/src/AWS.jl
+++ b/src/AWS.jl
@@ -44,6 +44,8 @@ using ..AWSExceptions: AWSException
 
 const user_agent = Ref("AWS.jl/1.0.0")
 const aws_config = Ref{AbstractAWSConfig}()
+const use_s3_acceleration = Ref{Bool}(false)
+const SERVICE_HOST = "amazonaws.com"
 
 """
     FeatureSet
@@ -80,6 +82,10 @@ function global_aws_config(; kwargs...)
     end
 
     return aws_config[]
+end
+
+function global_s3_acceleration(b::Bool)
+    return use_s3_acceleration[] = b
 end
 
 """
@@ -211,9 +217,13 @@ end
 # Needs to be included after the definition of struct otherwise it cannot find them
 include("AWSServices.jl")
 
-function generate_service_url(aws::AbstractAWSConfig, service::String, resource::String)
-    SERVICE_HOST = "amazonaws.com"
+function generate_service_url(aws::AbstractAWSConfig, service::String, resource::String; s3_acceleration::Bool=use_s3_acceleration[])
     reg = region(aws)
+
+    # S3 virtualhosts cannot have periods in their bucket names, default to path style in these cases
+    if service == "s3" && !contains(resource, ".")
+        return s3_virtualhost(service, resource, region; s3_acceleration=s3_acceleration)
+    end
 
     regionless_services = ("iam", "route53")
 
@@ -224,6 +234,30 @@ function generate_service_url(aws::AbstractAWSConfig, service::String, resource:
     return string(
         "https://", service, ".", isempty(reg) ? "" : "$reg.", SERVICE_HOST, resource
     )
+end
+
+function s3_virtualhost(service::AbstractString, resource::AbstractString, region::AbstractString; s3_acceleration=nothing)
+    if something(s3_acceleration, use_s3_acceleration[])
+        service = "s3-accelerated"
+    end
+
+    # Split on '?' first to see if there was a query string in our resource
+    split_resource = split(resource, '?')
+
+    # First element contains the bucket, and potential key; "/bucket-name/potential/path/to/key"
+    bucket_key = split(split_resource[1], '/')
+    bucket = bucket_key[2]
+
+    # Rejoin the path on '/'
+    key = length(bucket_key) > 2 ? join(bucket_key[3:end], '/') : ""
+
+    # Check if there was a query in our resource, otherwise default to an empty string
+    query = length(split_resource) > 1 ? string('?', split_resource[2]) : ""
+
+    # Combine key and query on '/', use either, or none
+    key_query = join(filter(!isempty, [key, query]), '/')
+
+    return "https://$(bucket).$(service).$(region).$(SERVICE_HOST)/$(key_query)"
 end
 
 """
@@ -254,8 +288,8 @@ function (service::RestXMLService)(
     feature_set::FeatureSet=FeatureSet(),
 )
     feature_set.use_response_type && _delete_legacy_response_kw_args!(args)
-
     return_headers = _pop!(args, "return_headers", nothing)
+    s3_acceleration = _pop!(args, "s3_acceleration", use_s3_acceleration[])
 
     request = Request(;
         _extract_common_kw_args(service, args)...,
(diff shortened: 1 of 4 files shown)


src/AWS.jl
```diff
@@ -210,8 +216,12 @@ end
 # Needs to be included after the definition of struct otherwise it cannot find them
 include("AWSServices.jl")
 
-function generate_service_url(aws::AbstractAWSConfig, service::String, resource::String)
-    SERVICE_HOST = "amazonaws.com"
+function generate_service_url(aws::AbstractAWSConfig, service::String, resource::String; s3_acceleration::Bool=use_s3_acceleration[])
+    # S3 virtualhosts cannot have periods in their bucket names, default to path style in these cases
+    if service == "s3" && !contains(resource, ".")
```
quinnj, line 221: What if `resource` contains a `.` and `s3_acceleration == true`? Should we throw an error? Or rely on the fact that amazon doesn't allow that already?

src/AWS.jl
```diff
@@ -225,6 +235,30 @@ function generate_service_url(aws::AbstractAWSConfig, service::String, resource:
     )
 end
 
+function s3_virtualhost(service::String, resource::String; s3_acceleration=use_s3_acceleration[])
+    if use_s3_acceleration[] || s3_acceleration
+        service = "s3-accelerated"
+    end
+
+    # Split on '?' first to see if there was a query string in our resource
+    split_resource = split(resource, '?')
```
quinnj, line 244: So if I understand correctly, this code (and just below) is for constructing the virtual-style host url? But it's only available for s3 calls, right? Do we want to eventually support this for other services as well?

src/AWS.jl
mattBrzezinski: This comment just caught an issue here (I'll be adding tests if we agree on this solution). The `resource` passed in here would be something like,

> mattbr-open-world/path/to/some/file.jpg

In this case we want to parse out the bucket name, `mattbr-open-world`, and check that this does not contain a period. From the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html) virtualhost will not work if the bucket contains a period, however it is a valid character for bucket names.

In the case where it does contain a period, default back to the path style requests instead.

iamed2: > From the [documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html) virtualhost will not work if the bucket contains a period, however it is a valid character for bucket names.

I think it does work with HTTP (but I don't think we support non-HTTPS connections) and if you don't perform the standard certificate validation. Better info here I think: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

I believe this is why SDKs support forcing `virtual` mode.

src/AWS.jl
```diff
@@ -225,6 +235,30 @@ function generate_service_url(aws::AbstractAWSConfig, service::String, resource:
     )
 end
 
+function s3_virtualhost(service::AbstractString, resource::AbstractString, region::AbstractString; s3_acceleration=use_s3_acceleration[])
+    if use_s3_acceleration[] || s3_acceleration
```
iamed2, line 239: This doesn't make sense with the way you've set up the function's arguments right?

mattBrzezinski: What do you mean?

We're checking to see if the user has set the global default value for `use_s3_acceleration[]`, or if they're passing `s3_acceleration` specifically for this request.

I guess the case we're not covering here is, if the user set the global default to `true`, then for this specific request did not want to use acceleration.

iamed2: > I guess the case we're not covering here is, if the user set the global default to true, then for this specific request did not want to use acceleration.

Yeah exactly. 

I think typically I would default to `nothing` in the signatures and then do:

```julia
s3_acceleration = s3_acceleration !== nothing ? s3_acceleration : use_s3_acceleration[]
if s3_acceleration
    service = "s3-accelerated"
end
```

This is typically how boto3 does things, where function arguments default to `None` instead of global settings, then fallback to global settings in code. This allows passing through chains of functions the notion of "I would like you to do the default please" rather than having multiple points where the default can creep in, where you can't tell whether the user intentionally passed it or not.

iamed2: EDIT:

```julia
if something(s3_acceleration, use_s3_acceleration[])
    service = "s3-accelerated"
end
```

mattBrzezinski:
bors try

mattBrzezinski:
bors try

mattBrzezinski:
bors try

iamed2:
@mattBrzezinski looks like logs got deleted; do you remember why this was failing tests?

iamed2:
bors try

mattBrzezinski:
> @mattBrzezinski looks like logs got deleted; do you remember why this was failing tests?

I do not, looks like from the last run something isn't being mocked properly? Tbh it's been a while since I've taken a look at this PR. I did mark it as RFC, and without looking through the code it was probably very loosely created just to get an idea out.

iamed2:
> > @mattBrzezinski looks like logs got deleted; do you remember why this was failing tests?
> 
> I do not, looks like from the last run something isn't being mocked properly? Tbh it's been a while since I've taken a look at this PR. I did mark it as RFC, and without looking through the code it was probably very loosely created just to get an idea out.

No worries, I was just trying to figure out where someone would start if picking it up from here.

mattBrzezinski:
> > > @mattBrzezinski looks like logs got deleted; do you remember why this was failing tests?
> > 
> > 
> > I do not, looks like from the last run something isn't being mocked properly? Tbh it's been a while since I've taken a look at this PR. I did mark it as RFC, and without looking through the code it was probably very loosely created just to get an idea out.
> 
> No worries, I was just trying to figure out where someone would start if picking it up from here.

I marked this as an RFC to see if anyone had a better solution for setting acceleration. Just went over it again now, and I believe that this is still good to go, just needs a review.

src/AWS.jl
```diff
@@ -211,19 +217,55 @@ end
...
         reg = ""
     end
 
-    return string(
-        "https://", service, ".", isempty(reg) ? "" : "$reg.", SERVICE_HOST, resource
-    )
+    return _generate_path_url(service, reg, resource)
+end
+
+function s3_paths(
```
iamed2, line 238: Is the result of this function ever used?

mattBrzezinski: Oh I suppose L228 should read,

> && return s3_paths(...

mattBrzezinski:
@iamed2 I don't think that I will realistically have time to take this across the finish line, but it does seem important to Invenia. If you want to take over the PR I'd be happy to review anything needed to get it shipped.

iamed2:
> @iamed2 I don't think that I will realistically have time to take this across the finish line, but it does seem important to Invenia. If you want to take over the PR I'd be happy to review anything needed to get it shipped.

Sounds good, thank you!

mattBrzezinski:
Closing this as it's just lingering and no one is going to look into it.

iamed2:
@mattBrzezinski Yeah just keep the branch around in case it becomes necessary for someone

PR #585, opened 2022-10-05
attempt to increase compat bounds for Compat

ExpandingMan:
I have not attempted to test this but it is starting to wreak havoc for me.  I figured this was better than just opening an issue.

Btw something is wrong with tagbot for Compat.jl, the github tags are out of sync with the registry.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -22,7 +22,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
 [compat]
-Compat = "3.29"
+Compat = "4.2"
 GitHub = "5"
 HTTP = "1"
 IniFile = "0.5"


ericphanson:
bors try

mattBrzezinski (approved):
LGTM

mattBrzezinski:
> Btw something is wrong with tagbot for Compat.jl, the github tags are out of sync with the registry.

Do you mean this? I've noticed it for the last week or so it's been weird.

Screen Shot 2022-10-05 at 4 25 58 PM

Project.toml
```diff
@@ -22,7 +22,7 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 XMLDict = "228000da-037f-5747-90a9-8195ccbf91a5"
 
 [compat]
-Compat = "3.29"
+Compat = "4.2"
```
ericphanson, line 25: Would be nice to not remove compat:
```suggestion
Compat = "3.29, 4.2"
```
(But I think for testing if it works, the tighter bound is OK so we know it ran with it).

ExpandingMan: Also I probably should have put `4` rather than `4.2` since if it works with `3.29` it's extremely likely to work with `4.0`.  Have changed this so it's more generous.

ExpandingMan:
I also bumped the AWS.jl version.

PR #594, opened 2022-12-13
Remove mention of Downloads.jl having threading issues

arnaudh:
The issues mentioned in the comment are closed, and seem fixed based on a cursory look at the comments there:
- https://github.com/JuliaLang/Downloads.jl/issues/110 (closed in Nov 2021)
- https://github.com/JuliaLang/Downloads.jl/issues/182 (closed in Apr 2022)

Is it fair to say Downloads.jl doesn't have threading issues anymore? I have zero knowledge around this, just trying to understand how safe it is to use this backend for our use case (at Invenia) where threading might be used.

### DIFF
--- a/src/utilities/downloads_backend.jl
+++ b/src/utilities/downloads_backend.jl
@@ -12,7 +12,7 @@ which is the `Downloads.Downloader` to use. If set to `nothing`, the default,
 then a global downloader object will be used.
 
 Downloads.jl tends to perform better under concurrent operation than HTTP.jl,
-particularly with `@async` / `asyncmap`.
+particularly with `@async` / `asyncmap`. Note that threading (e.g. `@spawn` or `@threads`) with Downloads.jl is broken on Julia releases prior to 1.8 (https://github.com/JuliaLang/Downloads.jl/issues/182#issuecomment-1069269944).
 """
 struct DownloadsBackend <: AWS.AbstractBackend
     downloader::Union{Nothing,Downloads.Downloader}


mattBrzezinski:
@ericphanson commented this 9 months ago in [this](https://github.com/JuliaCloud/AWS.jl/commit/82592afbdf23eec9c2008baa59219afd455ed43d) commit. 

I think we should just need to test and confirm the usages here to make sure it works as intended?

src/utilities/downloads_backend.jl
```diff
@@ -12,8 +12,7 @@ which is the `Downloads.Downloader` to use. If set to `nothing`, the default,
 then a global downloader object will be used.
 
 Downloads.jl tends to perform better under concurrent operation than HTTP.jl,
-particularly with `@async` / `asyncmap`. As of March 2022, threading (e.g. `@spawn` or `@threads`) with Downloads.jl is broken on all releases of Julia ([Downloads.jl#110](https://github.com/JuliaLang/Downloads.jl/issues/110)), and there are still reported issues on the upcoming
-1.7.3 and 1.8 releases ([Downloads.jl#182](https://github.com/JuliaLang/Downloads.jl/issues/182])).
+particularly with `@async` / `asyncmap`.
```
arnaudh, line 15:
```suggestion
particularly with `@async` / `asyncmap`. Note that threading (e.g. `@spawn` or `@threads`) with Downloads.jl is broken on Julia releases prior to 1.8 (https://github.com/JuliaLang/Downloads.jl/issues/182#issuecomment-1069269944).
```

Seems like it's not threadsafe on anything prior to 1.8 according to that comment by @StefanKarpinski (though couldn't ascertain 100% from looking over the linked issues, e.g. whether any fix finally made it to 1.7 or even 1.6?).

It would be nice if Downloads.jl mentioned in its README on which Julia versions it is threadsafe (bonus linking to issues).

ericphanson:
I think HTTP.jl v1 may also have fixed a lot of its concurrency issues. I myself haven't tried any threading experiments w/ HTTP.jl v1 or Downloads.jl on Julia 1.8 yet though, and haven't seen any discussion from other folks.

But I think it makes sense to limit the warnings to talk about only HTTP.jl pre-v1.0 and Downloads.jl pre-julia-1.8 since there aren't open issues as far as I know w/ the newest versions.

ericphanson:
BTW I think it would be great to have some kind of open source example application demonstrating high throughput to s3 (or between ec2 and s3) w/ different object sizes and backends which could also be used as a smoketest here to check for concurrency issues/regressions/which versions actually work correctly, as well as to show known-good patterns for doing that through the julia ecosystem.

mattBrzezinski (approved):
bors r+

Issue #551, opened 2022-05-12
Support `AWS_MAX_ATTEMPTS` environmental variable


ericphanson:
Ref https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html

PR #595, opened 2022-12-17
Make the number of retries configurable

iamed2:
This adds a `max_attempts` kwarg+field to AWSConfig, and a `max_attempts` accessor for `AbstractAWSConfig` (defaulting to 3, which is the default elsewhere in the code). 

I saw most of the tests are done with S3, and I also saw that the S3 throttling error (`SlowDown`) was not included. I added that error code and then added a test that uses it and tests throttling retries. 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204220867368963


iamed2:
I think likely there are a few more throttling error codes we could add in, in a future PR: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#standard-retry-mode

iamed2:
bors try

ericphanson:
AWS has a conventional env variable btw: https://github.com/JuliaCloud/AWS.jl/issues/551

src/utilities/request.jl
```diff
@@ -190,7 +191,7 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
         return false
     end
 
-    delays = AWSExponentialBackoff(; max_attempts=3)
+    delays = AWSExponentialBackoff(; max_attempts=max_attempts(aws))
```
arnaudh, line 194: How about making the whole delay strategy configurable?

```suggestion
    delays = delays(aws)
```

to give more control, e.g. tweak the initial delay or delay between retries.

Perhaps this would require exporting `AWSExponentialBackoff` so it's easy for users to tweak only the max attempt.

mattBrzezinski:
bors try

mattBrzezinski (approved):
bors r+

PR #602, opened 2023-02-15
Use URIs.escapepath when cleaning S3 URIs

### LINKED ISSUE #592: Bad signature for Keys containing Unicode
Using raw Unicode characters for `key`s leads to an invalid signature. This seems to be a reliable MWE:

```julia
using AWS: AWS
bucket = "mortenpi-test-bucket"
AWS.@service S3 use_response_type=true
aws_config = AWS.AWSConfig(region="ap-southeast-2")
params = Dict("body" => "123",)
S3.put_object(bucket, "📁", params; aws_config)
```

fails with

```
ERROR: AWS.AWSExceptions.AWSException: SignatureDoesNotMatch -- The request signature we calculated does not match the signature you provided. Check your key and signing method.

HTTP.Exceptions.StatusError(403, "PUT", "/mortenpi-test-bucket/📁", HTTP.Messages.Response:
"""
HTTP/1.1 403 Forbidden
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Mon, 21 Nov 2022 09:25:38 GMT
Server: AmazonS3

[Message Body was streamed]""")

<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message>
...
```

Doing `URIs.escapepath()` on `key` does solve the issue. But I think this should also work as-is, because AWS does do limited escaping already?

https://github.com/JuliaCloud/AWS.jl/blob/4bd8b958f7172124922b32be449502faadddd489/src/utilities/utilities.jl#L46-L63

mortenpi:
Just using `URIs.escapepath` to also escape e.g. Unicode characters seems to be enough to fix #592. That said, I am still not quite sure what the original issue there is.


mortenpi:
Friendly bump. It would be awesome if this could be merged and tagged :slightly_smiling_face:

mortenpi:
I believe I fixed the failing test -- just an oversight when deleting the bucket. I don't think I can trigger bors though, to test it?

mattBrzezinski:
bors try

mattBrzezinski:
> I believe I fixed the failing test -- just an oversight when deleting the bucket. I don't think I can trigger bors though, to test it?

Correct, I would need to add you to the list in bors or the org. Can do shortly!

mattBrzezinski (approved):
bors r+

PR #613, opened 2023-03-31
Fix AWSException construction for JSON-encoded responses with invalid bodies

Octogonapus:
This PR fixes a bug I ran into when receiving a JSON-encoded response with an invalid JSON body.
I received this response from a custom API GW endpoint, not from an official AWS endpoint.

### DIFF
--- a/src/AWSExceptions.jl
+++ b/src/AWSExceptions.jl
@@ -110,7 +110,7 @@ function AWSException(e::HTTP.StatusError, body::AbstractString)
     end
 
     # Sometimes info is a string, in which case there is nothing else to do
-    if (info isa AbstractDict)
+    if info isa AbstractDict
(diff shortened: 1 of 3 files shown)


src/AWSExceptions.jl
```diff
@@ -109,15 +109,18 @@ function AWSException(e::HTTP.StatusError, body::AbstractString)
         end
     end
 
-    # There are times when Errors or Error are returned back
-    info = get(info, "Errors", info)
-    info = get(info, "Error", info)
+    # Sometimes info is a string, in which case there is nothing else to do
+    if (info isa AbstractDict)
```
mattBrzezinski, line 113:
```suggestion
    if info isa AbstractDict
```

mattBrzezinski (approved):
bors r+

PR #620, opened 2023-05-09
Support `credential_process` config setting

omus:
Adds support for the AWS config setting [`credential_process`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html). I want to add support for this feature as I want I'm using [`aws-vault`'s integration](https://github.com/99designs/aws-vault/blob/master/USAGE.md#use-case-2-aws-vault-is-a-master-credentials-vault-for-aws-sdk) which means I can store my AWS credentials in one place.

As part of this I encountered that if you have an `~/.aws/credentials` file that the `credential_process` setting was ignored. This made me do some testing into if the credential preference order mirrors what the AWS CLI does. The results of these tests found the following:

- config file `sso_*` used over `~/.aws/credentials`
- `~/.aws/credentials` used over config file `credential_process`
- config file `credential_process` used over config file `aws_access_key_id`/`aws_secret_access_key`

This was using the AWS CLI version `aws-cli/2.11.13 Python/3.11.3 Darwin/22.4.0 source/arm64 prompt/off`. In this PR the `credential_process` is preferred over config file `aws_access_key_id`/`aws_secret_access_key` but the `sso_*` credentials have the wrong preference ordering. I'll try to address the credential preference ordering issues in another PR.

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -8,22 +8,22 @@ using ..AWSExceptions
 
 export AWSCredentials,
     aws_account_number,
-    aws_get_region,
     aws_get_profile_settings,
+    aws_get_region,
     aws_user_arn,
     check_credentials,
+    credentials_from_webtoken,
     dot_aws_config,
+    dot_aws_config_file,
     dot_aws_credentials,
     dot_aws_credentials_file,
-    dot_aws_config_file,
     ec2_instance_credentials,
     ecs_instance_credentials,
     env_var_credentials,
+    external_process_credentials,
     localhost_is_ec2,
-    localhost_maybe_ec2,
     localhost_is_lambda,
-    credentials_from_webtoken,
-    external_process_credentials
+    localhost_maybe_ec2
 
 function localhost_maybe_ec2()
     return localhost_is_ec2() || isfile("/sys/devices/virtual/dmi/id/product_uuid")
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -424,10 +424,14 @@ function dot_aws_config(profile=nothing)
         settings = _aws_profile_config(ini, p)
         isempty(settings) && return nothing
 
+        credential_process = get(settings, "credential_process", nothing)
         access_key = get(settings, "aws_access_key_id", nothing)
         sso_start_url = get(settings, "sso_start_url", nothing)
 
-        if !isnothing(access_key)
+        if !isnothing(credential_process)
+            cmd = Cmd(Base.shell_split(credential_process))
+            return credential_process_credentials(cmd)
+        elseif !isnothing(access_key)
             access_key, secret_key, token = _aws_get_credential_details(p, ini)
             return AWSCredentials(access_key, secret_key, token)
         elseif !isnothing(sso_start_url)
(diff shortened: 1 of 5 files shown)


omus:
bors try

omus:
Made some error during formatting:
```
{
  "Version": 2
}
Credential Process: Test Failed at /home/runner/work/AWS.jl/AWS.jl/test/AWSCredentials.jl:794
  Expression: external_process_credentials(gen_process(unhandled_version_resp))
    Expected: ErrorException("Credential process returned unhandled version 2:\nnothing")
      Thrown: ErrorException("Credential process returned unhandled version 2:\n{\n  \"Version\": 2\n}\n")
```

omus:
bors try

src/AWSCredentials.jl
```diff
@@ -22,7 +22,8 @@ export AWSCredentials,
     localhost_is_ec2,
     localhost_maybe_ec2,
     localhost_is_lambda,
-    credentials_from_webtoken
+    credentials_from_webtoken,
+    external_process_credentials
```
mattBrzezinski, line 26: [Nit] This list of exports is almost alphabetical

src/AWSCredentials.jl
```diff
@@ -424,10 +425,14 @@ function dot_aws_config(profile=nothing)
         settings = _aws_profile_config(ini, p)
         isempty(settings) && return nothing
 
+        credential_process = get(settings, "credential_process", nothing)
         access_key = get(settings, "aws_access_key_id", nothing)
         sso_start_url = get(settings, "sso_start_url", nothing)
 
-        if !isnothing(access_key)
+        if !isnothing(credential_process)
+            cmd = Cmd(Base.shell_split(credential_process))
```
mattBrzezinski, line 433: Lol.

```
help?> Base.shell_split
  No documentation found.

  Base.shell_split is a Function.

  # 1 method for generic function "shell_split":
  [1] shell_split(s::AbstractString) in Base at shell.jl:137
  ```

File: src/utilities/credentials.jl [hunk 3]
```diff
@@ -212,3 +212,38 @@ function _aws_get_sso_credential_details(profile::AbstractString, ini::Inifile)
...
+        )
+    end
+
+    access_key_id = json["AccessKeyId"]
+    secret_access_key = json["SecretAccessKey"]
+
+    # The presence of the "Expiration" key determines if the provided credentials are
+    # long-term credentials or temporary credentials. Temporary credentials must include a
+    # session token (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)
+    if haskey(json, "Expiration") || haskey(json, "SessionToken")
```
mattBrzezinski, line 240: Can we end up in a situation where we have `SessionToken` but not `Expiration`, and fail our parse?

omus: Definitely, from what I've gathered from the documentation a `SessionToken` is only present if the credentials are temporary. If you pass in an `Expiration` then the credentials are also deemed to be temporary so I would expect if either of these are set then both should be set. If only one is set then we throw a `KeyError` (this is tested)

src/AWSCredentials.jl
omus: I don't love using this but I've used it in the past with LibGit2 which is why I know about it

src/AWSCredentials.jl
omus: I'll order this

src/utilities/credentials.jl
omus: As the protocol isn't well defined here it seems best to me to error until we find an concrete example we want to add support for.

mattBrzezinski (approved):

PR #621, opened 2023-05-10
Update credential precedence to match AWS CLI

omus:
I noticed there were some credential precedence ordering differences between AWS.jl and AWS CLI. I ended up doing some experimentation with pairing different AWS CLI settings to determine the precedence ordering used by AWS CLI. Here are the results of those tests:

- aws `--profile` used over env `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`
- aws `--profile` used over env `AWS_PROFILE`
- env `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` used over env `AWS_PROFILE`
- env `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` used over config file `sso_*`
- config file `sso_*` used over `~/.aws/credentials` (if exists)
- `~/.aws/credentials` (if exists) used over config file `credential_process`
- config file `credential_process` used over config file `aws_access_key_id`/`aws_secret_access_key`
- config file `aws_access_key_id`/`aws_secret_access_key` used over EC2 instance metadata
- config file `aws_access_key_id`/`aws_secret_access_key` used over `AWS_CONTAINER_CREDENTIALS_FULL_URI`

Using `aws-cli/2.11.13 Python/3.11.3 Darwin/22.4.0 source/arm64 prompt/off`

Notes:
- Defining `sso_account_id` or `sso_role_name` in a profile without other `sso_*` keys results in an error about missing required configuration. Defining `sso_start_url` and `sso_region` by themselves doesn't produce this error.
- Specifying the AWS credential file with `AWS_SHARED_CREDENTIALS_FILE` just replaces `~/.aws/credentials`
- Tested this by specifying bad credentials in one source and valid ones in the other. As I didn't have an SSO setup to test against I could only force these to fail.
- Some additional testing was done to verify that the credential preference ordering is linear. I didn't find any examples of non-linear ordering.

### DIFF
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -110,12 +110,13 @@ Checks credential locations in the order:
 function AWSCredentials(; profile=nothing, throw_cred_error=true)
     creds = nothing
     credential_function = () -> nothing
+    explicit_profile = !isnothing(profile)
     profile = @something profile _aws_get_profile()
 
     # Define our search options, expected to be callable with no arguments.
     # Throw NoCredentials if none are found
     functions = [
-        env_var_credentials,
+        () -> env_var_credentials(explicit_profile),
         () -> dot_aws_credentials(profile),
         () -> dot_aws_config(profile),
         credentials_from_webtoken,
@@ -355,12 +356,15 @@ function ecs_instance_credentials()
 end
 
 """
-    env_var_credentials() -> Union{AWSCredential, Nothing}
+    env_var_credentials(explicit_profile::Bool=false) -> Union{AWSCredential, Nothing}
 
 Use AWS environmental variables (e.g. AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, etc.)
 to create AWSCredentials.
 """
-function env_var_credentials()
+function env_var_credentials(explicit_profile::Bool=false)
+    # Skip using environmental variables when a profile has been explicitly set
+    explicit_profile && return nothing
+
     if haskey(ENV, "AWS_ACCESS_KEY_ID") && haskey(ENV, "AWS_SECRET_ACCESS_KEY")
         return AWSCredentials(
             ENV["AWS_ACCESS_KEY_ID"],
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -397,7 +397,9 @@ end
 """
     dot_aws_credentials(profile=nothing) -> Union{AWSCredential, Nothing}
 
-Retrieve AWSCredentials from the `~/.aws/credentials` file
+Retrieve `AWSCredentials` from the AWS CLI credentials file. The credential file defaults to
+"~/.aws/credentials" but can be specified using the env variable
+`AWS_SHARED_CREDENTIALS_FILE`.
 
 # Arguments
 - `profile`: Specific profile used to get AWSCredentials, default is `nothing`
@@ -458,9 +460,11 @@ end
 """
     dot_aws_config(profile=nothing) -> Union{AWSCredential, Nothing}
 
-Retrieve AWSCredentials for the default or specified profile from the `~/.aws/config` file.
-If this fails, try to retrieve credentials from `_aws_get_role()`, otherwise return
-`nothing`.
+Retrieve `AWSCredentials` from the AWS CLI configuration file. The configuration file
+defaults to "~/.aws/config" but can be specified using the env variable  `AWS_CONFIG_FILE`.
+When no credentials can be found for the given `profile` then the associated
+`source_profile` will be used to recursively look up credentials of parent profiles. If
+still no credentials can be found then `nothing` will be returned.
 
 # Arguments
 - `profile`: Specific profile used to get AWSCredentials, default is `nothing`
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -465,7 +465,6 @@ function dot_aws_config(profile=nothing)
 
         credential_process = get(settings, "credential_process", nothing)
         access_key = get(settings, "aws_access_key_id", nothing)
-        sso_start_url = get(settings, "sso_start_url", nothing)
 
         if !isnothing(credential_process)
             cmd = Cmd(Base.shell_split(credential_process))
@@ -473,9 +472,6 @@ function dot_aws_config(profile=nothing)
         elseif !isnothing(access_key)
             access_key, secret_key, token = _aws_get_credential_details(p, ini)
             return AWSCredentials(access_key, secret_key, token)
-        elseif !isnothing(sso_start_url)
-            access_key, secret_key, token, expiry = _aws_get_sso_credential_details(p, ini)
-            return AWSCredentials(access_key, secret_key, token; expiry=expiry)
         else
             return _aws_get_role(p, ini)
         end
--- a/src/AWSCredentials.jl
+++ b/src/AWSCredentials.jl
@@ -117,6 +117,7 @@ function AWSCredentials(; profile=nothing, throw_cred_error=true)
     # Throw NoCredentials if none are found
     functions = [
         () -> env_var_credentials(explicit_profile),
+        () -> sso_credentials(profile),
         () -> dot_aws_credentials(profile),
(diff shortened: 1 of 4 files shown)


src/AWSCredentials.jl
```diff
@@ -427,17 +465,13 @@ function dot_aws_config(profile=nothing)
...
 
         if !isnothing(credential_process)
             cmd = Cmd(Base.shell_split(credential_process))
             return external_process_credentials(cmd)
         elseif !isnothing(access_key)
             access_key, secret_key, token = _aws_get_credential_details(p, ini)
             return AWSCredentials(access_key, secret_key, token)
-        elseif !isnothing(sso_start_url)
-            access_key, secret_key, token, expiry = _aws_get_sso_credential_details(p, ini)
-            return AWSCredentials(access_key, secret_key, token; expiry=expiry)
```
omus, line 440: Potential breakage if someone was using `dot_aws_config` directly to get credentials for SSO support.

test/AWSCredentials.jl
```diff
@@ -529,47 +669,22 @@ end
...
-                            """
-                            [profile $(test_values["Test-Config-Profile"])]
-                            aws_access_key_id = invalid
-                            aws_secret_access_key = invalid
-                            credential_process = $(abspath(credential_process_file))
-                            """,
-                        )
-                    end
-
-                    result = dot_aws_config(test_values["Test-Config-Profile"])
```
omus, line 566: Diff is pretty messy but this was refactored and moved into the "Credential Precedence" testset

omus:
bors try

omus:
bors try

omus:
bors try

src/AWSCredentials.jl
```diff
@@ -110,12 +111,14 @@ Checks credential locations in the order:
...
     profile = @something profile _aws_get_profile()
 
-    # Define our search options, expected to be callable with no arguments.
-    # Throw NoCredentials if none are found
+    # Define the credential preference order
+    # https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html#cli-chap-authentication-precedence
     functions = [
-        env_var_credentials,
+        () -> env_var_credentials(explicit_profile),
+        () -> sso_credentials(profile),
```
mattBrzezinski, line 121: I can't comment on the line cause it's too far down, but to match the documentation we need to swap `ecs_instance_credentials` and `ec2_instance_credentials`.

src/AWSCredentials.jl
```diff
@@ -355,12 +358,15 @@ function ecs_instance_credentials()
...
-    env_var_credentials() -> Union{AWSCredential, Nothing}
+    env_var_credentials(explicit_profile::Bool=false) -> Union{AWSCredential, Nothing}
 
 Use AWS environmental variables (e.g. AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, etc.)
 to create AWSCredentials.
 """
-function env_var_credentials()
+function env_var_credentials(explicit_profile::Bool=false)
+    # Skip using environmental variables when a profile has been explicitly set
+    explicit_profile && return nothing
```
mattBrzezinski, line 368: Users could have an unintended affect here if they passed in a profile and have ENV VARs set. However I think moving towards matching the CLI is the proper step.

src/AWSCredentials.jl
```diff
@@ -404,12 +410,43 @@ function dot_aws_credentials_file()
...
+    sso_credentials(profile=nothing) -> Union{AWSCredential, Nothing}
+
+Retrieve credentials via AWS single sign-on settings defined in the `profile` within the AWS
+config file. If no SSO settings are found for the `profile` `nothing` is returned.
+
+# Arguments
+- `profile`: Specific profile used to get `AWSCredential`s, default is `nothing`
+"""
+function sso_credentials(profile=nothing)
+    config_file = @mock dot_aws_config_file()
```
mattBrzezinski, line 423: I *believe* you can override the default AWS Config file, maybe a comment on this function stating if the user is doing this in their setup they need to set the `AWS_CONFIG_FILE` variable to point to it?

src/AWSCredentials.jl
mattBrzezinski: I *think* it's probably safe to remove, but I *feel* like we should keep it in place. Having the rug pulled under you would be quite annoying.

mattBrzezinski (approved):

src/AWSCredentials.jl
omus: Good catch. We'll also need to some special casing to prefer using ECS credentials when the environmental variables are set. Seems like the only reason things are working is that we're currently only using the ECS credentials when they are set.

src/AWSCredentials.jl
omus: I personal consider this a bugfix as the old behavior did this:
```
❯ eval $(aws-vault export demo1 --format=export-env)

❯ aws sts get-caller-identity --query Arn --output text
arn:aws:sts:::assumed-role/demo1/1683754266745769000

❯ aws sts get-caller-identity --query Arn --output text --profile demo2
arn:aws:sts:::assumed-role/demo2/botocore-session-1683754083

❯ julia -e 'using AWS; @service STS; println(STS.get_caller_identity()["GetCallerIdentityResult"]["Arn"])'
arn:aws:sts:::assumed-role/demo1/1683754434179050000

❯ julia -e 'using AWS; @service STS; global_aws_config(; profile="s3utils-ci"); println(STS.get_caller_identity()["GetCallerIdentityResult"]["Arn"])'
arn:aws:sts:::assumed-role/demo1/1683754434179050000

❯ unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN AWS_CREDENTIAL_EXPIRATION
```
Not in particular if you try to use a different profile AWS.jl would just silently ignore it.

src/AWSCredentials.jl
omus: > I _think_ it's probably safe to remove, but I _feel_ like we should keep it in place. Having the rug pulled under you would be quite annoying.
– https://github.com/JuliaCloud/AWS.jl/pull/621#discussion_r1190330203

omus: I'll add this back in with either a deprecation or just a comment

src/AWSCredentials.jl
omus: You definitely can use `AWS_CONFIG_FILE` to modify where the config file is located. I'm not using the `@mock` here so I can just remove that part. Do you want me to mention this in the docstring? Maybe I'm not following

mattBrzezinski: Yeah I meant in the docstr for the `dot_aws_config()` to state if you're using a non-default location you'll need to set that ENVVAR.

src/AWSCredentials.jl
omus: > The Amazon ECS agent populates the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable
– https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

> If your Amazon EC2 instance is using at least version `1.11.0` of the container agent and a supported version of the AWS CLI or SDKs, then the SDK client will see that the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` variable is available, and it will use the provided credentials to make calls to the AWS APIs.
– https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

Although not definitive it sounds like the these environmental variables are always set when the ECS credential provider should be used. The only counter example to this is the ordering here: https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html#cli-chap-authentication-precedence

Additionally, our implementation only ever used ECS credentials when the environmental variable is set so at worse we have some additional functionality to add later on when we have a concrete example to work off of.

omus:
bors try

src/AWSCredentials.jl
omus: Done

omus:
bors try

omus:
bors try

Issue #612, opened 2023-03-30
Add integration with `aws-vault` for credentials


kleinschmidt:
[`aws-vault`](https://github.com/99designs/aws-vault/) is convenient when using MFA, since it allows MFA-authenticated credentials to be cached for some time period, prompting the user for a new MFA when the credentials expire.  It's not immediately obvious to me whether the environment variable or server modes would be easier to integrate with how AWS.jl handles credentials; it looks like the ENV mode provides equivalent credentials to what AwsCredentials requires via environment variables, but AFAIU does not prompt for a refresh when the credentials expire.

omus:
So using `aws-vault exec $role` directly will assume a role and set ENV variables that AWS.jl can use (as you mentioned). The problem comes in that these session variables will eventually expire resulting in you having to call `aws-vault exec $role` again to refresh the credentials which is extra annoying when working within the Julia REPL.

A better approach would be to use the AWS CLI supported [`credential_process`](https://github.com/99designs/aws-vault/blob/master/USAGE.md#use-case-3-aws-vault-is-a-mfa-session-cache-for-aws-sdk) with `aws-vault export` as shown in this [use case](https://github.com/99designs/aws-vault/blob/master/USAGE.md#use-case-3-aws-vault-is-a-mfa-session-cache-for-aws-sdk).

Support for `credential_process` was added in #620 (released in 1.85.0) which makes this approach feasible. 

My personal recommended way to set this up is to make you AWS CLI config (e.g. `~/.aws/config`) follow this pattern:

```
[profile JSMITH]
mfa_serial = arn:aws:iam::111222333444:mfa/JSMITH
credential_process = sh -c "unset AWS_VAULT; aws-vault export --format=json --prompt=terminal JSMITH"

[profile ROLE]
role_arn = arn:aws:iam::111222333444:role/ROLE
source_profile = JSMITH
include_profile = JSMITH

[profile CHAINED_ROLE]
role_arn = arn:aws:iam::111222333444:role/CHAINED_ROLE
source_profile = ROLE
```

Implementation details of the AWS CLI config file:
- We unset `AWS_VAULT` in [`credential_process`](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html) to allow the profile to be changed even when running within a `aws-vault exec` subshell.
[...]

PR #622, opened 2023-05-11
Support providing a full URI to the ECS credential provider

omus:
Discovered that our ECS credential support was lacking support for the [`AWS_CONTAINER_CREDENTIALS_FULL_URI` and the `AWS_CONTAINER_AUTHORIZATION_TOKEN`](https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html) while I was working on `aws-vault` support (#612). With `aws-vault` you can use `aws-vault exec --ecs-server` to run the tool in server mode which emulates the ECS credential provider and uses these environmental variables.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


mattBrzezinski (approved):

omus:
bors try

PR #627, opened 2023-06-05
Use OIDC to assume CI AWS role

omus:
Using OpenID Connect in our GitHub Actions workflows to authenticate with AWS. For details see:
- https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services
- https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect

The primary benefit of this is that we no longer have to specify AWS credentials as GitHub Action secrets which is more secure as there are no longer any secrets that could be leaked. I've already deployed the update version of the `aws_jl_test.yaml` CloudFormation template and currently this co-exists with the old CloudFormation stack.

Sibling PR: https://github.com/JuliaCloud/AWSS3.jl/pull/284

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,10 +42,6 @@ jobs:
       MINIO_REGION_NAME: aregion
     steps:
       - uses: actions/checkout@v2
-      - name: Show JWT from GH OIDC
-        run: |
-          curl -s -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sts.amazonaws.com" | tee JSON
-          echo "$JSON" | jq -r .value | cut -d. -f2 | base64 -d | jq
       - name: Assume AWS role
         uses: aws-actions/configure-aws-credentials@v2
         with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,6 +41,9 @@ jobs:
       MINIO_REGION_NAME: aregion
     steps:
       - uses: actions/checkout@v2
+      - name: Show JWT from GH OIDC
+        run: |
+          curl -s -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sts.amazonaws.com" | jq -r .value | cut -d. -f2 | base64 -d | jq
       - name: Assume AWS role
         uses: aws-actions/configure-aws-credentials@v2
         with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -6,7 +6,6 @@ on:
       - master
       - staging
       - trying
-      - cv/oidc-ci-role
     tags: '*'
   schedule:
     - cron: '0 2 * * *'  # Daily at 2 a.m. UTC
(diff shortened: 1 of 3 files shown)


omus:
bors try

.github/workflows/CI.yml
```diff
@@ -36,6 +41,11 @@ jobs:
       MINIO_REGION_NAME: aregion
     steps:
       - uses: actions/checkout@v2
+      - name: Assume AWS role
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: arn:aws:iam::182697898961:role/AWS.jl
```
mattBrzezinski, line 47: The AWS Account number might want to be a secret

omus:
bors try

omus:
bors try

.github/workflows/CI.yml
```diff
@@ -36,6 +42,15 @@ jobs:
...
       - uses: actions/checkout@v2
+      - name: Show JWT from GH OIDC
+        run: |
+          curl -s -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=sts.amazonaws.com" | tee JSON
+          echo "$JSON" | jq -r .value | cut -d. -f2 | base64 -d | jq
+      - name: Assume AWS role
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: arn:aws:iam::182697898961:role/AWS.jl
+          aws-region: us-east-1
```
mattBrzezinski, line 53: Not sure if this needs to be specified since IAM is global and not region-specific?

.github/workflows/CI.yml
```diff
@@ -73,9 +88,6 @@ jobs:
           git config --global user.name Tester
           git config --global user.email te@st.er
       - uses: julia-actions/julia-runtest@latest
-        env:
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
   slack:
```
mattBrzezinski, line 91: We can probably axe this block out since the Invenia slack is no longer monitored.

.github/workflows/CI.yml
omus: I believe this specifies the default region to use during this session

omus:
bors try

omus:
bors try

.github/workflows/CI.yml
omus: I agree. I'll save this for another PR though

omus:
bors try

.github/workflows/CI.yml
omus: This actions also requires it:
```
Error: Input required and not supplied: aws-region
```

omus:
bors try

.github/workflows/CI.yml
mattBrzezinski: I would also re-write the commits to remove them since they would still exist in history

mattBrzezinski (approved):

omus:
Some post-merge follow up to do includes removing the old "AWS-jl" CFN stack as we now use the "AWS-jl-test" CFN stack.

omus:
The old "AWS-jl" CFN stack was manually removed.

omus:
Deleted the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` repo secret and the manually created AWS user `PublicCIUser`.

Issue #628, opened 2023-06-05
Credentials via SSO token provider configuration are not supported


omus:
AWS replaced [AWS SSO with AWS IAM Identity Center](https://aws.amazon.com/about-aws/whats-new/2022/09/aws-iam-identity-center-apis-manage-users-groups-scale/). AWS.jl supported AWS SSO credentials via setting a profile with [`sso_start_url`](https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-legacy.html#sso-configure-profile-manual) but doesn't support handling [`sso-session` sections](https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-token.html#sso-configure-profile-token-manual).

iamed2:
I don't believe this was related to the change to IAM Identity Center, I think it happened after that rename but separately.

The old method is still available: https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-legacy.html

PR #629, opened 2023-06-05
Improve exception when IAM Identity Center auth is used

omus:
AWS.jl currently doesn't support authentication via IAM Identity Center (successor to AWS SSO) (https://github.com/JuliaCloud/AWS.jl/issues/628) which results in a `NoCredentials` exception being raised which is misleading. This PR simply throws a more helpful exception when users attempt to use this currently unsupported authentication method.


omus:
bors try

mattBrzezinski (approved):

omus:
bors try

Issue #630, opened 2023-06-07
Publicly hosted bors-ng instance is deprecated


omus:
When bors workflows run successfully you now see a warning about the publicly hosted bors-ng instance is deprecated and will be removed in the future (https://github.com/JuliaCloud/AWS.jl/pull/627#issuecomment-1577415040). Details on the deprecation: https://bors.tech/newsletter/2023/05/01/tmib-76/

The suggestioned options are to:
1. Host our own bors instance
2. Switch over to GitHub's built-in merge queue

Using a GitHub merge queue doesn't solve our particular problem as what we want to do is limit which users can trigger CI workflows.

omus:
Looks like this may have just become urgent. I have PRs in AWS.jl and AWSS3.jl that don't seem to be responding to bors requests

omus:
We can configure Git repos to require approval before running workflows: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories.

omus:
I'm going to forge ahead with this work as it's blocking some other work of mine. We previously used "Require approval for first-time contributors" for this setting. I've now set it to "Require approval for all outside contributors" as we don't want a malicious actor submitting a trivial change and then being able to run whatever on CI:

![Screenshot 2023-06-07 at 13 05 32](https://github.com/JuliaCloud/AWS.jl/assets/1675958/f0d73d49-2240-437d-aaa8-b2d38bc2319a)

For our branch protection rules here's the old settings:


![Screenshot 2023-06-07 at 13 08 51](https://github.com/JuliaCloud/AWS.jl/assets/1675958/292b1cc1-e567-47db-93a2-5dbe20dcb1f0)

I've removed the `bors` requirement. Finally, I'll update our CI trigger rules from this:
```yaml
on:
  push:
    branches:
      - master
      - staging
      - trying
    tags: '*'
  schedule:
    - cron: '0 2 * * *'  # Daily at 2 a.m. UTC
```
To automatically trigger on PRs and the `master` branch.

omus:
GitHub is experiencing some issues with Pull Requests and Actions which may be the reason behind the bors issue I noticed. In any case bors is still deprecated and using a fully GitHub solutions seems better.

iamed2:
I didn't expect that given their docs and how "first-time contributors" worked! That's excellent.

[...]

PR #631, opened 2023-06-07
Drop use of bors to trigger CI tests

omus:
See: https://github.com/JuliaCloud/AWS.jl/issues/630. A couple of repo level changes were required to make this functional (documented [here](https://github.com/JuliaCloud/AWS.jl/issues/630#issuecomment-1581290496)). If we decide to roll this back we need to also roll those changes back.

### DIFF
--- a/test/resources/aws_jl_test.yaml
+++ b/test/resources/aws_jl_test.yaml
@@ -37,10 +37,12 @@ Resources:
             Condition:
               StringEquals:
                 token.actions.githubusercontent.com:aud: sts.amazonaws.com
+              StringLike:
+                # Allow pull requests, the master branch, and any tag to assume this role
                 token.actions.githubusercontent.com:sub:
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/trying
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/staging
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/master
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:pull_request
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/master
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/tags/*
 
   StackInfoPolicy:
     Type: AWS::IAM::ManagedPolicy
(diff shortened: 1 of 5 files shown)


omus:
Need to update our CFN template after #627 to allow all branches.

https://github.com/JuliaCloud/AWS.jl/blob/df874f1f8488a5e8d960475ae71bf4dbd4207baf/test/resources/aws_jl_test.yaml#L40-L43

test/resources/aws_jl_test.yaml
```diff
@@ -37,10 +37,12 @@ Resources:
...
                 token.actions.githubusercontent.com:aud: sts.amazonaws.com
+              StringLike:
+                # Allow pull requests, the master branch, and any tag to assume this role
                 token.actions.githubusercontent.com:sub:
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/trying
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/staging
-                    - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/master
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:pull_request
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/heads/master
+                  - !Sub repo:${GitHubOrg}/${GitHubRepo}:ref:refs/tags/*
```
omus, line 45: Not sure that the `schedule` trigger will be able to assume this role

mattBrzezinski: Is there any difference between [these](https://github.com/JuliaCloud/AWS.jl/blob/df874f1f8488a5e8d960475ae71bf4dbd4207baf/.github/workflows/CI.yml#L4-L11) triggers?

I don't see a reason why doing a `bors try` would succeed but a `schedule` would fail?

mattBrzezinski (approved):

test/resources/aws_jl_test.yaml
omus: I'm pretty sure the `schedule` just runs against the default branch so as long as we allow the `master` branch to work this should be fine.

omus:
@mattBrzezinski and I had a chat about this and came to the conclusion that this approach should lock things down enough that it fufills what we used bors for. We do want to validate if non-member PRs once approved to run will automatically run with each subsequent commit as this could be an attack vector.

Additionally, with the role we've setup for AWS.jl an attacker couldn't actually setup a bitcoin mining operation using our CI and the opportunities for abuse are fairly low.

If we do find out that the the GitHub actions approach isn't good enough then we can go through the effort of doing self-hosted bors which is something we would have had to do anyway if we decide to keep using bors.

PR #633, opened 2023-06-07
hello

kleinschmidt:


omus:
As expected the non-member contributor requires explicit approval to run the CI jobs: 

Untitled

omus:
I've approved the CI jobs to run. Now we'll see what happens with subsequent commits

omus:
The initial commit CI workflows were failing with:
```
Error: Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers
```
That isn't expected but doesn't invalidate this test PR.

omus:
With the second commit being pushed to this PR I'm being asked again to "Approve and run" the workflow. So this behaviour is exactly what we want to limit unwanted actions from taking place in the AWS account for our CI tests.

omus:
I approved the second commit as well to see if the error on the first commit happens again. It does.

PR #638, opened 2023-06-22
Create `assume_role` function

### LINKED ISSUE #623: make an `assume_role` function that uses a role ARN instead of `global_aws_config(; profile)`
Profiles are local (refers to aws config file) but role ARNs are forever.  Functionality is already basically present in AWSS3.jl (in tests), just needs to be cleaned up, tested, and released here.

Also nice to have (but probably a separate issue) woudl be to support role chaining in an ergonomic way.

omus:
Fixes #623. The `assume_role` function allows Julia users to easily assume another IAM role from within Julia. A version of this function was introduced into the test utilities of AWSS3.jl in https://github.com/JuliaCloud/AWSS3.jl/pull/287 as this was particularly useful for assuming testset specific roles.

### DIFF
--- a/src/utilities/role.jl
+++ b/src/utilities/role.jl
@@ -1,0 +1,105 @@
+"""
+    assume_role(role; kwargs...) -> AWSConfig
+    assume_role(::Type{AWSConfig}, role; kwargs...) -> AWSConfig
+    assume_role(::Type{AWSCredentials}, role; kwargs...) -> AWSCredentials
+
+Assumes the IAM `role` via temporary credentials. The current user or role (specified via
+the `aws_config` keyword) must be included in the trust policy of the `role`.
+
+[Role chaining](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_terms-and-concepts.html#iam-term-role-chaining)
+must be manually specified by nesting `assume_role` calls together (e.g. "role-a" has
+permissions to assume "role-b": `assume_role("role-a"; aws_config=assume_role("role_a")`).
+
+# Arguments
+- `role::AbstractString`: The AWS IAM role to assume. Either a full role ARN or just the
+  role name. If only the role name is specified the role will be assumed to reside in the
+  same account used in `aws_config`.
+
+# Keywords
+- `aws_config::AbstractAWSConfig` (optional): User or role attempting to assume the
+  specified `role`
+- `duration::Integer` (optional): Role session duration in seconds.
+- `mfa_serial::AbstractString` (optional): The identification number of the MFA device that
+  is associated with the user making the `AssumeRole` API call. Either a serial number for a
+  hardware device ("GAHT12345678") or an ARN for a virtual device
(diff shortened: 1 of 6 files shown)


src/utilities/role.jl
```diff
@@ -0,0 +1,130 @@
+"""
```
mattBrzezinski, line 1: [Optional] This _could_ live in the credentials file since it's quite related

src/utilities/role.jl
```diff
@@ -0,0 +1,130 @@
...
+
+assume_role(principal, role; kwargs...) = assume_role(AWSConfig, principal, role; kwargs...)
+
+function assume_role(::Type{AWSConfig}, principal::AWSConfig, role; kwargs...)
+    creds = assume_role(AWSCredentials, principal, role; kwargs...)
+    return AWSConfig(creds, principal.region, principal.output, principal.max_attempts)
+end
+
+function assume_role(
+    ::Type{AWSCredentials},
```
mattBrzezinski, line 42: Instead of using `::Type{AWSCredentials}` can we not do `::AbstractAWSConfig` and if folks have their own implementation of credentials they can create their own definition? Matching what we've done w/ `sign!` as an example?

src/utilities/role.jl
omus: It is. However, the `credentials.jl` file currently only contains private code so this seemed nicer. Additionally, the code base could use some refactoring as to the various include files we use.

src/utilities/role.jl
omus: So the intent with having `::Type{AWSCredentials}` is that people can extend the `assume_role` function with their own `::AbstractAWSConfig`. Having this type as the first arguments allows developers to call this function directly instead of having to create an `AWSConfig` first or have to reimplement the entire function. I did make one mistake in `assume_role(principal, role)` which is blocking extensibility though.

Also, having a function that returns `AWSCredentials` directly makes it much easier to create the `renew` function.

omus: The fix to allow for other `AbstractAWSConfig` types to extend `assume_role` was as follows:
```julia
function assume_role(principal, role; kwargs...)
    return assume_role(typeof(principal), principal, role; kwargs...)
end
```
In making this change it became evident that any additional methods of this function would take the form of `assume_role(::Type{T}, ::T, role; kwargs...)`. Additionally, it seems highly unlikely we would ever have want to return a different type than what was passed in via `principal`. With this in mind it makes more sense to define `assume_role(principal::AWSConfig, role; kwargs...)` (as @mattBrzezinski pointed out) leaving us just to deal with 
`assume_role(::Type{AWSCredentials}, principal, role; kwargs...)`. As simply dropping the `Type` argument would conflict with existing methods I decided to use a new function name `assume_role_creds`. In practise I suspect this new function will only be used externally by developers wanting to extend `assume_role`.

mattBrzezinski (approved):

PR #639, opened 2023-06-26
MFA integration tests for `assume_role`

omus:
Follow up to #638. Adds an IAM user with MFA devices such that we can validate the `mfa_serial` and `token` keywords provided by `assume_role`. In order to get this to work we needed to add the following:

1. Support generating TOTP codes (see `totp.jl`)
2. Creating an IAM user with access keys and at least one associated MFA device
3. A restrictive IAM user policy which requires MFA to be used
4. Efficient use of MFA devices such that concurrent tests can avoid being blocked when TOTP codes have been used/consumed.

The design of our MFA device setup is documented in the `setup.jl` file. The TLDR is by using multiple pre-setup MFA devices we can avoid having to wait or perform synchronization between multiple running tests using the AWS resources.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.89.1"
+version = "1.89.2"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.89.0"
+version = "1.89.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.89.2"
+version = "1.89.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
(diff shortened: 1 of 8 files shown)


test/resources/setup.jl
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env julia --project
```
omus, line 1: Having a separate project/manifest isn't strictly necessary but avoids situations where you try to bootstrap the AWS resource setup with an in development version of AWS.jl which may could be broken.

test/resources/setup.jl
```diff
@@ -0,0 +1,200 @@
...
+# attempted. If all TOTPs have been consumed we'll wait until th next time window and try
+# again with a new randomized list of MFA devices.
+#
+# The primary advantage of this approach is that it allows up to 8 API calls to occur
+# concurrently during the same time window. Concurrent integration tests will only be
+# throttled if all TOTP tokens have been consumed.
+function create_or_update_mfa_devices(; user_name, secret_id, num_devices=8)
+    # Assumes user exists
+    # TODO: Should be `list_mfa_devices` instead of `list_mfadevices`
+    r = IAM.list_mfadevices(Dict("UserName" => user_name))
```
mattBrzezinski, line 112: Do we not need to specify these perms in the YAML file too?

mattBrzezinski (approved):

test/resources/setup.jl
```diff
@@ -0,0 +1,200 @@
...
+            false
+        else
+            rethrow()
+        end
+    end
+
+    r = if !secret_exists
+        Secrets_Manager.create_secret(secret_id, params)
+    else
+        Secrets_Manager.update_secret(secret_id, params)
```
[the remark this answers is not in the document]
omus: The role running this script is expected to be a admin role in this AWS account

test/resources/setup.jl
omus: Same as https://github.com/JuliaCloud/AWS.jl/pull/639#discussion_r1242781101

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.89.0"
+version = "1.89.1"
```
omus, line 4:
```suggestion
version = "1.89.2"
```

omus:
As this doesn't actually need a version bump I'll remove that and merge this

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "AWS"
 uuid = "fbe9abb3-538b-5e4e-ba9e-bc94f4f92ebc"
 license = "MIT"
-version = "1.89.1"
+version = "1.89.2"
```
omus, line 4:
```suggestion
version = "1.89.1"
```

Issue #640, opened 2023-06-26
Support credentials via IMDSv2


omus:
AWS.jl's credential handling has support for getting credentials for EC2 instance metadata (specifically IMDSv1). We currently lack [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) support which includes additional security features. As EC2 instances can be configured to only use IMDSv2 we should add support for this.

PR #580, opened 2022-09-26
Get the region from the IMDS

Octogonapus:
This PR adds a step in `aws_get_region` to try the IMDS. This is in line with what the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence) does.

PR #647, opened 2023-07-06
Support data retrevial from Instance Metadata Service Version 2 (IMDSv2)

omus:
Supports retrieval of instance metadata from IMDSv2. AWS.jl already has support for IMDSv1 so that functionality was refactored into the `IMDS` submodule. With this new functionality AWS users can use AWS.jl on instances which [require the use of IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html) for security reasons.

For an extensive review of IMDSv2, see [Add defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities with enhancements to the EC2 Instance Metadata Service](http://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/).

Fixes https://github.com/JuliaCloud/AWS.jl/issues/640.

### DIFF
--- a/src/IMDS.jl
+++ b/src/IMDS.jl
@@ -143,6 +143,9 @@ end
 
 Determine the AWS region of the machine executing this code if running inside of an EC2
 instance, otherwise `nothing` is returned.
+
+# Arguments
+- `session` (optional): The IMDS `Session` used to store the IMDSv2 token.
 """
 region(session::Session) = get(session, "/latest/meta-data/placement/region")
 
--- a/src/IMDS.jl
+++ b/src/IMDS.jl
@@ -144,15 +144,7 @@ end
 Determine the AWS region of the machine executing this code if running inside of an EC2
 instance, otherwise `nothing` is returned.
 """
-function region(session::Session)
-    # TODO: The try/catch is used to work around testing scenarios which require IMDS to not
-    # be present.
-    return try
-        get(session, "/latest/meta-data/placement/region")
-    catch
-        nothing
-    end
-end
+region(session::Session) = get(session, "/latest/meta-data/placement/region")
 
 const _SESSION = Ref{Session}()
 
--- a/src/IMDS.jl
+++ b/src/IMDS.jl
@@ -37,8 +37,6 @@ function refresh_token!(session::Session, duration::Integer=session.duration)
     uri = HTTP.URI(; scheme="http", host=IPv4_ADDRESS, path="/latest/api/token")
     r = _http_request("PUT", uri, headers; status_exception=false)
 
-    # @show r
-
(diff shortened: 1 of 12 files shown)


src/IMDS.jl
```diff
@@ -0,0 +1,162 @@
...
+"""
+    region([session::Session]) -> Union{String, Nothing}
+
+Determine the AWS region of the machine executing this code if running inside of an EC2
+instance, otherwise `nothing` is returned.
+
+# Arguments
+- `session` (optional): The IMDS `Session` used to store the IMDSv2 token.
+"""
+region(session::Session) = get(session, "/latest/meta-data/placement/region")
```
omus, line 150: Replaces this deprecated code:
```julia
ec2_instance_region() =
    try
        ec2_instance_metadata("/latest/meta-data/placement/region")
    catch
        nothing
    end
```

The new code no longer suppresses any exceptions that occur. I believe the reason we originally did this was due to instance metadata failures we saw when running tests against GHA CI as there is a webservice running on the CI instance at the expected instance metadata IP address that isn't a instance metadata service. I've since updated the tests to correct this problem.

However, this is still technically a breaking code change. I'm not concerned about but if a review is we can always put the guard rails on the deprecation.

omus: The try/catch portion was added in https://github.com/JuliaCloud/AWS.jl/pull/580 which seems to be an work around for failing testcases

src/IMDS.jl
```diff
@@ -0,0 +1,162 @@
...
+
+const _SESSION = Ref{Session}()
+
+function __init__()
+    _SESSION[] = Session()
+    return nothing
+end
+
+get(path::AbstractString) = get(_SESSION[], path)
```
mattBrzezinski, line 159: I would move this directly underneath the other `get()` definition.

src/IMDS.jl
```diff
@@ -0,0 +1,162 @@
...
+
+Determine the AWS region of the machine executing this code if running inside of an EC2
+instance, otherwise `nothing` is returned.
+
+# Arguments
+- `session` (optional): The IMDS `Session` used to store the IMDSv2 token.
+"""
+region(session::Session) = get(session, "/latest/meta-data/placement/region")
+
+const _SESSION = Ref{Session}()
```
mattBrzezinski, line 152: And move both `_SESSION` and `__init__()` higher up in the file

mattBrzezinski (commented):
Just did a first pass, made a task for myself to go through and approve this tomorrow.

mattBrzezinski (approved):

PR #658, opened 2023-09-19
Enable streaming upload in S3 : put_object

oconnore:
There were two problems:

1) Request did not allow an IO
2) Signing did not permit an unsigned payload
    (https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html)


src/utilities/sign.jl
```diff
@@ -63,20 +63,18 @@ function sign_aws4!(aws::AbstractAWSConfig, request::Request, time::DateTime)
...
 
-    # SHA256 hash of content...
-    content_hash = bytes2hex(digest(MD_SHA256, request.content))
+    # SHA256 hash of content if data is in memory
+    if !haskey(request.headers, "x-amz-content-sha256")
+        content_hash = bytes2hex(digest(MD_SHA256, request.content))
+        request.headers["x-amz-content-sha256"] = content_hash
+    else
+        content_hash = "UNSIGNED-PAYLOAD"
+    end
```
ararslan, line 72: You can write this a bit more concisely like so:
```suggestion
    content_hash = get!(request.headers, "x-amz-content-sha256") do
        bytes2hex(digest(MD_SHA256, request.content))
    end
```

src/utilities/request.jl
```diff
@@ -121,6 +121,12 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
         "EC2ThrottledException",
     ]
 
+    if isa(request.content, IO)
+        request.headers["x-amz-content-sha256"] = "UNSIGNED-PAYLOAD"
+    else
+        request.headers["Content-MD5"] = base64encode(
+            digest(MD_MD5, request.content))
```
ararslan, line 128: This repository uses the Blue style guide, which puts closing parentheses on the following line in cases like this
```suggestion
            digest(MD_MD5, request.content)
        )
```
Supposedly automatic formatting suggestions are set up but that appears to not have occurred here.

ararslan (commented):
Apologies for the long silence here. Could you add some tests to ensure this works as intended?

PR #667, opened 2024-02-01
The parameter for Object Data should be "body" not "Body".

joaomarceloalencar:
Using "Body" as optional parameter creates empty object. "body" should be used.

This code: 
```
using AWS: @service
@service S3

function upload_file_to_s3(bucket_name, file_path)
    file_name = basename(file_path)
    open(file_path, "r") do file
        S3.put_object(bucket_name, file_name, Dict("Body" => read(file, String)))
    end
    println("File $file_name uploaded to $bucket_name.")
end

template_file="./README.md"
bucket_name="my-unique-bucket-name"
upload_file_to_s3(bucket_name, template_file)
```

Creates an empty object. If you use "body" instead of "Body", it creates the object with the expected content. The documentation uses "Body".

### DIFF
--- a/src/services/s3.jl
+++ b/src/services/s3.jl
@@ -5628,7 +5628,7 @@ GetBucketVersioning.  For more information about related Amazon S3 APIs, see the
 
 # Optional Parameters
 Optional parameters can be passed as a `params::Dict{String,<:Any}`. Valid keys are:
-- `"Body"`: Object data.
+- `"body"`: Object data.
 - `"Cache-Control"`:  Can be used to specify caching behavior along the request/reply
   chain. For more information, see
   http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.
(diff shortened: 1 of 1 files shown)


ararslan:
Hi @joaomarceloalencar, thanks for the contribution! The documentation your PR is editing is automatically generated based on the official AWS documentation, so your change will be overwritten the next time things are updated. As for why you're seeing differences in behavior between `Body` and `body`, have you by chance changed the backend in use? If I recall, there is some difference in behavior with regards to the case sensitivity of request parameters between the HTTP.jl and Downloads.jl backends.

joaomarceloalencar:
Hi @ararslan 

Thank you for your reply. Yeah, I really thought that this documentation was automatically generated. But I thought that the PR was the best way show exactly what confused me when using AWS.jl. I was kind of lost trying to make it work, then someone at Reddit help me out. It is a small thing, but very annoying. 

Regarding the backend, all I did was `pkg.add(AWS)` on Ubuntu 22.04 with Julia 1.10.0. How can I switch between HTTP.jl and Download.jl? Is there a way to AWS.jl pick a backend that agrees with the documentation?

Best Regards.

mattBrzezinski:
Unfortunately this is generated from `aws-sdk-js` which,

* https://github.com/aws/aws-sdk-js/blob/7bb5913e86dc6c28ce2322e05f9a3557fe33640f/apis/s3-2006-03-01.normal.json#L12416-L12429

Does not specify capitalization anywhere :/

PR #676, opened 2024-06-12
Fix JuliaFormatter to 1.0.34

mortenpi:
This downgrades and fixes JuliaFormatter to 1.0.34. I believe this should get rid of the formatting changes in #637.  [1.0.34](https://github.com/domluna/JuliaFormatter.jl/releases/tag/v1.0.34) was the latest JuliaFormatter version when the last API update was merged in #636. My suggestion is to get #637 in with the old formatting, and then figure out how to re-format the generated code.

Note: 1.0.34 doesn't seem to cause any formatting changes on master, so in that sense should be safe to merge.

### DIFF
--- a/.github/workflows/blue_style_formatter.yml
+++ b/.github/workflows/blue_style_formatter.yml
@@ -12,7 +12,7 @@ jobs:
         with:
           version: 1
       - run: |
-          julia  -e 'using Pkg; Pkg.add("JuliaFormatter")'
+          julia  -e 'using Pkg; Pkg.add(name="JuliaFormatter", version="1.0.34")'
           julia  -e 'using JuliaFormatter; format(".", BlueStyle(); verbose=true)'
       - uses: reviewdog/action-suggester@v1
         with:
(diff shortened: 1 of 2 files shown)


ararslan:
> 1.0.34 doesn't seem to cause any formatting changes on master

1.0.56 (what I get when I just `]add JuliaFormatter`) also doesn't seem to produce any changes though 😕

mortenpi:
> 1.0.56 (what I get when I just `]add JuliaFormatter`) also doesn't seem to produce any changes though 😕

On master it doesn't, because they have already been formatted into to `function foo(); ...; end` form, and it doesn't format them back. But if you re-generate the APIs (which generates the `foo() = ...` form), then you should see a difference -- 1.0.56 sticks with `foo() = ...`, but 1.0.34 formats to `function foo(); ...; end`.

ararslan:
Oh I see, thanks. What you have here seems like a reasonable solution then.

ararslan (approved):

mortenpi:
I guess #637 should also update shortly, but FWIW, running locally, this is what the diff should look like with the old formatter: https://github.com/JuliaCloud/AWS.jl/compare/master...mortenpi:AWS.jl:mp/api-diff -- the non-substantial changes seem to be gone.

ararslan:
That's great! Thanks for this

PR #678, opened 2024-06-12
Add dependabot and update actions/checkout to v4

ararslan:


ararslan:
The CI failures seem unrelated (failure to load credentials) but are concerning nonetheless...

PR #682, opened 2024-07-01
Add semver note

ericphanson:
Here I propose using feature releases for automated updates. I believe in the past these have been done as patch bumps, but then one must lock down AWS.jl all the way to the patch number to avoid such changes. Instead I think we should use feature bumps.

cc @mattBrzezinski @glennmoy


mattBrzezinski (approved):

mattBrzezinski:
CI is failing for reasons not cause by this PR, going to log an issue and just merge this in.

PR #689, opened 2024-07-02
Avoid duplicate `include`s causing warnings in tests

### LINKED ISSUE #685: Fix method overwrite warnings in tests
This happens because `include` is being called in a loop:
https://github.com/JuliaCloud/AWS.jl/blob/d7e4dcaadb2838f9efbbdf56e6d3b24c010a01c8/test/runtests.jl#L63-L75

Should be fixable by putting the tests for the different backends in their own module, e.g. have the loop body be something like
```julia
M = Module()
Core.include(M, "file1")
Core.include(M, "file2")
```
The annoying thing is just making sure that those modules load the packages they need and bring in things defined at the top level.

ararslan:
Currently we're calling `include` in a loop over supported backends. The files being `include`d contain definitions which get overwritten when `include` is called a second time, resulting in a lot of warnings that make the test logs noisy. The approach taken here to address this is to wrap things in expressions, define an anonymous module inside of the loop over backends, `Core.eval` the expressions into the anonymous module, and `Core.include` the files into it. There are better ways to do this but this *should* get the job done with minimal required changes.

Fixes #685.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,57 +1,60 @@
-using AWS
-using AWS: AWSCredentials, AWSServices, assume_role_creds
-using AWS.AWSExceptions:
-    AWSException, IMDSUnavailable, InvalidFileName, NoCredentials, ProtocolNotDefined
-using AWS.AWSMetadata:
-    ServiceFile,
-    _clean_documentation,
-    _filter_latest_service_version,
-    _generate_low_level_definition,
-    _generate_high_level_definition,
-    _generate_high_level_definitions,
-    _get_service_files,
-    _get_service_and_version,
-    _get_function_parameters,
-    _clean_uri,
-    _format_name,
-    _splitline,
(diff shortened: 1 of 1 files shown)


ararslan:
A bit cursed but does seem to work

quinnj (commented):
I personally think this makes the testing quite a bit more messy/hard to follow. Having some comments explaining all the evals would certainly help, but I think I'd much prefer splitting things up into proper test-util submodules that can be loaded normally in teh top-level runtests.jl and then relative imported in other submodules. I know it's a bit more rearranging, but I think ti would at least be more approachable/sensible for others who come along and want to contribute.

ararslan:
100% agree with that. This was the path of least effort, haha. I'll see if I can revisit this soon but in the meantime, for anybody reading this, do feel free to implement Jacob's suggestion in your own PR.

Issue #703, opened 2025-05-20
CI workflows fail for PR contributions from forks due to use of secrets


omus:
We use the GitHub Action secret `AWS_ACCOUNT_ID` in the CI workflow. For user contributions from forks these secrets are not populated resulting in CI workflow failing. We should rethink this as we definitely want to run tests on contributions (once maintainers allow those workflows to run).

The quick solution to this it to hard-code the AWS account ID into the workflow.

omus:
It was brought to my attention that OIDC also doesn't work on forks. Looks like our best bet is using a GitHub merge queue

omus:
We now use merge queues which allows us to have access to secrets on forks

PR #704, opened 2025-05-20
Use GitHub merge queues

omus:
In order to address #703 we need to deal with OIDC only being available from the current repository and not from forks. It appears that GitHub merge queues can help with the situation. The only real downside here compared to the old Bors system is that you cannot execute the tests without attempting a merge. This means that reviewers need to look over the code first (we required this before anyway) and approve the PR before tests are run. If they do so the PR can enter the merge queue and will be merged only if the CI tests pass.

In order to test this properly I've enabled the GitHub merge queue on the `master` branch.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,5 +1,6 @@
 name: CI
 on:
+  merge_group:
   pull_request:
   push:
     branches:
@@ -10,6 +11,7 @@ on:
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    if: ${{ github.event_name != 'pull_request' }}  # Avoid running on PRs as OIDC will fail when running on a fork
     permissions:
       # Required for interacting with GitHub's OIDC Token endpoint:
       # - Checking out the repository (`contents: read`)
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,9 +1,15 @@
 name: TagBot
 on:
-  schedule:
-    - cron: 0 * * * *
+  issue_comment:
+    types:
+      - created
+  workflow_dispatch:
+    inputs:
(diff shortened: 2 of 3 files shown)


omus:
I'm going to try out the "Merge with ready" without approval

omus:
Okay, this does wait for approval first

omus:
Note we could have tests still trigger automatically on `pull_request` triggers. Those would work like they did before if the PR was created from this repository however for forks we would still see failures. We could add some extra logic to circumvent that if we really wanted it but I think it's probably best for us to stick with the merge queue setup for all users to ensure that it's functional all the time.

mattBrzezinski (approved):

DilumAluthge:
With that being said, I'm a little worried that the current setup here won't work.

When the job is skipped, the job name is: `Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}`. You can see that job name on this PR (click on "View Details").

That is, when the job is skipped, GitHub does not populate the matrix variables into the job name.

However, when the job is run, GitHub will populate the job name.

But, AFAICT, GitHub's branch protection settings don't understand that, and thus AFAICT there's no way to have both (both the "populated" and "unpopulated" versions) as "required status checks".

In RegistryCI.jl and CompatHelper.jl, we worked around this by not having our job name depend on matrix variables.

omus:
Thanks for pointing these issues out. We do have "require status checks" enabled but didn't add any checks to the list. Interesting no checks show up at all there. Not even checks which don't use variables in the job name (like formatting). That makes me think the checks may need to be run on `master` to show up but I'll look into that.

omus:
> But, AFAICT, GitHub's branch protection settings don't understand that, and thus AFAICT there's no way to have both (both the "populated" and "unpopulated" versions) as "required status checks".

It looks like you can add both. Does this not work?

Screenshot 2025-05-21 at 09 00 25

PR #705, opened 2025-05-20
Create separate documentation workflow

omus:
I noticed that the `docs/Manifest.toml` was quite out of date so I updated the documentation job to no longer require this. Additionally, it made sense to make the documentation into a separate workflow to only execute it when necessary.


.github/workflows/Documenter.yml
```diff
@@ -0,0 +1,49 @@
+name: Documenter
+on:
+  merge_group:
```
omus, line 3: Not currently a required status check
```suggestion
```

mattBrzezinski (approved):

PR #708, opened 2025-05-20
CI: Use the same job name for skipped and non-skipped jobs, run CI on non-fork PRs, and skip CI on fork PRs

DilumAluthge:


DilumAluthge:
@omus @mattBrzezinski This PR is intended to fix the problem I described in https://github.com/JuliaCloud/AWS.jl/pull/704#issuecomment-2895666367.

.github/workflows/CI.yml
```diff
@@ -9,71 +9,37 @@ on:
...
+  test-julia-lts:
+    if: ${{ github.event_name != 'pull_request' || github.repository == github.event.pull_request.head.repo.full_name }}
+    uses: ./.github/workflows/reusable_test_workflow.yml
+    with:
+      julia-version: 'lts' # Long Term Stable
+  test-julia-v1:
+    if: ${{ github.event_name != 'pull_request' || github.repository == github.event.pull_request.head.repo.full_name }}
+    uses: ./.github/workflows/reusable_test_workflow.yml
+    with:
+      julia-version: '1' # Latest Release
```
omus, line 42: Why not use a matrix with the resuable workflow? https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow

I assume you do this for the fixed job name? Couldn't you work around that by passing the job name in as an argument to the reusable workflow? That way the name would be fully resolved

.github/workflows/reusable_test_workflow.yml
```diff
@@ -0,0 +1,56 @@
...
+        required: true
+        type: string
+jobs:
+  test:
+    permissions:
+      actions: write # Required for deleting old caches: https://github.com/julia-actions/cache#usage
+      contents: read # Required for checking out the repository
+      id-token: write # Required for interacting with GitHub's OIDC Token endpoint:
+      # For more info, see the OIDC docs:
+      # https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-permissions-settings
```
omus, line 16: As workflows get more complex eventually permissions sets need to be merged. Due to that I usually document the required permissions above and not inline as things can get messy fast: 
```suggestion
    # These permissions are needed to:
    # - Checking out the repository (`contents: read`)
    # - Interact with GitHub's OIDC Token endpoint: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-permissions-settings
    # - Delete old caches: https://github.com/julia-actions/cache#usage
    permissions:
      actions: write
      contents: read
      id-token: write
```

omus:
I've come up with a cleaner solution to this problem in: https://github.com/JuliaCloud/AWS.jl/pull/712

PR #710, opened 2025-05-21
Rename formatting workflow job name

omus:
Since we want to add the formatting job as a required check for merging it made sense to me to change the name at this time.


omus:
I'm going to add this to the merge queue and we'll see how the required checks worked. I have some concerns that I've added s set of checks that is impossible to fufill

PR #711, opened 2025-05-21
GitHub merge queue testing

omus:
Trying a GitHub merge queue where jobs fail. Here are the branch protection rules used:

fake-main

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -54,23 +54,25 @@ end
 testset_role(role_name) = "AWS.jl-$role_name"
 
 @testset "AWS.jl" begin
-    include("AWSExceptions.jl")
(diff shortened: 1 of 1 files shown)


omus:
So without adding required checks it appears that no workflows are executed with the `merge_group` trigger. My expectation was that workflows with this trigger would execute after the PR is added to the merge queue but it appears this trigger is needed AND the check needs to be required.

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#configuring-continuous-integration-ci-workflows-for-merge-queues

That means we need to both run the check in the PR first to be able to be added to the merge queue at all. That's not great as it means forks will still fail due to the use of secrets.

PR #714, opened 2025-05-22
CI cleanup

omus:
Some minor changes to wrap up the recent CI work

### DIFF
--- a/.github/workflows/JuliaFormatter.yaml
+++ b/.github/workflows/JuliaFormatter.yaml
@@ -1,3 +1,4 @@
+---
 name: JuliaFormatter
 on:
   merge_group:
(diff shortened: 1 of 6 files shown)


.github/workflows/JuliaFormatter.yaml
```diff
@@ -34,3 +35,6 @@ jobs:
         if: ${{ !cancelled() && github.event_name == 'pull_request' }}
         with:
           tool_name: JuliaFormatter
+          # Post results on all results and not just changed files:
```
mattBrzezinski, line 38: I forget off hand, but is the entire repo formatted? Would be annoying for the first PR to be made if it hasn't been

Issue #717, opened 2025-06-06
Avoid passing AWS configuration into API calls


omus:
It's always been a bit odd that AWS.jl allows users to pass in an AWS configuration for each API call. This package also supports a globally set AWS configuration but this approach is problematic when using threading. To address this problem we've ended up supporting passing an AWS configuration into all low-level or high-level API calls.

Additionally, I've noticed a pattern with packages providing alternate interfaces which utilize AWS.jl where the types provided by these packages end up keeping a copy of the AWS configuration within the struct. For example `S3Path`:

https://github.com/JuliaCloud/AWSS3.jl/blob/772152707c7d3f7b4c2d45a85b844470cda44715/src/s3path.jl#L18

Embedding the AWS configuration in the struct is problematic in that it makes it hard to switch IAM roles and utilize the same `S3Path` instances.

Note that other AWS libraries like boto3 use an OOO approach where one generates a client with all of the configuration first. Users can make alternate clients with different roles if they wish.

I think we should consider embedding the AWS configuration within types such as `S3Path` to be bad practise. That does leave us with the problem that globally setting the AWS configuration doesn't work well with threading. I think the approach we should use is to use a `ScopedValue` temporarily switch the AWS configuration. With this approach AWS API calls can call `current_aws_config` to get the active configuration to use. The use of `ScopedValue`s will allow the function to be thread safe and avoid having to pass the configuration to each call and should eliminate the need to embed AWS configuration information into types.

I also foresee further downstream effects from this change helping with changes such as https://github.com/JuliaCloud/AWS.jl/issues/341 since we no longer need to worry about possible keyword conflicts.

ericphanson:
I think allowing passing the config per-usage is both good design and in-line with other clients.

I think replacing the existing global config with ScopedValues is good, but we should continue to allow the config to be passed on a per-function basis, and just default to the scoped one. So it is overridable per call but with a safer fallback (ScopedValue instead of regular global).

I will try to articulate my reasoning below.

## Passing config per-usage is a good design

### Design reason 1: local info is easier to reason about

[...]

PR #718, opened 2025-06-06
Support scoped AWS configurations

omus:
Fixes: https://github.com/JuliaCloud/AWS.jl/issues/717

### DIFF
--- a/src/AWSConfig.jl
+++ b/src/AWSConfig.jl
@@ -92,3 +92,46 @@ function _update_creds!(aws_config::AWSConfig)
 
     return creds
 end
+
+const _SCOPED_AWS_CONFIG = ScopedValue{AbstractAWSConfig}()
+const _INITIAL_AWS_CONFIG = Ref{AbstractAWSConfig}()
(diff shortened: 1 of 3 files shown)


src/AWSConfig.jl
```diff
@@ -92,3 +92,46 @@ function _update_creds!(aws_config::AWSConfig)
...
+    # Load non-scoped configuration upon first usage. We'll make use of a lock to avoid race
+    # conditions but only lock when necessary to avoid unnecessary overhead with subsequent
+    # calls.
+    if !isassigned(_INITIAL_AWS_CONFIG)
+        @lock _INITIAL_AWS_CONFIG_LOCK begin
+            if !isassigned(_INITIAL_AWS_CONFIG)
+                _INITIAL_AWS_CONFIG[] = AWSConfig()
+            end
+        end
+    end
```
omus, line 116: I think we should defer this until it is absolutely necessary. For instance if a user only accesses the AWS configuration passed in using `with` then there is no reason to load the standard AWS configuration. In fact, doing so could be a bad idea in limited CI environments where not default configuration is available.

PR #726, opened 2025-07-02
Separate unit from integration tests

### LINKED ISSUE #725: Separate tests into unit/integration tests
When we switched to using merge queues in #713 we decided to only run tests that require OIDC access to be run only in the merge queue. The result of this is that no tests are run during the PR.

There are quite a few tests however which do not require OIDC access and those tests could be run as part of the PR and in the merge queue. Doing this requires that we separate the unit tests from the integration tests

The setup for this separation was started in #720 due to an issue with running some specific tests on multiple backends which was not required: https://github.com/JuliaCloud/AWS.jl/issues/720#issuecomment-2970735866_

omus:
With the switch to using GHA merge queues to deal better with supporting contributions from forks we've discovered a need to run our unit tests and our integration tests separately. By separating these tests we can run the tests which do not require access to the AWS account to be triggered by PR events and reserve the integration tests to be run in the merge queue.

Additionally, this allows for users to run the AWS.jl unit tests locally without any AWS account setup which should allow issues to be caught earlier.

Summary of changes:
- Separate existing tests into unit tests and integration tests. Unit tests now reside under `test/unit` where as integration tests reside under `test/integration`. For the most part tests were left as is unless changes were required.
- Rename `test/configs` directory to `test/config`.
- Rename `test/resources` directory to `test/resource`.
- Removed reference `AWS_CONFIG` in favor of `global_aws_config`
- Constant `TEST_MINIO` was renamed to `RUN_MINIO_INTEGRATION_TESTS`
- Unit tests attempt to create an environment where an AWS configuration is not present
- Unit tests use `AWSConfig(; creds=nothing, region="...")` to create a dummy config
- Integration tests are now opt-in instead of opt-out.

Depends on:
- #727

Fixes #725


maganaluis (approved):

omus:
Looks like the unit tests are polluting the global AWS configuration:
```
RequestError: Could not resolve host: sts.us-example-1.amazonaws.com while requesting https://sts.us-example-1.amazonaws.com/
```

PR #730, opened 2025-07-03
Refactor tests to avoid replace warnings

omus:
Alternate approach which replaces: https://github.com/JuliaCloud/AWS.jl/pull/689. Refactors the tests such that performing multiple includes of the integration tests doesn't result in "replacing module" warnings.


omus:
Interesting this happened now. I think I'll clear the cache and move on as this bug pre-existed before this PR. It's interesting it hasn't happened before.

PR #737, opened 2025-10-03
Change OIDC debugger actions

omus:
The action we were using was broken and has been archived: https://github.com/github/actions-oidc-debugger


omus:
I'm not having any luck with https://github.com/steve-todorov/oidc-debugger-action directly but the copied logic works here. I'll call that a win and move on

omus:
I created an issue for this: https://github.com/steve-todorov/oidc-debugger-action/issues/1

PR #744, opened 2025-11-19
Allow failures from OIDC debug step

omus:
Noticed this failure which only occurred once it ran on `master`. As this step is only here for debugging purposes we can continue on failures

https://github.com/JuliaCloud/AWS.jl/actions/runs/19514947303/job/55864204971


omus:
Looks like this should be addressed in the latest, unreleased, version: https://github.com/steve-todorov/oidc-debugger-action/blob/b6d83b24a379e77233fb6b26f1c1a662c4bf443c/action.yml#L33. Doesn't hurt to add this in though.

PR #770, opened 2026-04-07
Retry request too skewed

### LINKED ISSUE #496: Add retry for `RequestTimeTooSkewed`

ericphanson:
closes #496
redo of #497 

I wanted to add a test but couldn't quickly find where to add it, so I asked opus to write the test (2nd commit) with the prompt:

> I have a very small diff in requst.jl to add a retry for `REQUEST_TIME_TOO_SKEWED`. I want to add a minimal test that it now gets retried, but I'm not sure there are any tests for retries - I can't find them. Propose a small test: where should it go, how should it work, etc. Existing tests are in the test dir.

in plan mode, which I then accepted its plan. I stopped it after it implemented and started running tests since I didn't want to waste tokens on it reading test output. Then I ran tests myself which passed. This used $0.42 of tokens.

### DIFF
--- a/src/utilities/request.jl
+++ b/src/utilities/request.jl
@@ -99,7 +99,7 @@ Submit the request to AWS.
 function submit_request(aws::AbstractAWSConfig, request::Request; return_headers=nothing)
     aws_response = nothing
     TOO_MANY_REQUESTS = 429
-    REQUEST_TIME_TOO_SKEWED = "RequestTimeTooSkewed"
+    TIME_SKEWED_CODES = ["RequestTimeTooSkewed"]
     EXPIRED_ERROR_CODES = ["ExpiredToken", "ExpiredTokenException", "RequestExpired"]
     REDIRECT_ERROR_CODES = [301, 302, 303, 304, 305, 307, 308]
     THROTTLING_ERROR_CODES = [
@@ -192,7 +192,7 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
         end
 
         # https://github.com/JuliaCloud/AWS.jl/issues/496
-        if e.code == REQUEST_TIME_TOO_SKEWED
+        if e.code in TIME_SKEWED_CODES
             return true
         end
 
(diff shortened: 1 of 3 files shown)


src/utilities/request.jl
```diff
@@ -99,6 +99,7 @@ Submit the request to AWS.
 function submit_request(aws::AbstractAWSConfig, request::Request; return_headers=nothing)
     aws_response = nothing
     TOO_MANY_REQUESTS = 429
+    REQUEST_TIME_TOO_SKEWED = "RequestTimeTooSkewed"
```
omus: Mirroring the rest of the logic in this function
```suggestion
    TIME_SKEWED_CODES = ["RequestTimeTooSkewed"]
```

src/utilities/request.jl
```diff
@@ -189,6 +190,12 @@ function submit_request(aws::AbstractAWSConfig, request::Request; return_headers
                 ),
             )
         end
+
+        # https://github.com/JuliaCloud/AWS.jl/issues/496
+        if e.code == REQUEST_TIME_TOO_SKEWED
```
omus:
```suggestion
        if e.code in TIME_SKEWED_CODES
```

omus (approved):
Looks reasonable

test/patch.jl
```diff
@@ -87,6 +87,26 @@ function _throttling_patch(retries::Ref{Int})
     return p
 end
 
+function _time_too_skewed_patch(retries::Threads.Atomic{Int})
```
ericphanson:
```suggestion
function _time_too_skewed_patch(retries)
```

omus (approved):

PR #775, opened 2026-06-24
Support HTTP.jl 2.x alongside 1.x

krynju:
## Summary

Adds support for **HTTP.jl 2.x** while retaining 1.x, selected at load time via a small compatibility layer. `Project.toml` now declares `HTTP = "1, 2"`.

HTTP.jl 2.0 is a breaking release; the relevant changes for AWS.jl are bridged in `src/utilities/http_compat.jl` (feature-flagged on the presence of the `HTTP.Exceptions` submodule, which exists only on 1.x):

- `StatusError` constructor changed from `(status, method, target, response)` → `(status, response)`.
- `RequestError`, `iserror`, `isbytes`, `resource`, and the `HTTP.Exceptions` submodule were removed — `HTTP.isrecoverable` and local shims replace them.
- Response bodies are buffered as `Vector{UInt8}` (1.x exposed a stream object when a `response_stream` was supplied), header names are canonicalized, and connection-attempt timeouts surface as `TimeoutError` rather than `ConnectError`.

Tests construct version-appropriate exceptions/bodies. The IMDS mock no longer reimplements HTTP.jl's private retry internals (which differ across versions) and instead honors only the `retry_check` path IMDS actually relies on.

## GitHub.jl / HTTP 2 interaction

`GitHub.jl` (used only by the `parse_aws_metadata` API-generation tooling) is retained as a dependency. Registered GitHub.jl currently pins `HTTP < 2`, so **with GitHub.jl present the resolver selects HTTP 1.x today**. Once GitHub.jl publishes a release with HTTP 1+2 support, HTTP 2 resolves with no further change (widen the `GitHub` compat bound if that support arrives in a new major version).

> Note: the first commit temporarily replaced GitHub.jl with hand-rolled REST calls to unblock HTTP 2 immediately; the second commit restores GitHub.jl. The **net** change to `src/api_generation/` versus `master` is none. Happy to squash before merge.

## Testing

Unit suite passes under both HTTP 1.11 and HTTP 2.4 (the latter exercised by temporarily dropping GitHub.jl to allow HTTP 2 resolution).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ericphanson:
I'm not super familiar with http v2 vs v1 but the code looks more or less reasonable to me. However I think something that would de-risk this a lot is if you updated CI to run on BOTH major versions of HTTP. This would ensure we are truly compatible with both.

We have done that before with Onda.jl and its dependency on Arrow, see

https://github.com/beacon-biosignals/Onda.jl/blob/b8b7720549c35cb57ee587b30a1b610065304253/.github/workflows/CI.yml#L27-L29

https://github.com/beacon-biosignals/Onda.jl/blob/b8b7720549c35cb57ee587b30a1b610065304253/.github/workflows/CI.yml#L39-L45

and I think something similar could probably done here

krynju:
> I'm not super familiar with http v2 vs v1 but the code looks more or less reasonable to me. However I think something that would de-risk this a lot is if you updated CI to run on BOTH major versions of HTTP. This would ensure we are truly compatible with both.
> 
> We have done that before with Onda.jl and its dependency on Arrow, see
> 
> https://github.com/beacon-biosignals/Onda.jl/blob/b8b7720549c35cb57ee587b30a1b610065304253/.github/workflows/CI.yml#L27-L29
> 
> https://github.com/beacon-biosignals/Onda.jl/blob/b8b7720549c35cb57ee587b30a1b610065304253/.github/workflows/CI.yml#L39-L45
> 
> and I think something similar could probably done here

Thanks! I've set up HTTP v1 CI now. Couldn't do it before, because GitHub.jl didn't have a HTTP.jl v2 compatible version.
I gave it two more Fable reviews and I'll bump our nightly to the latest version to see if anything breaks.

After that I think we can merge next week

krynju:
Should be good to go, I don't see anything new appearing in our nightly runs.

tanmaykm:
I think we can merge in a bit if there are no other comments!

tanmaykm (approved):

PR #780, opened 2026-08-13
AWS high-level interface uses keywords

omus:


omus:
Made some good progress here. I've been experimenting with a `@kwalias` macro which allows for alternative keyword sets to be defined for Julia methods. The thought is we can support both the original, typically, CamelCase keywords and also snake_case keywords.

Doing this revealed that some of the keyword arguments (e.g. `Export`) are Julia keywords when converting to snake_case so some special casing is needed. Ideally, I'll get this PR ready by the end of the week.

PR #785, opened 2026-09-15
CI: download MinIO from GitHub releases

krynju:
## Summary

The MinIO integration tests fail in every merge queue and `master` run since mid-August. The `MinIO server setup` step in `CI.yaml` downloads the server binary from `dl.minio.io`, which now returns HTTP 410 Gone for every path:

> The open-source MinIO Server, MinIO Client (mc) and MinIO KES projects are archived and no longer maintained. [...] These files are no longer served from this site.

Because the download used `curl -sSLO` without `-f`, the error page was saved as `./minio`, made executable and started in the background, where it failed with `line 1: 410: command not found`. The step still reported success, and the tests later failed with:

```
HTTP.ConnectError for url = `http://127.0.0.1:9000/anewbucket`: IOError: connect: connection refused (ECONNREFUSED)
```

This only shows up in merge queue and `master` runs because integration tests are skipped on `pull_request` events.

## Fix

- Download from the [minio/minio GitHub releases](https://github.com/minio/minio/releases), which still serve binaries for all platforms. The release is pinned to `RELEASE.2025-09-07T16-13-09Z`, the most recent one with attached binaries. The 2025-10-15 release has none.
- Pass `-f` to curl so a bad download fails the step instead of producing a dead server.

The asset naming follows the same `linux-amd64` / `darwin-amd64` scheme the step already uses for `host_os`, so the macOS branch keeps working.

## Verification

Downloaded the pinned `linux-amd64` binary and started it locally with the workflow's exact flags (`server --compat --quiet`) and credentials. It answered the health endpoint within two seconds.

Once merged, the merge queue for #783 should go green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


krynju:
Folded into #783, since neither change alone gets CI green.

